PR #58, opened 2017-02-13
Build client on Travis

igboyes:


igboyes:
Need to put client and server build into same pull request.

Issue #149, opened 2017-09-26
Isolate name overflow


igboyes:
In the isolate editor, isolate names can currently messily overflow the isolate nav button on the left side.

Use ellipses to make this look better.

igboyes:
Fixed in #152

PR #152, opened 2017-10-03
Client improvements

igboyes:
- re-implement some synchronous file operations with ``aiofiles``
- fix block in ``CreateSample`` caused by Skewer running without ``--quiet`` set
- replace ``halogen`` with ``halogenium`` and make loading icon use consistent (``ClipLoader``)
- get all settings inputs working
- add query to GET /api/viruses to retrieve a id-name map
- fix some Codacy issues


igboyes:
Fixes #149

PR #157, opened 2017-10-06
Potential fix for Travis CI timeouts

igboyes:
Remove pytest timeout plugins.

I originally used these to figure out why long running subprocesses were hanging during builds and also local testing. This issue was fixed a while ago and now ``pytest-faulthandler`` is not necessary and is actually failing long-running tests.

The plugins ``pytest-faulthandler`` and ``pytest-timeout`` have been removed from ``requirements.txt`` and the ``pytest`` call is updated in ``.travis.yml``.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ before_script:
 - sleep 15
 
 script:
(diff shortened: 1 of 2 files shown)


igboyes:
Looks like this worked.

Issue #153, opened 2017-10-04
Create API key authentication system


igboyes:
Requires the following parts:
- key storage in user document
- API endpoints for: create, update, remove
- API tests
- client UI

igboyes:
Resolved in #161

igboyes:
Resolved in #161

PR #161, opened 2017-10-10
Authorization system

igboyes:
- make API keys work with endpoints
- use ``Authorization`` header
- put API key in their own Mongo collection
- add setting ``enable_api`` for controlling API availability
- add endpoint for removing all API keys for an authorized account


igboyes:
Resolves #153

PR #352, opened 2018-02-06
Schema API 

igboyes:
- extend virus edit API endpoint to allow modification schema
- add schema field to all virus documents during database init
- fix API key UI
- fix lint issues in `UploadOverlay`


igboyes:
Will leave this for another more complete PR.

PR #370, opened 2018-02-12
Virus schema backend

### LINKED ISSUE #102: Virus schemas
Allow the assignment of a schema to each virus. A schema describes what an isolate of that virus should look like.

Benefits are:
- the segment names can be displayed instead of the sequence id (accession) and definition
- the ordering of the segments will inform the layout of the coverage viewer
- validating user inputs

Goals:
- [x] API endpoints for editing schemas
- [x] API endpoints for annotating sequences as schema segments 
- [x] history context for schema modifications
- [ ] server-side validation of isolates based on schema rules
- [ ] schema edit interface in virus view
- [ ] _required_ setting for schema segments (ie. some can be optional)
- [ ] segment selection `<select />` element in sequence list items
- [ ] order pathoscope coverage charts based on schema instead of length

Miminko:
- resolves #102 
- fixed minor tab spacing in uploadOverlay.js


igboyes:
Can you make the following changes:

- [ ] move *Schema* tab between *Virus* and *History*
- [ ] render a `<NoneFound />` component from virtool base components when no segments are defined
- [ ] send an API request every time the order changes instead of having save button

*Modal*
- [ ] change title to *Add Segment*
- [ ] change *Segment Name* input label to *Name*
- [ ] justify modal checkbox to left rather than right
- [ ] make sure modal hides when the greyed-out background is clicked (check `onHide` prop of `<Modal />`)

*Segment List Item*
- [ ] make the `required` tag look better; suggest using a `<Label />` component with `warning` or `info` styling
- [ ] spread out elements more evenly
- [ ] put `5px` of padding between edit and remove icons

PR #420, opened 2018-02-28
Input form validation

### LINKED ISSUE #404: Consistent input and form errors
Develop consistent standard for styling form and input errors.

#### Error Types
1. Input error
- specific input values are unacceptable
- errors appear the same whether they where generated on the client or server
- deprecate `<Popover />`-based errors

2. Form error
- form input taken as a whole is invalid
- example is incorrect username and password
- display of error makes it obvious it applies to the whole form

#### Style Ideas
- no validation icons necessary
- apply color to input element border
- position input-specific errors below bottom-left or -right corner of input
- make it obvious form errors apply to whole form
- take a look at [HTML5 validation](https://pageclip.co/blog/2018-02-20-you-should-use-html5-form-validation.html)

#### To Do
- [x] modify base `<Input />` to easily add error messages to component
- [x] make sure changes to `<Input />` are compatible with `<InputSave />`
- [x] make all occurences of `<Input />` in the app consistently styled

### LINKED ISSUE #417: Changing password under Settings>Users>Change Password results in 404
Cannot change passwords via edit user, but can change passwords from under Account>General

Miminko:
- resolves #404 
- resolves #417 
- fixes `QuickAnalyze` modal errors

### DIFF
--- a/client/src/js/account/components/Email.js
+++ b/client/src/js/account/components/Email.js
@@ -1,7 +1,6 @@
 import React from "react";
-import { map } from "lodash-es";
 import { connect } from "react-redux";
-import { Alert, Col, Panel, Row } from "react-bootstrap";
+import { Col, Panel, Row } from "react-bootstrap";
 
 import { updateAccount } from "../actions";
 import { Button, Input } from "../../base";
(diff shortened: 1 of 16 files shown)


igboyes:
Changes:

1. Increase padding/margin below input and error components so error message isn't smushed against elements below.
2. Slightly increase space between error text and input box
3. Prevent downward shift of elements when error message displays. Easiest way is probably to have some hidden text when there is no error to maintain the same spacing as when there is an error.
4. Use input error when virus name in `<CreateVirus />` modal already exists. Currently the error is shown in an `<Alert />` component.

PR #454, opened 2018-03-12
Add hmm annotations on import completion

### LINKED ISSUE #303: Refresh HMM annotations when import complete

Miminko:
- resolves #303

### DIFF
--- a/client/src/js/hmm/components/Installer.js
+++ b/client/src/js/hmm/components/Installer.js
@@ -17,14 +17,23 @@ const steps = [
 
 class HMMInstall extends React.Component {
 
-    handleRefresh = () => {
-        this.props.onRefresh();
+    getProgress = () => {
+        const progress = 20 * (steps.indexOf(this.props.process.step) + this.props.process.progress);
+        return progress;
+    }
+
+    componentDidUpdate () {
+        const progress = this.getProgress();
+
+        if (progress === 100) {
+            this.props.onRefresh();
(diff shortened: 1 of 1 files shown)


client/src/js/hmm/components/Installer.js
```diff
@@ -17,14 +17,23 @@ const steps = [
 
 class HMMInstall extends React.Component {
 
-    handleRefresh = () => {
-        this.props.onRefresh();
+    getProgress = () => {
+        const progress = 20 * (steps.indexOf(this.props.process.step) + this.props.process.progress);
+        return progress;
```
igboyes: Unnecessary `const` definition.

```
getProgress = () => (
    20 * (steps.indexOf(this.props.process.step) + this.props.process.progress)
);

File: client/src/js/hmm/components/Installer.js [hunk 2]
```diff
@@ -17,14 +17,23 @@ const steps = [
...
+    }
+
+    componentDidUpdate () {
+        const progress = this.getProgress();
+
+        if (progress === 100) {
```
igboyes: To be safe, you should use `prevProps` and the conditional  `(prevProps.progress !== 100 && props.progress === 100)`.

Doing this would probably also require making `getProgress` a pure function that takes `props` as an input so that you can pass `prevProps` to it.

Also, you have an unnecessary definition of `progress` here.

PR #477, opened 2018-03-15
Virus filter unverified

### LINKED ISSUE #470: Filter viruses by unverified flag
- [x] add right-justified icon with tooltip to virus item when unverified (`!verified`)
- [x] add filter button in *Viruses* toolbar to filter viruses to show only unverified ones

Unverified viruses only can be fetched from the API using `/api/viruses?verified=false`.

Miminko:
- resolves #470


igboyes:
This works really well. Can you add an [active state](https://react-bootstrap.github.io/components/buttons/#buttons-active) to the button when the filter is active.

Then I will merge.

PR #501, opened 2018-03-20
Resource limits request handling

### LINKED ISSUE #495: Task-specific limits warnings
Apply the following errors to **Task-specific limits** in the jobs settings view:
- [ ] invalid value entered in task-specific limits results in context-specific error
- [ ] adjustment of task-specific limits to match changed resource limits results in Alert at top of view

Complete this after requisite API changes.

Miminko:
- resolves #495

### DIFF
--- a/client/src/js/base/InputSave.js
+++ b/client/src/js/base/InputSave.js
@@ -21,9 +21,15 @@ export class InputSave extends React.Component {
         onSave: PropTypes.func.isRequired,
         label: PropTypes.string,
         type: PropTypes.string,
+        min: PropTypes.number,
+        max: PropTypes.number,
+        step: PropTypes.number,
         initialValue: PropTypes.any,
         autoComplete: PropTypes.bool,
+        onInvalid: PropTypes.func,
+        onChange: PropTypes.func,
         disabled: PropTypes.bool,
+        noMargin: PropTypes.bool,
         error: PropTypes.string
     };
 
@@ -48,9 +54,15 @@ export class InputSave extends React.Component {
      * make the form lose focus. The form is intentionally blurred once an updated initialValue is received in props.
      */
     handleBlur = (e) => {
+        // If click on focus element that does not submit, reset value
         if (!this.state.pending && e.relatedTarget && e.relatedTarget.type !== "submit") {
             this.setState({value: this.props.initialValue});
+        // If click on non focus element, reset value
+        } else if (!e.relatedTarget) {
+            this.setState({value: this.props.initialValue});
         }
+        // If click on focus element that does submit, keep value
+        this.props.onChange(e);
     };
 
     /**
@@ -62,6 +74,7 @@ export class InputSave extends React.Component {
         if (!this.props.disabled) {
             this.setState({value: e.target.value});
         }
+        this.props.onChange(e);
     };
 
     /**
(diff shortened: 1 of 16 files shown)


client/src/js/base/InputSave.js
```diff
@@ -97,25 +110,31 @@ export class InputSave extends React.Component {
     };
 
     blur = () => {
+        this.inputNode.blur();
         this.buttonNode.blur();
     };
 
     render () {
+        const formStyle = this.props.noMargin ? "0px" : "15px";
```
igboyes: Change to `formMargin`. The implied convention is that variables with the `Style` suffix are always style objects compatible with React components.

Here you are describing a margin value.

By the way, `0` values for CSS properties can just be 0. No unit is necessary.

client/src/js/settings/components/Jobs/Resources.js
```diff
@@ -1,59 +1,193 @@
...
+        }
+    };
+
+    handleSaveMem = (e) => {
+        if (e.value <= this.state.memUpperLimit && e.value >= this.props.memLowerLimit) {
+            this.props.onUpdateMem(e);
+        } else {
+            this.setState({ errorMem: true });
+        }
+    };
```
igboyes: Lines 56 - 90 are repetitive.

The four small handler functions could be reduced to two by assigning `name` props to your `<InputSave />` components and using that to update the state correctly.

The `name` prop could also be used to condense your save handlers into a single function.

client/src/js/settings/components/Jobs/Resources.js
```diff
@@ -1,59 +1,193 @@
...
+
+    render () {
+        if (this.props.resources === null) {
+            return <LoadingPlaceholder />;
+        }
+
+        const errorMessageProc = this.state.errorProc ? "Cannot go over or under resource limits" : null;
```
igboyes: Can you reformat this message to "Value must be between {min} and {max}"?

This logic can be put into a separate function defined at the top of the module so that it can be reused for both the `proc` and `mem` values.

client/src/js/settings/components/Jobs/Resources.js
```diff
@@ -1,59 +1,193 @@
...
+
+class Resources extends React.Component {
+
+    constructor (props) {
+        super(props);
+
+        this.state = {
+            errorProc: false,
+            errorMem: false,
+            procUpperLimit: this.props.proc,
```
igboyes: Is it necessary to move `procUpperLimit` and `memUpperLimit` into state?

I think you are doing it to deal with lag between the component loading and the `getResources` request returning...

client/src/js/settings/components/Jobs/Task.js
```diff
@@ -11,34 +11,78 @@ const readOnlyFields = ["create_subtraction", "rebuild_index"];
...
+                </span>
+            </div>
+        );
+
+        const procUpperLimit = currentLimitProc < proc ? currentLimitProc : proc;
+        const memUpperLimit = currentLimitMem < mem ? currentLimitMem : mem;
+
         return (
-            <ListGroupItem allowFocus>
+            <ListGroupItem allowFocus className={this.state.error ? "list-group-item-danger" : ""}>
```
igboyes: You should be able to assign danger styling with `bsStyle="danger"`

Miminko:
I'm having trouble updating the limits through the UI
- resource limits cannot be set if they are lower than any task-specific limits
- task limits cannot be set at any valid value for some reason; they give the same "exceeds -- resource limit" error response
I'm not sure if this is something wrong in the client side logic or if the server side defaults (e.g. mem=16) are the issue.

igboyes:
> resource limits cannot be set if they are lower than any task-specific limits

This is expected. The user has to change the task-specific limits before reducing the resource limits.

> ask limits cannot be set at any valid value for some reason; they give the same "exceeds -- resource limit" error response
I'm not sure if this is something wrong in the client side logic or if the server side defaults (e.g. mem=16) are the issue.

I have fixed some server-side code that may have been causing this problem. Merge in the latest changes from `v2.2.0`. This will get rid of the merge conflicts too.

PR #533, opened 2018-04-16
UI reorganization and submenu

### LINKED ISSUE #531: User interface reorganization
**Layout**
- make main view and nav bar take user entire viewport
- make sure changes don't negatively affect rendering on mobile (as crappy as it is)

**Sidebar**
- add a basic sidebar to the left side of the main view
- add certain links in toolbars to sidebar instead:
  - jobs > resources
  - samples > files
  - viruses > indexes
  - subtraction > files
- actions buttons (eg. create, filter) stay in toolbar
- move domain-specific settings to sidebar
  - viruses
    - source types
    - internal control
  - samples
    - unique sample names
    - default sample rights
  - jobs
    - all limit settings

**Sidebar Example**

**Settings**
- the settings navbar link is going to be changed to _Administration_ and moved to the account dropdown menu
- move leftover administration tabs to sidebar

Miminko:
- resolves #531 
- added a fixed column to the left-hand side for domain-specific sidebar
- modified css container size to take up more of the viewport width
- settings tab has been moved to drop down menu in top left corner and renamed to "Administration"
- settings sub tabs are moved to sidebar 

- one problem is that when in the "/administration" tab, reloading the page results in a 404 error, but not when navigating from a different page.

### DIFF
--- a/client/src/js/nav/components/Sidebar.js
+++ b/client/src/js/nav/components/Sidebar.js
@@ -20,6 +20,8 @@ const isHomeActive = (location) => {
         activeTab = "Subtraction";
     } else if (startsWith(location.pathname, "/administration")) {
         activeTab = "Administration";
+    } else if (startsWith(location.pathname, "/hmm")) {
+        activeTab = "HMM";
     }
 
     return activeTab;
@@ -38,7 +40,7 @@ const Sidebar = (props) => {
                     <LinkContainer to="/jobs">
                         <NavItem className="sidebar-item">
                             <Icon name="menu" />
-                            <div>Jobs List</div>
+                            <div>List</div>
                         </NavItem>
                     </LinkContainer>
                     <LinkContainer to="/jobs/resources">
@@ -50,7 +52,7 @@ const Sidebar = (props) => {
                     <LinkContainer to="/jobs/settings">
                         <NavItem className="sidebar-item">
                             <Icon name="cog" />
-                            <div>Limit Settings</div>
+                            <div>Settings</div>
                         </NavItem>
                     </LinkContainer>
                 </Nav>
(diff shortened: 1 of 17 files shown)


igboyes:
This looks like a good start. Can you do the following:

- remove context word from icon labels (eg. Sample Settings -> Settings)
- make sidebar border only show on right side of div

client/src/js/nav/components/Sidebar.js
```diff
@@ -0,0 +1,184 @@
...
+};
+
+const Sidebar = (props) => {
+
+    const activeTab = isHomeActive(props.location);
+    let sidebarComponents;
+
+    switch (activeTab) {
+
+        case "Jobs":
```
igboyes: In this `switch`, you can directly return the component for each case instead of assigning it to `sidebarComponents` and returning at the end of the statement. This will also make `break` statements unnecessary.

The sidebar items are perfect candidates for being made into a reusable component.

client/src/js/samples/components/Samples.js
```diff
@@ -1,22 +1,53 @@
...
+            </strong>
+        </h3>
+        <UniqueNames />
+        <SampleRights />
     </div>
 );
 
-export default Samples;
+const Samples = (props) => {
+    let content;
```
igboyes: Drop `content` and directly `return` components.

client/src/js/viruses/components/Viruses.js
```diff
@@ -1,19 +1,50 @@
...
+            </strong>
+        </h3>
+        <SourceTypes />
+        <InternalControl />
     </div>
 );
 
-export default Viruses;
+const Viruses = (props) => {
+    let content;
```
igboyes: Direct `return` statements.

PR #738, opened 2018-05-29
UI tweaks

### LINKED ISSUE #734: Update reference edit pencil icon
- show grey lock for remote references
- show icon only when _Manage_ tab is active
- reduce size to match other icons in `ViewHeader` components

### LINKED ISSUE #735: Render created_at and user values in ViewHeader
Currently, the `created_at` date and `id` of the creating user are shown in tables in detail pages. This is hard to read and is also not implement consistently:

image

image

Standardize the rendering of the this data as small-styled text below the `<ViewHeader />` title.

It may make sense to add a margin below the `<ViewHeader />` so the page is not too cramped.

**Also remove `<PageHint />` component**

Miminko:
- resolves #734, #733

### DIFF
--- a/client/src/js/hmm/components/Installer.js
+++ b/client/src/js/hmm/components/Installer.js
@@ -83,7 +83,7 @@ const mapStateToProps = (state) => {
     return {
         ...status,
         process
-    }
+    };
 };
 
 const mapDispatchToProps = (dispatch) => ({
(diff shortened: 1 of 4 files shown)


igboyes:
This doesn't fully resolve #735. The intent was to to make display of `created_at` and `user` consistent across all situations. The logic for doing this should be embedded in `<ViewHeader />`.

We can discuss this is #735 is not clear.

PR #789, opened 2018-06-06
Add creation info to index header

### LINKED ISSUE #787: Index detail view header
Use _created at..._ subtitle in reference index detail view in place of table.

image

Miminko:
- resolves #787

### DIFF
--- a/client/src/js/indexes/components/Detail.js
+++ b/client/src/js/indexes/components/Detail.js
@@ -1,4 +1,5 @@
 import React from "react";
+import Moment from "moment";
 import { connect } from "react-redux";
 import { Switch, Redirect, Route } from "react-router-dom";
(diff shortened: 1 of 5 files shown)


igboyes:
I've noticed you are changing the rendered dates to use the `Moment.calendar()` method over the `<RelativeTime />` component.

Is there a reason for this? If there isn't I would prefer the remain consistent with the representation of dates in the rest of the application.

Miminko:
Ok, I'll switch them all to `<RelativeTime />`. 
`Moment.calendar()` was originally used to display the creation times for sample details, and I had thought that it looked nicer.

igboyes:
Then there is a reason. Just leave it as is, but please fix the capitalization problem.

Issue #825, opened 2018-06-12
Job Cancellation


Miminko:
Attempting to cancel a job while it is running results in a 
`POST http://localhost:9950/api/jobs/:jobId/cancel 404 (Not Found)` error

igboyes:
This end point accepts the `PUT` method.

PR #827, opened 2018-06-12
Job detail progress

### LINKED ISSUE #770: Job status component
Add a component to the `<JobDetail />` component that renders that current state/status of the job.

Something like this comes to mind: http://react-component.github.io/steps/.

Miminko:
- resolves #770 
- adds vertical list of status steps to job detail view to replace progress bar
- fixes job cancellation request #825 and icon

### DIFF
--- a/client/src/js/jobs/api.js
+++ b/client/src/js/jobs/api.js
@@ -14,7 +14,7 @@ export const get = ({ jobId }) => (
 );
 
 export const cancel = ({ jobId }) => (
-    Request.post(`/api/jobs/${jobId}/cancel`)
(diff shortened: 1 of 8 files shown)


igboyes:
This looks great. There are a few changes that would make it better I think.

Having all of the text below the progress diagram is a bit jarring as the job proceeds. For now do this:
- move started by and at information to `<ViewHeader />` like in other contexts
- get rid of _Unique ID` row
- put core and mem info below `<ViewHeader />` maybe in format **2 CPUs / 4 GB**
- move task_args `<Table />` above progress diagram

PR #849, opened 2018-06-13
Update reference creation modals

### LINKED ISSUE #826: New reference modal form changes
- _Organism_ input should be text not select
- _Organism_ and _Data Type_ should be automatically assigned from source when cloning (users can change those values)
- _Data Type_ is required so make the default option `genome`
- `None` should not be an option for _Data Type_

Miminko:
- resolves #826


igboyes:
Great. There is only one more issue I think:
- In the _Data Type_ input the option and value are capitalized differently (ie. Genome vs genome)

igboyes:
Looks perfect.

Issue #994, opened 2018-07-11
Coverage data management on backend


igboyes:
- calculate and store cropped alternate of data
- look into delta-compression to reduce size of payload and database document
- store coverage data as coordinates rather than `list(int)`

igboyes:
Will be covered in #974.

PR #996, opened 2018-07-11
Pathoscope table

### LINKED ISSUE #984: Percent reads mapped
In the analysis tab, values are given for "Library Read Count" and "Mapped Reads". Could we also include a line for "Percent Reads Mapped"?

igboyes:
- resolves #984
- combine `user` and `created_at` display in one table row
- improve coordinate filling implementation
- move most data formatting `state` from analysis components to Redux
- remove IQR cropping of depth charts (this will be done on the server #994)
- return the `reference.name` value for an analysis from the API
- show mapped read percentage
- show the maximum mean/median depth instead of `maxDepth`
- show the reference name and index version in analysis list item
- show the reference name and index version in detail table in Pathoscope view


igboyes:
@Miminko, I can't get the client tests to succeed even after updating the snapshots that were not matching.

Any idea what I'm doing wrong?

Miminko:
@igboyes 
Did you try manually updating the files to match the snapshot or using the command line to update the snapshots to match the files?

I think `yarn test -u` should overwrite the snapshots to match the current test files.
Committing the snapshots to the PR should then enable the tests to pass

igboyes:
Yes, I did that. I think it was due to out-of-date testing dependencies. I've updated the snapshots with the latest dependencies.

I wish there was a warning when your `node_modules` doesn't match the `yarn.lock` file...

PR #1240, opened 2018-12-13
Fix 1222

### LINKED ISSUE #1222: Wrong noun used in sample deletion modal
image

It should be _sample_ instead of _OTU_ here.

igboyes:
- resolves #1222 
- change sample remove modal noun to 'sample' instead of typo 'OTU'


igboyes:
Closed due to Drone issue.

PR #1577, opened 2019-12-03
Use throttle in some reference sample sagas

### LINKED ISSUE #1569: Debounce official reference installation request
Prevent rapid submission of installation requests that can occur on a rapid double click.

jinxuchen:
- resolve #1569

### DIFF
--- a/client/src/js/references/sagas.js
+++ b/client/src/js/references/sagas.js
@@ -100,14 +100,14 @@ export function* updateRemoteReference(action) {
 }
 
 export function* watchReferences() {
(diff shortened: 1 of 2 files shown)


client/src/js/samples/components/Create/Create.js
```diff
@@ -108,6 +108,7 @@ class CreateSample extends React.Component {
     };
 
     render() {
+        this.state.subtraction = getActiveSubtraction(this.props);
```
igboyes: This should not be in this commit/PR. Also see https://reactjs.org/docs/state-and-lifecycle.html#using-state-correctly

PR #1726, opened 2020-02-19
Add NoneFound component

### LINKED ISSUE #1720: Show NoneFound when no targets created for ref
image

- add `BoxGroupSection` with `NoneFound` inside

jinxuchen:
- resolve #1720 
- show `NoneFound` component when no reference targets defined

### DIFF
--- a/client/src/js/references/components/Detail/Targets/Targets.js
+++ b/client/src/js/references/components/Detail/Targets/Targets.js
@@ -2,7 +2,7 @@ import { map, reject } from "lodash-es";
 import React from "react";
 import { connect } from "react-redux";
 import styled from "styled-components";
-import { BoxGroup, BoxGroupHeader } from "../../../../base";
+import { BoxGroup, BoxGroupHeader, NoneFoundSection } from "../../../../base";
 import { checkRefRight } from "../../../../utils/utils";
 import { editReference } from "../../../actions";
 import AddTarget from "./Add";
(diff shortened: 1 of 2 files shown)


client/src/js/references/components/Detail/Targets/Targets.js
```diff
@@ -78,6 +78,12 @@ export class Targets extends React.Component {
             );
         }
 
+        const none = (
+            <BoxGroupSection>
```
igboyes: - use `NoneFoundSection` base component
- need to provide `noun` prop ('targets' should be value)
- why define `none` when you aren't going to render it? This is the better way to do it:
```js
let noneFound

if (!targetComponents.length) {
  noneFound = <NoneFoundSection noun="targets">
}
```

client/src/js/references/components/Detail/Targets/Targets.js
```diff
@@ -88,7 +94,7 @@ export class Targets extends React.Component {
                     <p>Manage the allowable sequence targets for this barcode reference.</p>
                 </TargetsHeader>
 
-                <div>{targetComponents}</div>
+                <div>{targetComponents.length ? targetComponents : none}</div>
```
igboyes: This can be changed to the below block when the previous comment is fixed:
```js
{targetComponents}
{noneFound}
```
There is no need for an extra wrapping `div` here.

PR #1730, opened 2020-02-20
Add condition for checking if segment exist

### LINKED ISSUE #1724: Error in RemoveSegment modal
image

- occurs when no segments are defined in schema tab
- probably need to handle case where no segments are defined and thus `activeName` cannot be derived

jinxuchen:
- resolve #1724
- fix unhandled client error in schema view when no segments are defined
- add condition for checking if segments exist

### DIFF
--- a/client/src/js/otus/components/Detail/Schema.js
+++ b/client/src/js/otus/components/Detail/Schema.js
@@ -162,9 +162,7 @@ class Schema extends React.Component {
                         onHide={this.handleClose}
                         onSubmit={this.handleSubmit}
                     />
-                ) : (
-                    ""
-                )}
+                ) : null}
             </div>
         );
     }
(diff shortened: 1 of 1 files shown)


client/src/js/otus/components/Detail/Schema.js
```diff
@@ -148,20 +147,24 @@ class Schema extends React.Component {
...
+
+                {segArray.length ? (
+                    <RemoveSegment
+                        activeName={this.state.selected.name}
+                        show={this.state.showRemove}
+                        onHide={this.handleClose}
+                        onSubmit={this.handleSubmit}
+                    />
+                ) : (
+                    ""
```
igboyes: React convention for this is using `null` for non-render instead of `""`.

PR #1769, opened 2020-03-30
Reimplement buttons

### LINKED ISSUE #1754: Reimplement Buttons
#### General
- base on latest `release-4.0.0` with update theme values
- use `LinkButton` when the button follows a link instead of taking an `onClick` prop
- improve code reuse between `LinkButton` and new `Button` components
- consider naming new `Button` component `Button2` while replacing old `Button` usages; we can clean this up when the PR is merged

#### Remove Props
Remove the following props and their usages:
- `block` *
- `bsSize`
- `bsStyle` (replace with `color` prop)
- `iconStyle` (not used anywhere)
- `pad` *
- `pullRight` *
- `style` *

*=Individual usages of these props should be replaced by styling separately using `styled-components` depending on the situation. An example would be replacing a usage of `pullRight` with putting the button in a flex parent

#### Colors
- The colors in the theme are pulled from https://tailwindcss.com/docs/customizing-colors
- Maintain button constrast similarly to how it is done in Bootstrap (dark buttons have white text)
- Replace `bsStyle` with `color` prop. Get the colors from the `styled-components` theme
- Use functions for getting theme values to keep code readable and reduce repetition:
  ```js
  // This code will make sense if you look at src/js/app/theme.js and review the styled-components theming docs.
  const getButtonBackgroundColor = ({ color, theme }) => get(theme, [color, `${color}Light`]);

  const Button = styled(Button)`
    background: ${getButtonBackgroundColor};
  `;
  ```

jinxuchen:
- resolve #1754 
- reimplement Button component
- update tests


igboyes:
I merged `release-4.0.0` into this PR. Please **do not** force push again into this PR. Add new commits.

I've noticed a number of visual glitches with the buttons. For example:
image

The appearance of the regular white/grey buttons has consistency issues. I'll review when this is fixed. Thanks!

PR #1922, opened 2020-10-13
Add task for generating FASTA files

### LINKED ISSUE #1882: Add task for generating FASTA files for existing subtractions
# Initial Work
- add framework for spawning tasks on application start (look at `virtool.startup`; could add `init_tasks` function here)

# Subtraction File Task
Older subtractions did not have saved `.fa.gz` files at `<data_path>/subtractions/:id`. We want to support downloading the subtraction `.fa.gz` for these older subtractions.

- add new task for creating subtraction files for all subtractions
- spawn task on app start if there are subtractions without `has_file` set to `True`
- add `has_file` to subtraction `PROJECTION`
- only allow one process to be used for this (no multi-threading)
- process on subtraction at a time to minimize resource use
- steps
  - check subtraction directories for files and set `has_file` to boolean based on whether `.fa.gz`  exists
  - for each subtraction with a missing file:
    - use `bowtie-inspect` to generate FASTA file from `.bt2` files (use [asyncio subprocess](https://docs.python.org/3/library/asyncio-subprocess.html))
    - gzip file (use `compress_file` and `decompress_file` and `app["run_in_thread"]`)
    - set `has_file` set `True` when complete

tianshengsui:
resolve #1882

### DIFF
--- a/virtool/startup.py
+++ b/virtool/startup.py
@@ -365,6 +365,9 @@ async def init_version(app: typing.Union[dict, aiohttp.web.Application]):
 
 
 async def init_tasks(app: aiohttp.web.Application):
+    if app["config"].get("no_check_db"):
+        return logger.info("Skipping subtraction FASTA files checks")
+
     db = app["db"]
     logger.info("Checking subtraction FASTA files")
 
--- a/virtool/startup.py
+++ b/virtool/startup.py
@@ -30,6 +30,9 @@ import virtool.routes
 import virtool.sentry
 import virtool.settings.db
 import virtool.software.db
+import virtool.subtractions.utils
+import virtool.subtractions.db
+import virtool.tasks.db
 import virtool.utils
 import virtool.version
 
@@ -359,3 +362,24 @@ async def init_version(app: typing.Union[dict, aiohttp.web.Application]):
     logger.info(f"Mode: {app['mode']}")
 
     app["version"] = version
+
+
+async def init_tasks(app: aiohttp.web.Application):
+    db = app["db"]
+    settings = app["settings"]
+    logger.info("Checking subtraction files")
+
+    subtraction_without_file = await virtool.subtractions.utils.check_subtraction_file(app)
+    scheduler = get_scheduler_from_app(app)
+
+    if subtraction_without_file:
+        for subtraction_id in subtraction_without_file:
+            context = {
+                "subtraction_id": subtraction_id,
+                "path": os.path.join(settings["data_path"], "subtractions", subtraction_id)
+            }
+
+            task = await virtool.tasks.db.register(db, "create_subtraction", context=context)
+            t = virtool.subtractions.db.CreateSubtractionTask(app, task["id"])
+
+            await scheduler.spawn(t.run())
--- a/virtool/startup.py
+++ b/virtool/startup.py
@@ -366,20 +366,11 @@ async def init_version(app: typing.Union[dict, aiohttp.web.Application]):
 
 async def init_tasks(app: aiohttp.web.Application):
     db = app["db"]
-    settings = app["settings"]
-    logger.info("Checking subtraction files")
+    logger.info("Checking subtraction FASTA files")
 
-    subtraction_without_file = await virtool.subtractions.utils.check_subtraction_file(app)
     scheduler = get_scheduler_from_app(app)
 
-    if subtraction_without_file:
-        for subtraction_id in subtraction_without_file:
-            context = {
-                "subtraction_id": subtraction_id,
-                "path": os.path.join(settings["data_path"], "subtractions", subtraction_id)
-            }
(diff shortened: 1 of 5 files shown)


virtool/subtractions/utils.py
```diff
@@ -45,3 +46,26 @@ def join_subtraction_index_path(settings: dict, subtraction_id: str) -> str:
         join_subtraction_path(settings, subtraction_id),
         "reference"
     )
+
+
+async def check_subtraction_file(app):
```
igboyes, line 51: Rename to `check_subtraction_fasta_files()`.

virtool/subtractions/utils.py
```diff
@@ -45,3 +46,26 @@ def join_subtraction_index_path(settings: dict, subtraction_id: str) -> str:
...
+    db = app["db"]
+    settings = app["settings"]
+    cursor = db.subtraction.find()
+    subtraction_without_file = []
```
igboyes, line 55: Rename to `subtractions_without_fasta`

virtool/subtractions/utils.py
```diff
@@ -45,3 +46,26 @@ def join_subtraction_index_path(settings: dict, subtraction_id: str) -> str:
...
+
+    async for subtraction in cursor:
```
igboyes, line 57:
```python
async for subtraction in db.subtraction.find():
```

virtool/subtractions/utils.py
```diff
@@ -45,3 +46,26 @@ def join_subtraction_index_path(settings: dict, subtraction_id: str) -> str:
...
+        path = join_subtraction_path(settings, subtraction["_id"])
+
+        if not glob.glob(f'{path}/*.fa.gz'):
```
igboyes, line 60: This is cleaner:
```python
has_file = True

if not glob.glob(f'{path}/*.fa.gz'):
  has_file = False
  subtraction_without_file.append(subtraction["_id"])\
```

virtool/subtractions/db.py
```diff
@@ -12,10 +16,69 @@
     "job",
     "name",
     "nickname",
-    "user"
+    "user",
+    "has_file"
 ]
 
 
+class CreateSubtractionTask(virtool.tasks.task.Task):
```
igboyes, line 24: Change to `WriteSubtractionFASTATask` and `write_subtraction_fasta`.

Currently, this can cause confusion with the job called `CreateSubtraction`

virtool/startup.py
```diff
@@ -359,3 +362,24 @@ async def init_version(app: typing.Union[dict, aiohttp.web.Application]):
...
+async def init_tasks(app: aiohttp.web.Application):
+    db = app["db"]
+    settings = app["settings"]
+    logger.info("Checking subtraction files")
+
+    subtraction_without_file = await virtool.subtractions.utils.check_subtraction_file(app)
+    scheduler = get_scheduler_from_app(app)
+
+    if subtraction_without_file:
+        for subtraction_id in subtraction_without_file:
```
igboyes, line 376: I want a single task to go through each subtraction sequentially. Start a separate task for each subtraction could cause resource issues.

igboyes (changes requested):
Naming changes for accuracy and to avoid confusion with other names in application.

The main functional change is to move the looping through each subtraction with missing files to the task itself. Only one task should run to generate FASTA files for all subtractions. It could be a two step task:

1. Get list of subtraction IDs where FASTA should be generated
2. Generate FASTA files for all IDs.

igboyes:
@tianshengsui  Do you understand why those seemingly unrelated tests failed and you had to updated their snapshots?

tianshengsui:
> @tianshengsui Do you understand why those seemingly unrelated tests failed and you had to updated their snapshots?

Yes. Because the ```write_subtraction_fasta``` task also generated a random ID when application starts, which makes the index of ```test_random_alphanumeric.history``` no longer correspond.

virtool/startup.py
```diff
@@ -359,3 +362,15 @@ async def init_version(app: typing.Union[dict, aiohttp.web.Application]):
     logger.info(f"Mode: {app['mode']}")
 
     app["version"] = version
+
+
+async def init_tasks(app: aiohttp.web.Application):
+    db = app["db"]
```
igboyes, line 368: Don't run the task if the `no-check-db` option is set. This will prevent the ID clash in the tests as the test server has the option set by default.

See here for example of how this is used to skip DB checks in `virtool.startup`: https://github.com/virtool/virtool/blob/031c1e75ca29ddf5f1e34039ab035aab608c65dd/virtool/startup.py#L62

igboyes (changes requested):
Please reset to 5bb40de13eed7ebd887fc97c5c279b4124996321 to remove the changes to the snapshot files.

Then address requested changed related to `no_db_checks`. This should allow tests to pass again.

PR #1927, opened 2020-10-14
Rewrite install task

### LINKED ISSUE #1909: Use Task class to rewrite some function-based tasks
- [ ] `virtool.software.db.install()`
- [ ] `virtool.hmm.db.install()`

tianshengsui:
resolve #1909

### DIFF
--- a/virtool/hmm/db.py
+++ b/virtool/hmm/db.py
@@ -38,123 +38,148 @@ PROJECTION = [
 ]
 
 
-class HmmInstallTask(virtool.tasks.task.Task):
+class HMMInstallTask(virtool.tasks.task.Task):
+    """
+    Runs a background Task that:
+        - downloads the official profiles.hmm.gz file
+        - decompresses the vthmm.tar.gz file
+        - moves the file to the correct data path
+        - downloads the official annotations.json.gz file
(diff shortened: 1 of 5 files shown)


virtool/hmm/db.py
```diff
@@ -38,6 +38,125 @@
 ]
 
 
+class HmmInstallTask(virtool.tasks.task.Task):
```
igboyes, line 41: Rename to `HMMInstallTask`.

igboyes (changes requested):
One rename.

Why didn't you split the task functions into separate steps when moving to `Task`?

tianshengsui:
> One rename.
> 
> Why didn't you split the task functions into separate steps when moving to `Task`?

Because all the code are under the context of ```tempdir```. If I split them into several functions, I think the ```tempdir``` will be closed and removed.

igboyes:
Okay. That is what I thought.

Can you add the tempdir generation and cleanup to the base `Task` class and make `temp_dir` an attribute on that class? That way the temporary directory is available to all the tasks but is still automatically deleted at the end.

tianshengsui:
I will do it.

igboyes (approved):

PR #1944, opened 2020-10-22
Add prefix check for Redis connection string

### LINKED ISSUE #1942: FileNotFoundError: [Errno 2] No such file or directory While Starting Server or Runner
When starting up the server or runner using the command:
```
python run.py server --redis-connection-string 127.0.0.1:6379
``` 

The error was bypassed by adding `redis://` before the IP.

The stack trace for the error is:

tianshengsui:
resolve #1942

### DIFF
--- a/virtool/redis.py
+++ b/virtool/redis.py
@@ -8,9 +8,13 @@ import aioredis
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


igboyes (approved):
Thanks for fixing spelling error as well.

PR #1947, opened 2020-10-23
Use task to create JSON files for indexes

### LINKED ISSUE #1148: Use task to create JSON files for all latest indexes
Similar to creating FASTA files for subtractions.

Currently, `json.gz` export files for references are created on the fly. **The goal here is to create files for all reference indexes in advance**.

- [ ] see `virtool.reference.db.export` and `virtool.downloads.api.download_reference` for on-the-fly reference exporting currently works

- [ ] add `has_json` field set to `True` on index document when JSON has been successfully generated for an index

- [ ] store `reference.json.gz` file in `<data_path>/references/:ref_id/:index_id`

- [ ] share code for creating `json.gz` file with that used in the _BuildIndex_ job as described in #1214.

tianshengsui:
resolve #1148


virtool/references/db.py
```diff
@@ -433,6 +435,83 @@ async def update_reference(self):
...
+
+    async def check_index_json_files(self):
+        db = self.db
+        settings = self.app["settings"]
+
+        async for index in db.indexes.find():
+            index_id = index["_id"]
+            ref_id = index["reference"]["id"]
+            index_path = os.path.join(settings["data_path"], "references", ref_id, index_id)
+            has_json = True
```
igboyes, line 458: I think it is safer `has_json` to only be set at line 508.

If this task as aborted and  the `json.gz` file is only partially written, the next time it runs the partial file will be assumed to be a complete file just because it exists at the expected path.

I think it is safer to check for `has_json` in the DB and overwrite and existing file if `has_json` is not defined or `False`. You will also save time not checking against the filesystem for every index.

What do you think?

igboyes (changes requested):
Thanks. Respond to question in review. Otherwise, looks good!

virtool/references/db.py
tianshengsui: It makes sense! I didn't think of that point. I will fix it. Thanks!

virtool/references/db.py
```diff
@@ -433,6 +434,70 @@ async def update_reference(self):
...
+        super().__init__(app, task_id)
+
+        self.steps = [
+            self.check_index_json_files,
+            self.create_index_json_files
+        ]
+
+        self.index_without_json = []
+
+    async def check_index_json_files(self):
```
igboyes, line 449: Drop this method and move to `create_index_json_files`. You can just use a Mongo query to replace lines 452-454:
```python
db.indexes.find({"has_json": {"$ne": True}})
```

igboyes (changes requested):
Sorry. One more tweak to streamline things.

virtool/references/db.py
igboyes: For reference: https://docs.mongodb.com/manual/reference/operator/query/

igboyes (approved):

igboyes (approved):
Thanks!

PR #1953, opened 2020-10-26
Add  build_json step to the BuildIndex job

### LINKED ISSUE #1214: Write reference.json.gz during index build
- [ ] Add a _Build JSON_ (`build_json`) step to the `BuildIndex` job that creates a `reference.json.gz` file in the index directory:`<data_path>/references/:ref_id/:index_id`.

- [ ] See `virtool.reference.db.export` and `virtool.downloads.api.download_reference` for on-the-fly reference exporting currently works

This will happen for all new index builds. Active indexes (most recent for a reference) without `json.gz` files will have them generated in a task. See #1148.

- [ ] Share code for creating `json.gz` file with that used in the task described in #1148.

- [ ] Set `has_json` to `False` when build starts and set it to `True` when build is complete

tianshengsui:
resolve #1214

### DIFF
--- a/virtool/jobs/build_index.py
+++ b/virtool/jobs/build_index.py
@@ -48,7 +48,7 @@ async def check_db(job):
 
 async def mk_index_dir(job):
     """
-    Make dir for the new index at ``<data_path/references/<index_id>``.
+    Make dir for the new index at ``<data_path>/references/<index_id>``.
 
     """
     await job.run_in_executor(
@@ -158,6 +158,10 @@ async def upload(job):
 
 
 async def build_json(job):
+    """
+    Create a reference.json.gz file at ``<data_path>/references/<ref_id>/<index_id>``.
+
+    """
     await job.db.indexes.find_one_and_update({"_id": job.params["index_id"]}, {
         "$set": {
             "has_json": False
--- a/virtool/jobs/build_index.py
+++ b/virtool/jobs/build_index.py
@@ -1,10 +1,12 @@
 import asyncio.tasks
+import json
 import os
 import shutil
 import typing
 
 import aiofiles
 
+import virtool.api.json
 import virtool.errors
 import virtool.history.db
 import virtool.history.utils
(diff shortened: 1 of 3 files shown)


virtool/utils.py
```diff
@@ -96,6 +96,11 @@ def compress_file_with_pigz(path, target, processes):
         subprocess.call(command, stdout=f)
 
 
+def compress_json_with_gzip(json_string, target):
```
igboyes, line 99: Please add a docstring and typing to this function.

virtool/jobs/build_index.py
```diff
@@ -154,6 +157,59 @@ async def upload(job):
         })
 
 
+async def build_json(job):
+    """
+    Create a reference.json.gz file at ``<data_path>/references/<ref_id>/<index_id>``.
+
+    """
+    await job.db.indexes.find_one_and_update({"_id": job.params["index_id"]}, {
```
igboyes, line 165: Move this to when an index document is first created. This is in `virtool.indexes.api`.

virtool/jobs/build_index.py
```diff
@@ -154,6 +157,59 @@ async def upload(job):
...
+    })
+
+    document = await job.db.references.find_one(job.params["ref_id"], ["data_type", "organism", "targets"])
+
+    app_dict = {
+        "db": job.db,
+        "settings": job.settings
+    }
+
+    otu_list = await virtool.references.db.export(
```
igboyes, line 178: Change signature of `virtool.references.db.export` to `def export(db, settings, ref_id)` to avoid creating `app_dict`.

Update usage in `virtool.downloads.api.download_reference`.

igboyes (changes requested):
Thanks! Minor changes.

virtool/jobs/build_index.py
tianshengsui: I think if `export` function has changed, we also need to change the signature of `virtool.history.db.patch_to_version` because it has been called in `export` function and it also needs the `app` parameter.

Do you want this to be changed as well?

igboyes: Ah. What a pain. Leave it as is then.

igboyes:
@tianshengsui I removed one review comment. Only two left to deal with. Thanks.

PR #2008, opened 2020-12-08
Add migrate function for subtraction list

### LINKED ISSUE #2003: Migrate existing analyses to subtractions list
For existing analyses, subtraction information is stored at `subtraction.id`:
```json
{
  "subtraction": {
    "id": "foo"
  }
}
```

Add migration function, `change_to_subtractions_list()` in `virtool.analyses.migrate` to transform this field to a list:
```json
{
  "subtractions": [
    "foo"
  ]
}
```
This function should be called in `migrate_analyses()`. This means the function will automatically be called on server start.
- [x] add `subtractions` field
- [x] remove the old `subtraction` field
- [x] write tests

tianshengsui:
resolve #2003

### DIFF
--- a/virtool/analyses/migrate.py
+++ b/virtool/analyses/migrate.py
@@ -11,3 +11,18 @@ async def migrate_analyses(app: virtool.types.App):
 
     """
     await virtool.db.utils.delete_unready(app["db"].analyses)
+    await change_to_subtractions_list(app["db"])
+
+
+async def change_to_subtractions_list(db):
+    async for document in db.analyses.find():
+        if "subtraction" in document:
+            await db.analyses.update_one({"_id": document["_id"]}, {
+                "$set": {
+                    "subtractions": [document["subtraction"]["id"]]
+                },
+                "$unset": {
+                    "subtraction": ""
+                }
+            })
+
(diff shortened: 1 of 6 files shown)


virtool/analyses/migrate.py
```diff
@@ -11,3 +11,18 @@ async def migrate_analyses(app: virtool.types.App):
...
+    async for document in db.analyses.find():
+        if "subtraction" in document:
+            await db.analyses.update_one({"_id": document["_id"]}, {
+                "$set": {
+                    "subtractions": [document["subtraction"]["id"]]
+                },
+                "$unset": {
+                    "subtraction": ""
+                }
+            })
```
igboyes, line 27: This will be very slow for large analyses collections. You can use a query instead of the conditional at line 19:
```python
async for document in db.analyses.find({"subtraction": {"$exists": True}}):
```

Using [bulk operations](https://motor.readthedocs.io/en/stable/examples/bulk.html) would also greatly speed this up.

igboyes (changes requested):
Thanks. Needs some optimization of update code.

igboyes (approved):

PR #2010, opened 2020-12-09
Add function to attach subtraction list

### LINKED ISSUE #2006: Add function to join subtraction data for subtraction lists
Call function `virtool.subtractions.db.attach_subtractions()`.

When analysis data is returned in a response from the API server, the subtraction list in the analysis document will be expanded to a list of objects that more fully describe the subtractions.

The starting list field in the document:
```json
"subtractions": [
  "foo",
  "bar"
]
```

The field should look like this after calling `attach_subtractions()`:
```json
"subtractions": [
  {
    "id": "foo",
    "name": "Foo"
  },
  {
    "id": "bar",
    "name": "Bar"
  }
]
```

- [x] add function
- [x] write complete tests
- [x] replace usage of `virtool.subtractions.db.attach_subtraction()` in request handler (`virtool.analyses.api.get()`).

See `virtool.subtractions.db.attach_subtraction()` for basic idea.

tianshengsui:
resolve #2006

### DIFF
--- a/tests/analyses/snapshots/snap_test_api.py
+++ b/tests/analyses/snapshots/snap_test_api.py
@@ -16,10 +16,16 @@ snapshots['test_get[uvloop-None-True] format_analysis'] = {
     'sample': {
         'id': 'baz'
     },
-    'subtraction': {
-        'id': 'plum',
-        'name': 'Plum'
-    },
+    'subtractions': [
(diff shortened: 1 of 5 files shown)


igboyes (approved):
Great. Thanks!

PR #2014, opened 2020-12-15
Update label edit endpoint

### LINKED ISSUE #1990: OperationFailure: '$set' is empty. You must specify a field like so: {$set: {<field>: ...}}
Sentry Issue: [BACKEND-15G](https://sentry.io/organizations/cph/issues/2055583871/?referrer=github_integration)

```
OperationFailure: '$set' is empty. You must specify a field like so: {$set: {<field>: ...}}
(16 additional frame(s) were not displayed)
...
  File "pymongo/collection.py", line 2873, in _find_and_modify
    out = self._command(sock_info, cmd,
  File "pymongo/collection.py", line 235, in _command
    return sock_info.command(
  File "pymongo/pool.py", line 603, in command
    return command(self.sock, dbname, spec, slave_ok,
  File "pymongo/network.py", line 165, in command
    helpers._check_command_response(
  File "pymongo/helpers.py", line 159, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)
```
Occuring at `PATCH /api/labels/:id`. If the request body contains no key-value pairs, this should probably return a `422` error. Please check and reply in issue.

tianshengsui:
resolve #1990

### DIFF
--- a/virtool/labels/api.py
+++ b/virtool/labels/api.py
@@ -2,7 +2,7 @@ import virtool.http.routes
 import virtool.utils
 import virtool.validators
 import virtool.db.utils
-from virtool.api.response import bad_request, json_response, no_content, not_found
+from virtool.api.response import bad_request, empty_request, json_response, no_content, not_found
 
 routes = virtool.http.routes.Routes()
 
@@ -10,7 +10,7 @@ routes = virtool.http.routes.Routes()
 @routes.get("/api/labels")
 async def find(req):
     """
-        Get a list of all label documents in the database.
+    Get a list of all label documents in the database.
 
     """
     db = req.app["db"]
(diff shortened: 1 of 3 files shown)


virtool/labels/api.py
```diff
@@ -99,14 +99,17 @@ async def create(req):
...
+    Edit an existing label.
 
     """
     db = req.app["db"]
     data = req["data"]
 
     label_id = req.match_info["label_id"]
 
+    if not data:
+        return empty_request("The request body is empty")
```
igboyes, line 111: Since you already have a similar default message, should this just be:
```python
return empty_request()
```

igboyes (changes requested):

igboyes (approved):

PR #2018, opened 2020-12-18
Support subtraction list at sample creation endpoint

tianshengsui:
Resolve [Story #50](https://app.clubhouse.io/virtool/story/50/allow-zero-or-more-subtractions-at-sample-creation-api-endpoint)

### DIFF
--- a/virtool/db/utils.py
+++ b/virtool/db/utils.py
@@ -36,6 +36,27 @@ def apply_projection(document, projection):
     return {key: document[key] for key in document if projection.get(key, False)}
 
 
+async def check_missing_ids(collection, id_list, query=None):
+    """
+    Check if all IDs in the ``id_list`` exist in the database.
+
+    :param collection: the Mongo collection to check ``id_list`` against
+    :type collection: :class:`motor.motor_asyncio.AsyncIOMotorCollection`
+
+    :param id_list: the IDs to check for
+    :type: id_list: list
+
+    :param query: a MongoDB query
+    :type: query: dict
+
+    :return: all non-existent IDs
+    :rtype: set
+
+    """
+    existent_ids = await collection.distinct("_id", query)
+    return set(id_list) - set(existent_ids)
+
+
 async def get_new_id(collection, excluded=None):
     """
     Returns a new, unique, id that can be used for inserting a new document. Will not return any id that is included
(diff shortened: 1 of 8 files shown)


virtool/samples/api.py
```diff
@@ -226,9 +226,10 @@ async def create(req):
     if name_error_message:
         return bad_request(name_error_message)
 
-    # Make sure a subtraction host was submitted and it exists.
-    if not await db.subtraction.count_documents({"_id": data["subtraction"], "is_host": True}):
-        return bad_request("Subtraction does not exist")
+    # Make sure each subtraction host was submitted and it exists.
+    for subtraction in data["subtractions"]:
```
igboyes, line 230: This can be done more efficiently with an `$in` operator and `db.subtraction.distinct()`. Compare the subtraction ID lists from the database and the API request by using the [`set`](https://docs.python.org/3/library/stdtypes.html#set) data structure.

You should return all invalid subtraction IDs in the response instead of just the first one.

Since this same check is done for analyses, you should reuse the code. Maybe add a new function: ` virtool.db.utils.check_missing_ids(id_list)`.

igboyes (changes requested):
Please also add a migration function like you did for analyses. Add tests as well. Thanks.

virtool/samples/migrate.py
```diff
@@ -33,3 +34,15 @@ async def recalculate_all_workflow_tags(db):
     for chunk in virtool.utils.chunk_list(sample_ids, 50):
         coros = [virtool.samples.db.recalculate_workflow_tags(db, sample_id) for sample_id in chunk]
         await asyncio.gather(*coros)
+
+
+async def change_to_subtractions_list(db):
```
igboyes, line 39: Please add docstring for this function

virtool/db/utils.py
```diff
@@ -36,6 +36,27 @@ def apply_projection(document, projection):
     return {key: document[key] for key in document if projection.get(key, False)}
 
 
+async def check_missing_ids(collection, id_list, query=None):
```
igboyes, line 39: 1. Use typing in function signature instead of docstring
2. Add tests as needed.

igboyes (changes requested):

igboyes (approved):

Issue #26, opened 2017-02-08
Required enhancements to the file manager


igboyes:
There are some things the file manager currently does not do that it needs to:

- remove untracked files from the `files` data path
- remove database documents for files that have been manually removed from the `files` data path
- remove tracked files that have expired

igboyes:
Resole in #96

PR #2021, opened 2020-12-21
Add endpoints for K8S health check

tianshengsui:
Related [story #26](https://app.clubhouse.io/virtool/story/26/kubernetes-liveness-and-readiness-checks)

### DIFF
--- a/tests/health/test_api.py
+++ b/tests/health/test_api.py
@@ -1,0 +1,20 @@
+async def test_is_alive(spawn_client):
+    client = await spawn_client(authorize=True, administrator=True)
+
+    resp = await client.get("/api/health/alive")
+
+    assert resp.status == 200
+    assert await resp.json() == {
+        "alive": True
+    }
+
+
+async def test_is_ready(spawn_client):
+    client = await spawn_client(authorize=True, administrator=True)
+
+    resp = await client.get("/api/health/ready")
+
+    assert resp.status == 200
+    assert await resp.json() == {
+        "ready": True
+    }
--- a/tests/health/test_api.py
+++ b/tests/health/test_api.py
@@ -1,3 +1,6 @@
+import pytest
+
+
 async def test_is_alive(spawn_client):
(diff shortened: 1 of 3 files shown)


igboyes (changes requested):
Place return simple JSON objects and add API tests for these. Example:
```python
{
  "alive": True
}
```

We may want to expand this response and I'd rather keep the content type JSON.

BlakeASmith:
Are there going to be situations in which the Virtool API is alive but not ready? 

If not then we could get away with having only the /api/health/live endpoint. If this does occur 
then we should add some mechanism to change the servers "readiness" status within the code.

BlakeASmith:
> Are there going to be situations in which the Virtool API is alive but not ready?
> 
> If not then we could get away with having only the /api/health/live endpoint. If this does occur
> then we should add some mechanism to change the servers "readiness" status within the code.

You could store a "ready" variable in the `Application` object's config as described [here](https://docs.aiohttp.org/en/stable/web_advanced.html#application-s-config). Then change the `/api/health/ready` route handler to somethinglike;

```python
async def handler(request):
	if request.config_dict["ready"]:
		# report that the server is ready (HTTP 200)
		...
	else:
		# report that the server is not ready (HTTP 500, or another status code)
```

tianshengsui:
@BlakeASmith This looks good. I will do it.

BlakeASmith (approved):

igboyes:
The `ready` value is never actually set and will always be `True`.

After your looking into this, I am not sure if it makes sense to merge. I think the `/api` endpoint could be used for both readiness and liveness. Agreed @tianshengsui and @BlakeASmith?

I will close for now if so.

tianshengsui:
> The `ready` value is never actually set and will always be `True`.
> 
> After your looking into this, I am not sure if it makes sense to merge. I think the `/api` endpoint could be used for both readiness and liveness. Agreed @tianshengsui and @BlakeASmith?
> 
> I will close for now if so.

Yes, I agree.

PR #2022, opened 2021-01-05
Postgres setup

tianshengsui:
- Add Postgres connection on application start
- Related story: [17](https://app.clubhouse.io/virtool/story/17/postgres-connection-on-application-start)

### DIFF
--- a/virtool/postgres.py
+++ b/virtool/postgres.py
@@ -13,9 +13,19 @@ async def connect(postgres_connection_string: str) -> sqlalchemy.engine:
         sys.exit(1)
 
     try:
-        postgres = create_engine(postgres_connection_string, echo=True)
+        postgres = create_engine(postgres_connection_string)
+        await check_version(postgres)
 
         return postgres.connect()
     except ConnectionRefusedError:
         logger.fatal("Could not connect to PostgreSQL: Connection refused")
         sys.exit(1)
+
+
+async def check_version(postgres):
+    with postgres.connect() as con:
+        info = con.execute('SHOW server_version').fetchone()
+
+    version = info[0].split()[0]
+    logger.info(f"Found PostgreSQL {version}")
+    return
--- a/virtool/postgres.py
+++ b/virtool/postgres.py
@@ -1,0 +1,21 @@
+import logging
+import sys
+
+import sqlalchemy
+from sqlalchemy import create_engine
+
+logger = logging.getLogger(__name__)
+
+
+async def connect(postgres_connection_string: str) -> sqlalchemy.engine:
+    if not postgres_connection_string.startswith("postgresql://"):
+        logger.fatal("Invalid PostgreSQL connection string")
+        sys.exit(1)
+
+    try:
+        postgres = create_engine(postgres_connection_string, echo=True)
+
+        return postgres.connect()
+    except ConnectionRefusedError:
+        logger.fatal("Could not connect to PostgreSQL: Connection refused")
+        sys.exit(1)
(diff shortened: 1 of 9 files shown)


virtool/postgres.py
```diff
@@ -0,0 +1,33 @@
...
+
+        return connection
+    except ConnectionRefusedError:
+        logger.fatal("Could not connect to PostgreSQL: Connection refused")
+        sys.exit(1)
+
+
+async def check_version(connection):
+    with connection as con:
+        info = con.execute('SHOW server_version').fetchone()
```
igboyes, line 29: This is not async.

This version of `sqlalchemy` is supposed to have asyncio support. Can you check into this please?

igboyes (changes requested):
Great work. Database access needs to be async though.

virtool/postgres.py
```diff
@@ -0,0 +1,32 @@
...
+        logger.fatal("Could not connect to PostgreSQL: Connection refused")
+        sys.exit(1)
+
+
+async def check_version(connection):
+    info = await connection.execute(text('SHOW server_version'))
+
+    version = info.first()[0].split()[0]
+    logger.info(f"Found PostgreSQL {version}")
+    return
```
igboyes, line 32: Don't need explicit return.

virtool/postgres.py
```diff
@@ -0,0 +1,32 @@
...
+        async with postgres.connect() as connection:
+            await check_version(connection)
+
+            return connection
+    except ConnectionRefusedError:
+        logger.fatal("Could not connect to PostgreSQL: Connection refused")
+        sys.exit(1)
+
+
+async def check_version(connection):
```
igboyes, line 27: Add docstring and typing.

virtool/postgres.py
```diff
@@ -0,0 +1,32 @@
...
+import sys
+
+import sqlalchemy
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import create_async_engine
+
+logger = logging.getLogger(__name__)
+
+
+async def connect(postgres_connection_string: str) -> sqlalchemy.engine:
```
igboyes, line 11: Add docstring

virtool/startup.py
```diff
@@ -228,6 +229,14 @@ async def init_paths(app: aiohttp.web_app.Application):
         virtool.utils.ensure_data_dir(app["settings"]["data_path"])
 
 
+async def init_postgres(app: aiohttp.web_app.Application):
+    postgres_connection_string = app["config"].get("postgres_connection_string")
```
igboyes, line 233: Since the connection string is sourced from the application configuration, which has a default value, we should not use `get()`. Having an unset `postgres_connection_string` would be a bug and we should allow a `KeyError` to be raised if it doesn't exist.

```
app["config"]["postgres_connection_string"]
```

igboyes (changes requested):
A few more things to fix.

igboyes (approved):

virtool/postgres.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+async def check_version(connection: AsyncConnection):
+    """
+    Check and log the Postgres sever version.
+
+    :param connection:an AsyncConnection object
+
+    """
+    info = await connection.execute(text('SHOW server_version'))
```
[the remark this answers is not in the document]
igboyes: It works. The version is successfully logged into the console on start.

I think this is down to the fact that the database is not modified by this query and there is nothing to commit. Can you bring this up in the SQL channel on Slack?

This PR was merged already.

Issue #1974, opened 2020-11-19
More information in logging


peterk87:
It'd be nice to have more information in the logs, e.g. function name, line number, full module name, so it'd be easier to debug issues when developing.

[virtool/logs.py](https://github.com/virtool/virtool/blob/master/virtool/logs.py) logging with more info:

```python
import logging.handlers

import coloredlogs


def configure(verbose=False):
    logging_level = logging.DEBUG if verbose else logging.INFO

    logging.captureWarnings(True)

    log_format = '{asctime:<20} {module:<11} {levelname:<8} {message} [{name}:{funcName}:{lineno}]'

    coloredlogs.install(
        level=logging_level,
        fmt=log_format,
        style='{'
    )

    logger = logging.getLogger()

    handler = logging.handlers.RotatingFileHandler("virtool.log", maxBytes=1000000, backupCount=5)
    handler.setFormatter(logging.Formatter(log_format, style='{'))

    logger.addHandler(handler)

    return logger
```

igboyes:
For this issue specifically, we can change the formatter based on the `dev` configuration value to provide debug information. Without `dev` set, logging should be quite sparse like any other application.

I would like to create a logging epic to generally improve logging and determine how to integrate logging from multiple Virtool instances with a logging platform (eg. Azure monitoring, [ELK](https://www.elastic.co/log-monitoring)).

peterk87:
Sounds good! It'd be nice to have different streams/files for logging (access, events, error, analyses, etc).

What are your thoughts on structured logging (e.g. [structlog](https://github.com/hynek/structlog))? I've never used it before, but I can see the potential benefits.

peterk87:
@igboyes What do you think about adding [Rich](https://github.com/willmcgugan/rich) for logging? It has a ton of features, most of which we'd probably never use (unless you want to emojify logs :fire:). Tracebacks like [this](https://github.com/willmcgugan/rich#tracebacks) for dev would be really nice or being able to pretty print local vars. However, I'm not sure if the current library coloredlogs has similar features.

igboyes:
Yup. I have created a story for this (ch-114).

@jakeale has already used Rich for `virtool-cli` so I am putting him on it. I think we can point him to any further logging-related stories you may want to create.

Closing this issue.

PR #2023, opened 2021-01-07
Switch to asyncpg for async postgres as recommended by SQLAlchemy docs

peterk87:
## Changes:

- `requirements.txt`: Added asyncpg. Removed psycopg2-binary.
- Changed postgres connection strings from `"postgresql://"` to `"postgresql+asyncpg://"`
- `docker-compose.dev.yml`: Changed persistent volumes for services created in relative path, not system root (not a good idea to pollute Linux system root dir)
- `virtool/logs.py`: Added a bit more info to verbose/dev logs to make local dev a bit easier (issue #1974)
- `run.py`: Added shebang so you can run Virtool with `$ ./run.py --dev server` although may be a good idea to run with `$ python run.py ...` if Python3 is not the default system/env Python

See [SQLAlchemy docs](https://docs.sqlalchemy.org/en/14/orm/extensions/asyncio.html#synopsis-core) for use of [asyncpg](https://github.com/MagicStack/asyncpg) and connection strings beginning with `postgresql+asyncpg://` in example code.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -5,7 +5,7 @@ services:
         ports:
             - "27017:27017"
         volumes:
-            - ./mongo/data:/data/db
+            - /mongo/data:/data/db
         logging:
             driver: none
 
@@ -14,7 +14,8 @@ services:
         ports:
             - "6379:6379"
         volumes:
-            - ./redis/data:/data
+            - /redis/data:/data
+
         logging:
             driver: none
 
@@ -27,7 +28,7 @@ services:
             POSTGRES_USER: virtool
             POSTGRES_PASSWORD: virtool
         volumes:
-            - ./postgres/data:/var/lib/postgresql/data
+            - /postgres/data:/var/lib/postgresql/data
         logging:
             driver: none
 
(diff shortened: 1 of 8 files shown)


peterk87:
@igboyes Sorry for slipping in some minor unrelated changes into this PR. I was trying out the latest changes on `release-5.0.0` branch and noticed that the connection string I was providing was being called "Invalid" and noticed that [asnycpg](https://github.com/MagicStack/asyncpg) wasn't being used for async Postgres, so I created this PR.

igboyes (commented):
Looks good to me. However, I think via should not be using bind mounts at all. I didn't notice this in my previous review of this code.

Agree? Do you mind fixing this in another commit?

peterk87:
Hey @igboyes, if bind mounts aren't used, then how will DB/Redis data be persisted? Any state built up in dev would disappear on restart of services. I'm okay with bind mounts set to a path under the project root, but definitely not system root. 

It may be possible to set bind mounts from the CLI, but I'd have to check the docs for docker-compose.

igboyes:
Sorry for the lack of explanation. We should be using [Docker volumes](https://docs.docker.com/storage/volumes/). There is a section on volumes in compose as well: https://docs.docker.com/storage/volumes/#use-a-volume-with-docker-compose.

I can do this in a subsequent PR if that works better.

igboyes (approved):

Issue #114, opened 2017-07-21
Upload read file and create samples


igboyes:

igboyes:
Implemented in #113

PR #2031, opened 2021-01-12
Replace coloredlogs with Rich log handling

jakeale:
Related to [#114](https://app.clubhouse.io/virtool/story/114/use-built-in-rich-log-handler-for-python-console-log)
- I removed the `{name}:{funcName}:{lineno}` in the console log format since Rich is able to directly link to the path of the log call, though this doesn't seem to work in PyCharm for some reason.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -67,6 +67,7 @@ python-dateutil==2.8.1
 regex==2020.4.4
 requests==2.24.0
 requests-oauthlib==1.3.0
+rich==9.8.0
 semver==2.9.1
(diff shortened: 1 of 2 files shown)


jakeale:
I may have overlooked how this could mess up the logging to `virtool.log`. Doesn't seem to be logging anything at the moment.

igboyes (changes requested):
Make sure time stamp is logged for every line:
image

igboyes:
Closing for now as this does not look great in practice. We can re-open when there is more time to spend on it.

PR #2040, opened 2021-01-18
Add subtraction file download endpoint

tianshengsui:
ch-133

### DIFF
--- a/virtool/downloads/api.py
+++ b/virtool/downloads/api.py
@@ -13,6 +13,7 @@ import virtool.bio
 import virtool.analyses.db
 import virtool.db.utils
 import virtool.downloads.db
+import virtool.downloads.utils
 import virtool.history.db
 import virtool.otus.db
 import virtool.references.db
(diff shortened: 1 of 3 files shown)


virtool/downloads/api.py
```diff
@@ -204,3 +205,30 @@ async def download_sequence(req):
...
+    """
+    db = req.app["db"]
+    subtraction_id = req.match_info["subtraction_id"]
+
+    document = await db.subtraction.find_one(subtraction_id)
+
+    if document is None:
+        return virtool.api.response.not_found()
+
+    if "has_file" not in document or document["has_file"] is False:
```
igboyes, line 224:
```py
if not document.get("has_file", False):
```

virtool/downloads/api.py
```diff
@@ -204,3 +205,30 @@ async def download_sequence(req):
...
+        return virtool.api.response.not_found("Subtraction FASTA file not found")
+
+    path = os.path.join(
+        virtool.subtractions.utils.join_subtraction_path(req.app["settings"], subtraction_id),
+        "subtraction.fa.gz"
+    )
+
+    return web.FileResponse(path, headers={
+        "Content-Type": "application/gzip"
```
igboyes, line 233: Use `Content-Disposition` to change file name to something more meaningful. Currently, downloaded file is called `<subtraction_id>.gz`.

```
// Template
subtraction-<id>-<name>.fa.gz

// Actual
subtraction-12345-arabidopsis-thaliana.fa.gz 
```

Lowercase subtraction `name` and replace spaces with `-`.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

igboyes (changes requested):

igboyes (approved):

PR #2044, opened 2021-01-19
Return sample count in label API responses

tianshengsui:
ch-102

### DIFF
--- a/virtool/labels/db.py
+++ b/virtool/labels/db.py
@@ -1,2 +1,2 @@
-async def count_samples(db, label_id):
-    return await db.samples.count_documents({"labels": {"$in": [label_id]}})
+async def attach_sample_count(db, document, label_id):
+    document.update({"count": await db.samples.count_documents({"labels": {"$in": [label_id]}})})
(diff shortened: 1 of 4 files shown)


virtool/labels/db.py
```diff
@@ -0,0 +1,2 @@
+async def attach_sample_count(db, document, label_id):
+    document.update({"count": await db.samples.count_documents({"labels": {"$in": [label_id]}})})
```
igboyes, line 2: This should work:
```py
await db.samples.count_documents({"labels": label_id})
```

Right now this function relies on a side-effect to work: it modifies `document` from the outside scope. A purer approach is to return a new `dict`:

```py
async def attach_sample_count(db, document, label_id):
  return {
    **document,
    "count": await db.samples.count_documents({"labels": {"$in": [label_id]}})
  }
```

virtool/labels/db.py
```diff
@@ -0,0 +1,2 @@
+async def attach_sample_count(db, document, label_id):
```
igboyes, line 1: Needs docstring and typing.

Write specific tests for this function instead of relying on the integration tests for the API endpoints. In the API endpoint tests, you only test the case where each sample has one label. Add unit tests for this function that work with cases where a sample has zero or more labels.

igboyes (changes requested):

igboyes (approved):
Great. Thanks!

PR #2049, opened 2021-01-20
Return label information in sample API responses

tianshengsui:
ch-209


tests/samples/test_api.py
```diff
@@ -63,13 +66,24 @@ def __init__(self):
...
     time_3 = arrow.get(static_time.datetime).shift(hours=2).datetime
 
+    label_1 = Label(id=1, name="Bug", color="#a83432", description="This is a bug")
+    label_2 = Label(id=2, name="Info", color="#03fc20", description="This is a info")
+    label_3 = Label(id=3, name="Question", color="#0d321d", description="This is a question")
+
+    async with test_session as session:
+        session.add(label_1)
+        session.add(label_2)
+        session.add(label_3)
```
igboyes, line 84: Use this?:

https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.add_all

virtool/samples/db.py
```diff
@@ -94,6 +99,22 @@
...
+            async with AsyncSession(app["postgres"]) as session:
+                result = await session.execute(select(Label).filter_by(id=label_id))
+                label = result.scalar()
+
+            document["labels"][index] = {
+                "id": label_id,
+                "name": label.name,
+                "description": label.description,
+                "color": label.color
+            }
```
igboyes, line 114: Can this be done in a single SQL query?

---

Declaring an empty `list`, then appending each label `dict` is cleaner and less error prone than updating the old list by index. In the return statement, you can write:
```py
return {
  **document,
  "labels": labels
}
```

---

You are writing this in a number of places:
```py
{
    "id": label_id,
    "name": label.name,
    "description": label.description,
    "color": label.color
}
```

I think the idea with SQLAlchemy is that you should have a method on the `Label` class (eg. `Label.as_dict()` that does this for you. Can you look into this to confirm that I am right?

igboyes (changes requested):

igboyes (approved):

PR #2050, opened 2021-01-20
Support label assignment when creating samples through API

jakeale:
[ch148]


virtool/samples/utils.py
```diff
@@ -52,6 +57,16 @@ def calculate_workflow_tags(analyses: list) -> dict:
     }
 
 
+async def check_labels(labels, postgres_db):
```
igboyes, line 60: Add docstring. Change `postgres_db` to `pg` and put as first argument.

virtool/samples/utils.py
```diff
@@ -52,6 +57,16 @@ def calculate_workflow_tags(analyses: list) -> dict:
...
+    non_existent_labels = list()
+    for label_id in labels:
+        async with AsyncSession(postgres_db) as session:
```
igboyes, line 63: This is very inefficient. Can you find a method for accomplishing this in one query?

virtool/samples/api.py
```diff
@@ -361,11 +373,7 @@ async def edit(req):
             return bad_request(message)
 
     if "labels" in data:
-        non_existent_labels = list()
-        for label_id in data["labels"]:
-            async with AsyncSession(req.app["postgres"]) as session:
-                if (await session.execute(select(Label).filter_by(id=label_id))).first() is None:
-                    non_existent_labels.append(str(label_id))
+        non_existent_labels = check_labels(data["labels"], req.app["postgres"])
```
igboyes, line 376: You need to use `await` here.

virtool/samples/api.py
```diff
@@ -281,6 +292,7 @@ async def create(req):
         "user": {
             "id": user_id
         },
+        "labels": data.get("labels"),
```
igboyes, line 295: The `labels` field should always be an `list`. In this case, you are setting it to `None`. It should be an empty list.

igboyes (changes requested):

virtool/samples/utils.py
```diff
@@ -57,14 +57,18 @@ def calculate_workflow_tags(analyses: list) -> dict:
...
 
-async def check_labels(labels, postgres_db):
-    non_existent_labels = list()
-    for label_id in labels:
-        async with AsyncSession(postgres_db) as session:
-            if (await session.execute(select(Label).filter_by(id=label_id))).first() is None:
-                non_existent_labels.append(str(label_id))
-
-    return non_existent_labels
+async def check_labels(pg, labels: list) -> list:
```
igboyes, line 60: Label IDs are integers.

Need better typings. See https://docs.python.org/3/library/typing.html

```py
async def check_labels(pg: <need type for this>, labels: List[int]) -> List[int]:
```

virtool/samples/utils.py
```diff
@@ -57,14 +57,18 @@ def calculate_workflow_tags(analyses: list) -> dict:
...
+    Check for existence of labels given in sample creation request
+
+    :param pg: PostgreSQL database connection object
+    :param labels: list of labels given in the sample creation request
+    :return: a list containing any labels given in the request that aren't in the Labels db, if any
+    """
+    async with AsyncSession(pg) as session:
+        results = await session.execute(select(Label.id).filter(Label.id.in_(labels)))
+
+        return [str(label) for label in labels if label not in set(results.all())]
```
igboyes, line 71: Since IDs are integers, this shouldn't be necessary.

igboyes (changes requested):
Thanks for figuring that out. Typings need some improvement.

virtool/samples/utils.py
jakeale: I believe this was done in order to format a response since `join()` does not work with integers.
https://github.com/virtool/virtool/blob/8ed2beb9000c24800ba5752fefaf39ec2ab60353/virtool/samples/api.py#L240-L241

I could simply have the response return the list itself? e.g. `Labels do not exist: [1, 2, 3]`

igboyes: They are integers. You should not be changing their types so they work with `join()`. You can convert them for the purposes of the error response:

```py
f"Labels do not exist: {', '.join(str(label_id) for label_id in non_existent_labels)}"
```

igboyes (changes requested):

virtool/samples/utils.py
```diff
@@ -52,6 +58,20 @@ def calculate_workflow_tags(analyses: list) -> dict:
...
+    :param pg: PostgreSQL database connection object
+    :param labels: list of labels given in the sample creation request
+    :return: a list containing any labels given in the request that aren't in the Labels db, if any
+    """
+    async with AsyncSession(pg) as session:
+        results = await session.execute(select(Label.id).filter(Label.id.in_(labels)))
+
+        return [label for label in labels if label not in set(results.all())]
+
+
```
igboyes, line 74: Please add unit tests for this function.

virtool/samples/utils.py
```diff
@@ -68,6 +88,16 @@ def get_sample_rights(sample: dict, client):
...
+def create_bad_labels_response(labels: List[int]):
+    """
+    Creates a response for a bad request involving labels that do not exist
+
+    :param labels: A list of labels that do not exist in the Labels db
+    :return: A message to give in a `bad_request()` response
+    """
+    return f"Labels do not exist: {', '.join(str(label) for label in labels)}"
+
+
```
igboyes, line 100: Add unit tests.

The naming of the function needs to be more accurate. It is not creating a response. That is the duty of `bad_request`. It is creating an error message. You got it right in the `:return:` tag in the docstring.

The wording of the doctring is off. 

```
Creates an error message intended indicating that some label IDs do not exist. labels that do not exist

:param labels: a list of labels to include in the error message
```

igboyes (changes requested):

igboyes (approved):

PR #2055, opened 2021-01-25
Model uploads in SQL

jakeale:
[ch211]

### DIFF
--- a/virtool/uploads/models.py
+++ b/virtool/uploads/models.py
@@ -1,0 +1,24 @@
+from sqlalchemy import Column, String, Boolean, Integer, DateTime
+
+from virtool.postgres import Base
+
+
+class Upload(Base):
+    __tablename__ = "uploads"
+
+    id = Column(Integer, primary_key=True)
+    created_at = Column(DateTime)
+    field = Column(String)
+    name = Column(String, unique=True)
+    name_on_disk = Column(String, unique=True)
+    ready = Column(Boolean)
+    removed = Column(Boolean)
+    reserved = Column(Boolean)
+    size = Column(Integer)
+    type = Column(String)
+    uploaded_at = Column(DateTime)
+
+    def __repr__(self):
+        return f"<Upload(id={self.id}, created_at={self.created_at}, name={self.name}, " \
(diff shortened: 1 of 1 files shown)


virtool/uploads/models.py
```diff
@@ -0,0 +1,24 @@
...
+from virtool.postgres import Base
+
+
+class Upload(Base):
+    __tablename__ = "uploads"
+
+    id = Column(Integer, primary_key=True)
+    created_at = Column(DateTime)
+    field = Column(String)
+    name = Column(String, unique=True)
```
igboyes, line 12: Will not be unique. The user may upload the same file twice. This is why we use `name_on_disk` to keep the upload unique when the file name may be the same.

virtool/uploads/models.py
```diff
@@ -0,0 +1,24 @@
...
+    name_on_disk = Column(String, unique=True)
+    ready = Column(Boolean)
+    removed = Column(Boolean)
+    reserved = Column(Boolean)
+    size = Column(Integer)
+    type = Column(String)
```
igboyes, line 18: Please look into [enumerated types](https://www.postgresql.org/docs/13/datatype-enum.html) for this.

Possible values are:
- reference
- reads
- subtraction
- `null`

For `null`, I am not sure if this has to be in the enum.

When `type` is `null` it means that the file type is not explicitly understood by Virtool and does not map to a specific resources type (eg. reads -> sample, subtraction -> subtraction).

igboyes (changes requested):

igboyes:
The `user` subdocument was supposed to collapse to a column called `user`. I mistakenly typed `field` in the story.

virtool/uploads/models.py
```diff
@@ -1,21 +1,30 @@
-from sqlalchemy import Column, String, Boolean, Integer, DateTime
+import enum
+
+from sqlalchemy import Column, String, Boolean, Integer, DateTime, Enum
 
 from virtool.postgres import Base
 
 
+class ResourceType(enum.Enum):
```
igboyes, line 8: Why `ResourceType`? I think `UploadType` makes more sense.

igboyes (approved):

PR #2067, opened 2021-01-28
Additional work for attach_labels()

jakeale:


tests/samples/test_db.py
```diff
@@ -279,7 +279,7 @@ async def test_file_not_found(self, tmpdir, dbi):
         assert not await dbi.samples.count_documents({})
 
 
-async def test_attach_labels(pg_session, pg_engine):
+async def test_attach_labels(spawn_client, pg_session, pg_engine):
```
igboyes, line 282: This shouldn't be necessary. Does the test fail without it?

jakeale: Yes, this is the topmost exception that is thrown right after calling `await session.commit()`:
```
E   RuntimeError: Task <Task pending name='Task-5' coro=<test_attach_labels() running at /home/jake/Documents/projects/virtool/tests/samples/test_db.py:288> cb=[_run_until_complete_cb() at /usr/lib/python3.8/asyncio/base_events.py:184]> got Future <Future pending cb=[Protocol._on_waiter_completed()]> attached to a different loop
```

igboyes: Okay. I'll merge this. Can you please create a story describing the problem and assign to me?

igboyes (approved):

PR #2071, opened 2021-01-29
Remove all code related to the job agent

jakeale:


virtool/jobs/runner.py
```diff
@@ -80,116 +80,3 @@ async def _get_job_from_lists(self):
 
     async def _run_job(self):
         pass
-
-
-class JobAgent(Base):
```
igboyes, line 85: Since `Base` is only subclassed by `JobRunner` now, `JobRunner` and `Base` can be merged into a single class.

igboyes (changes requested):

igboyes (approved):

PR #2082, opened 2021-02-05
Reimplement tasks in SQL

tianshengsui:
[ch271]
[ch280]

- Re-implement tasks API endpoints in SQL
- Migrate `virtool.tasks.db` to` virtool.tasks.pg`
- Remove `FIRST_STEPS` dict
- Create a `TaskRunner` to store and run tasks using asyncio.Queue
- Re-implement `ProgressTracker`:
    - `get_tracker()` can return a `ProgressTracker` with initial progress and how much progress need to complete in the current step. Normally, progress will be evenly divided by the number of steps.
    - if a `file_size` is given,  developers can use `ProgressTracker.add()` to add value to progress, the value will be automatically convert to progress based on the value and file size, and update to the database.


tests/fixtures/client.py
```diff
@@ -59,6 +60,7 @@ async def connect(
         self.server = self._client.server
         self.app = self.server.app
         self.db = self.app.get("db", None)
+        self.postgres = self.app.get("postgres", None)
```
igboyes, line 63: Can be simplified to ``self.db = self.app.get("db")``. The default value for ``default`` in `get()` is `None`.

Please rename `postgres` to `pg`. Trying to standardize on this.

igboyes (changes requested):
* You write `tracker.initial + tracker.total` a lot. Can this be moved into the tracker logic so the task developer is not relied on to do it correctly everytime?
* Please type and document all code you worked.
* Can `register()` and `add_task()` be combined?
* Add tests for `Task` class

I think we could pass around task classes directly instead of using the string names (eg. "delete_reference").

For example, a call to register would look like this:
```py
await register(pg, DeleteReferenceTask, context=context)
```

The class will have an attribute on it called `task_type` that can be used to get that string value for writing to the database etc.

tianshengsui:
> I think we could pass around task classes directly instead of using the string names (eg. "delete_reference").
> 
> For example, a call to register would look like this:
> 
> ```python
> await register(pg, DeleteReferenceTask, context=context)
> ```
> 
> The class will have an attribute on it called `task_type` that can be used to get that string value for writing to the database etc.

I think we cannot get the attribute without instantiating the task class, and we need the `task_id` to instantiate it.

igboyes (approved):

PR #2083, opened 2021-02-08
Reimplement uploads API using SQL

jakeale:
Decided to use both `removed` and `removed_at` in the SQL model as I could not find any specific best practices.


virtool/uploads/api.py
```diff
@@ -53,59 +58,136 @@ async def naive_writer(req, file_id):
...
+                "name_on_disk": f"{upload_id}-{name}",
+                "uploaded_at": uploaded_at
+            })
+
+            logger.debug(f"Upload succeeded: {upload_id}")
+
+            headers = {
+                "Location": f"/api/uploads/{upload_id}"
             }
-        })
```
igboyes, line 92: Please get rid of references to `document` here. This is a MongoDB concept.

You should return the model object from `db.create()` then use `to_dict()` to get the dictionary representation of the data.

virtool/uploads/api.py
```diff
@@ -53,59 +58,136 @@ async def naive_writer(req, file_id):
...
+
+
+@routes.get("/api/uploads")
+async def find(req):
+    db = req.app["postgres"]
+    document = list()
+    filters = list()
+    user = req.query.get("user")
+
+    if user:
```
igboyes, line 123: Use consistent strategies for `user` and `type`. I'd say `get()` is cleaner.

virtool/uploads/api.py
```diff
@@ -53,59 +58,136 @@ async def naive_writer(req, file_id):
...
+    try:
+        filters.append(Upload.type == req.query["type"])
+    except KeyError:
+        pass
+
+    async with AsyncSession(db) as session:
+        if filters:
+            result = await session.execute(select(Upload).filter(*filters))
+        else:
+            result = await session.execute(select(Upload))
```
igboyes, line 135: Should work like this I think:
```py
query = select(Upload)

if filters:
  query.filter(*filters))

await session.execute(query)
```

virtool/uploads/api.py
```diff
@@ -53,59 +58,136 @@ async def naive_writer(req, file_id):
...
+
+    async with AsyncSession(db) as session:
+        result = (await session.execute(select(Upload).filter_by(id=upload_id))).scalar()
+
+        if not result:
+            return not_found("Upload record not found")
+
+    # check if the file has been removed as a result of a `DELETE` request
+    if result.removed:
+        return not_found("Uploaded file has already been removed")
```
igboyes, line 158: These should appear the same to the API consumer: just `not_found()`.

virtool/uploads/api.py
```diff
@@ -53,59 +58,136 @@ async def naive_writer(req, file_id):
...
+        if result.removed:
+            return bad_request("Uploaded file has already been removed")
 
-        headers = {
-            "Location": f"/api/files/{file_id}"
-        }
+        try:
+            os.remove(Path(req.app["settings"]["data_path"]) / "files" / result.name_on_disk)
+        except FileNotFoundError:
+            return not_found("Uploaded file not found at expected location")
```
igboyes, line 186: Just `pass` for this. If the file is already gone, I wouldn't consider it an error state.

virtool/uploads/api.py
```diff
@@ -53,59 +58,136 @@ async def naive_writer(req, file_id):
...
+            return not_found("Upload record not found")
+
+        if result.removed:
+            return bad_request("Uploaded file has already been removed")
 
-        headers = {
-            "Location": f"/api/files/{file_id}"
-        }
+        try:
+            os.remove(Path(req.app["settings"]["data_path"]) / "files" / result.name_on_disk)
```
igboyes, line 184: Need to use executor for this because it will block the server significantly if the filesystem is slow.

```py
await app["run_in_thread"](
  os.remove,
  path
)
```

virtool/uploads/api.py
```diff
@@ -53,59 +58,136 @@ async def naive_writer(req, file_id):
...
+    upload_id = int(req.match_info["id"])
+
+    async with AsyncSession(db) as session:
+        result = (await session.execute(select(Upload).where(Upload.id == upload_id))).scalar()
+
+        if not result:
+            return not_found("Upload record not found")
+
+        if result.removed:
+            return bad_request("Uploaded file has already been removed")
```
igboyes, line 181: Return `not_found()` for both.

virtool/uploads/db.py
```diff
@@ -0,0 +1,38 @@
...
+            ready=False,
+            removed=False,
+            reserved=reserved,
+            type=upload_type,
+            user=user
+        )
+
+        session.add(upload)
+
+        await session.flush()
```
igboyes, line 33: Why is flush needed here? Just curious.

virtool/uploads/db.py
```diff
@@ -0,0 +1,38 @@
...
+        document = upload.to_dict()
+
+        await session.commit()
+
+        return document
```
igboyes, line 38: Simplify to:
```py
await session.commit()

return upload.to_dict()
```

igboyes (changes requested):

virtool/uploads/db.py
jakeale: It seems like unique IDs are not generated for each `Upload` instance until they are flushed or committed, so this was to ensure that the returned dictionary included a proper `id` key.

virtool/uploads/db.py
jakeale: I was wanting to have it like this but I seem to lose access to `upload` completely after the commit.

This is what `upload` refers to before the `commit()`:
`<Upload(id=None, created_at=2015-10-06 20:00:00, name=Test.fq.gz, name_on_disk=None, ready=False, removed=False, reserved=False, size=None, type=hmm, user=test, uploaded_at=None>`

After:
`Unable to get repr for <class 'virtool.uploads.models.Upload'>`

virtool/uploads/api.py
```diff
@@ -53,59 +58,137 @@ async def naive_writer(req, file_id):
...
+            await session.commit()
 
-        await db.files.update_one({"_id": file_id}, {
-            "$set": {
+            upload_.update({
                 "size": size,
-                "ready": True
+                "name_on_disk": f"{upload_id}-{name}",
+                "uploaded_at": uploaded_at
+            })
```
igboyes, line 98: In this block, `name_on_disk` is derived in three different places. Why isn't it done in `db.create()`? You can create the record and get its `name_on_disk` in one transaction.

The duplication between updating the model at line 88 and updating the `dict` at line 94 can be simplified using `flush()` as you did in `db.create()`.

virtool/uploads/api.py
```diff
@@ -53,59 +58,137 @@ async def naive_writer(req, file_id):
...
+            return aiohttp.web.Response(status=499)
+
+
+@routes.get("/api/uploads")
+async def find(req):
+    db = req.app["postgres"]
+    upload_ = list()
+    filters = list()
+    user = req.query.get("user")
+    type_ = req.query.get("type")
```
igboyes, line 122: You name this `upload_type` in other conceptually similar situations. Make this consistent. If you are having to use a trailing `_` because your name shadows a built-in (`type`), you probably need a more informative name.

tests/uploads/test_api.py
```diff
@@ -1,68 +1,150 @@
...
+    @pytest.fixture
+    async def prepare_db(self, pg_session, static_time):
+        upload_1 = Upload(id=1, name="test.fq.gz", type="reads", user="danny")
+        upload_2 = Upload(id=2, name="test.fq.gz", type="subtraction", user="lester")
+        upload_3 = Upload(id=3, name="test.fq.gz", user="jake")
+
+        async with pg_session as session:
+            session.add_all([upload_1, upload_2, upload_3])
+
+            await session.commit()
```
igboyes, line 70: Should be module-level function. Does not need to be in class.

igboyes (changes requested):
See comments.

Please do the following:
* Move the database code into `uploads.db` and test separately. In some cases your tests won't catch the situation where the handler works, but no change is actually made in the DB.
* Refer to the Postgres client as `pg`, not `db`. It is confusing because `db` conventionally refers to MongoDB in Virtool.
* Document and type all new code as well as that in `uploads.models`. Please always do this.
* Restore the test for `DELETE /api/uploads/:id` that you removed after the first review. It is useful to test that the deletion handles the situation where the file is missing.

PR #2088, opened 2021-02-12
Add return value in check_mongo_version()

tianshengsui:
[ch335]
The problem is because `check_mongo_version` didn't have a return value, so the `mongo_version` is null which makes home page keep loading.


igboyes (approved):
Thanks for figuring that out @tianshengsui

PR #2089, opened 2021-02-12
Upload improvements

jakeale:
- Set `ready` field to `true` when the upload is finalized
- Separate concerns between `db` and `api` modules
- Don't show `removed` uploads in find response
- Add `documents` key to find response
- Improve model definition using `nullable` and `default`
- Use queries instead of application code for checking records


igboyes (approved):
This looks perfect now!

PR #2095, opened 2021-02-16
Add task for populating files field on subtraction documents

tianshengsui:
[ch369]
- Add a task for adding `files` field on subtraction documents and rename Bowtie2 index files
- Spawn the task on application start

### DIFF
--- a/virtool/subtractions/db.py
+++ b/virtool/subtractions/db.py
@@ -13,6 +13,8 @@ import virtool.tasks.db
 import virtool.tasks.task
 import virtool.utils
 
+from virtool.types import App
+
 PROJECTION = [
     "_id",
     "count",
@@ -28,7 +30,7 @@ PROJECTION = [
 
 class AddSubtractionFilesTask(virtool.tasks.task.Task):
 
-    def __init__(self, app, task_id):
+    def __init__(self, app: App, task_id: str):
         super().__init__(app, task_id)
 
         self.steps = [
@@ -37,6 +39,10 @@ class AddSubtractionFilesTask(virtool.tasks.task.Task):
         ]
 
     async def rename_index_files(self):
+        """
+        Change Bowtie2 index name from 'reference' to 'subtraction'
+
+        """
         settings = self.app["settings"]
 
         async for subtraction in self.db.subtraction.find({"deleted": False, "files": {"$exists": False}}):
@@ -44,6 +50,10 @@ class AddSubtractionFilesTask(virtool.tasks.task.Task):
             await self.app["run_in_thread"](virtool.subtractions.utils.rename_bowtie_files, path)
 
     async def add_files_field(self):
+        """
+        Add a 'files' field to subtraction documents to list what files can be downloaded for that subtraction
+
+        """
         settings = self.app["settings"]
 
         async for subtraction in self.db.subtraction.find({"deleted": False, "files": {"$exists": False}}):
--- a/virtool/subtractions/db.py
+++ b/virtool/subtractions/db.py
@@ -25,6 +25,65 @@ PROJECTION = [
     "has_file"
 ]
 
+FILES = [
+    "subtraction.fa.gz",
+    "subtraction.1.bt2",
+    "subtraction.2.bt2",
+    "subtraction.3.bt2",
+    "subtraction.4.bt2",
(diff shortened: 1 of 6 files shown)


virtool/subtractions/db.py
```diff
@@ -25,6 +25,65 @@
...
+
+    async def rename_index_files(self):
+        settings = self.app["settings"]
+
+        async for subtraction in self.db.subtraction.find({"deleted": False}):
+            path = virtool.subtractions.utils.join_subtraction_path(settings, subtraction["_id"])
+            for file in os.listdir(path):
+                if file.endswith(".bt2"):
+                    file_path = os.path.join(path, file)
+                    os.rename(file_path, file_path.replace("reference", "subtraction"))
```
igboyes, line 57: This should not run if it has already been done for the subtraction. Using a query like the following would work to skip these cases I think:
```py
{"files": {"$exists": false}}
```

igboyes, line 57: This should run in a thread in case the filesystem is slow. Modifying the file system synchronously will block the whole application.

I suggest extracting lines 54-57 to a separate function and use app["run_in_thread"] to execute.

virtool/subtractions/db.py
```diff
@@ -25,6 +25,65 @@
...
+
+    def __init__(self, app, task_id):
+        super().__init__(app, task_id)
+
+        self.steps = [
+            self.rename_index_files,
+            self.add_files_field,
+        ]
+
+    async def rename_index_files(self):
```
igboyes, line 49: Needs docstring.

virtool/subtractions/db.py
```diff
@@ -25,6 +25,65 @@
...
+    "subtraction.3.bt2",
+    "subtraction.4.bt2",
+    "subtraction.rev.1.bt2",
+    "subtraction.rev.2.bt2"
+]
+
+
+class AddSubtractionFilesTask(virtool.tasks.task.Task):
+
+    def __init__(self, app, task_id):
```
igboyes, line 41: Needs typing

virtool/subtractions/db.py
```diff
@@ -25,6 +25,65 @@
...
+        settings = self.app["settings"]
+
+        async for subtraction in self.db.subtraction.find({"deleted": False}):
+            path = virtool.subtractions.utils.join_subtraction_path(settings, subtraction["_id"])
+            for file in os.listdir(path):
+                if file.endswith(".bt2"):
+                    file_path = os.path.join(path, file)
+                    os.rename(file_path, file_path.replace("reference", "subtraction"))
+
+    async def add_files_field(self):
```
igboyes, line 59: Needs docstring.

virtool/subtractions/db.py
```diff
@@ -25,6 +25,65 @@
...
+        settings = self.app["settings"]
+
+        async for subtraction in self.db.subtraction.find({"deleted": False}):
```
igboyes, line 62: Use query to exclude documents where `files` already exists.

virtool/subtractions/db.py
```diff
@@ -25,6 +25,65 @@
...
+                    file_path = os.path.join(path, file)
+                    document = {
+                        "size": virtool.utils.file_stats(file_path)["size"],
+                        "name": file
+                    }
+
+                    if file.endswith(".fa.gz"):
+                        document["type"] = "fasta"
+                    if file.endswith(".bt2"):
+                        document["type"] = "reference"
```
igboyes, line 77: Change "reference" to "bowtie2".

virtool/subtractions/db.py
```diff
@@ -25,6 +25,65 @@
...
+
+                    files.append(document)
```
igboyes, line 79: This will block on slow filesystems so extract to function and use `app["run_in_thread"]` again.

igboyes (changes requested):
In addition to comments, tests need to be written for the task and two functions that need to be added.

igboyes (changes requested):
Thanks. This works perfectly.

Please update `test_add_subtraction_files_task` to test that a subtraction with a `files` field is ignored by the task. With large numbers of subtractions this is an essential behaviour that must be tested!

igboyes (approved):
Thanks!

PR #2098, opened 2021-02-17
Fix broken password visibility toggle

OfficialArms:
The previous implementation defaulted to using the initial state at the beginning of all renders after the eye icon is pressed more than once

### DIFF
--- a/client/src/js/base/Input.js
+++ b/client/src/js/base/Input.js
@@ -132,7 +132,12 @@ export const PasswordInput = props => {
     return (
         <InputContainer align="right">
             <Input type={show ? "text" : "password"} {...props} />
-            <InputIcon name={show ? "eye-slash" : "eye"} onClick={() => {setShow(prevShow => !prevShow)}} />
+            <InputIcon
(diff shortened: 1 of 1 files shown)


client/src/js/base/Input.js
```diff
@@ -132,7 +132,12 @@ export const PasswordInput = props => {
     return (
         <InputContainer align="right">
             <Input type={show ? "text" : "password"} {...props} />
-            <InputIcon name={show ? "eye-slash" : "eye"} onClick={() => {setShow(prevShow => !prevShow)}} />
+            <InputIcon
+                name={show ? "eye-slash" : "eye"}
+                onClick={() => {
+                    setShow(prevShow => !prevShow);
+                }}
```
igboyes, line 139: This can be simplified to:
```js
onClick={() => setShow(prevShow => !prevShow)}
```

No need for curly brackets.

igboyes (changes requested):

igboyes (approved):

PR #2100, opened 2021-02-17
Support upload of analysis files

jakeale:
- Implement SQL model for `analysis_files`, a table to store analysis result files
- Support uploading at `POST /api/analyses/:id/files`
- Support downloading at `GET /api/analyses/:id/files/:id`
- Attach detailed file information from `analysis_files` SQL table to `GET /api/analyses/:id` calls

### DIFF
--- a/tests/analyses/test_api.py
+++ b/tests/analyses/test_api.py
@@ -1,3 +1,7 @@
+import os
+import sys
+from pathlib import Path
+
 import pytest
 from aiohttp.test_utils import make_mocked_coro
 
@@ -233,6 +237,68 @@ async def test_remove(mocker, error, spawn_client, resp_is):
     assert await client.db.analyses.find_one() is None
 
     assert m_remove.called_with("data/samples/baz/analyses/foobar", True)
+
+
+@pytest.mark.parametrize("error", [None, "400", "404", "422"])
+async def test_upload_file(error, resp_is, spawn_client, static_time, snapshot, tmpdir):
+    """
+    Test that an analysis result file is properly uploaded, a record is inserted into the `analysis_files` SQL table,
+    and that the `list` field in a given analysis is updated to reflect the new file.
+
+    """
+    tmpdir.mkdir("files")
+
+    path = Path(sys.path[0]) / "tests" / "test_files" / "aodp" / "reference.fa"
+
+    data = {
+        "file": open(path, "rb")
+    }
+
+    client = await spawn_client(authorize=True, administrator=True)
+
+    client.app["settings"]["data_path"] = str(tmpdir)
+
+    if error == "400":
+        format_ = "foo"
+    else:
+        format_ = "fasta"
+
+    if error != "404":
+        await client.db.analyses.insert_one({
+            "_id": "foobar",
+            "ready": True,
+            "job": {
+                "id": "hello"
+            },
+            "files": []
+        })
+
+    if error == "422":
+        resp = await client.post_form(f"/api/analyses/foobar/files?format=fasta", data=data)
+    else:
+        resp = await client.post_form(f"/api/analyses/foobar/files?name=reference.fa&format={format_}", data=data)
+
+    if error is None:
+        assert resp.status == 201
+
+        snapshot.assert_match(await resp.json())
+
+        assert os.listdir(tmpdir / "analyses") == ["1-reference.fa"]
+
+        document = await client.db.analyses.find_one("foobar")
+
+        assert document["files"] == ["1-reference.fa"]
+
+    elif error == "400":
+        assert await resp_is.bad_request(resp, "Unsupported analysis file format")
+
+    elif error == "404":
+        assert resp.status == 404
+
+    elif error == "422":
+        assert await resp_is.invalid_query(resp, {
+            "name": ["required field"]
+        })
 
 
 @pytest.mark.parametrize("error", [None, "400", "403", "404_analysis", "404_sequence", "409_workflow", "409_ready"])
--- a/tests/analyses/test_api.py
+++ b/tests/analyses/test_api.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import pytest
 from aiohttp.test_utils import make_mocked_coro
 
+import virtool.analyses.db
 from virtool.utils import base_processor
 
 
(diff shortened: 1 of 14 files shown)


igboyes:
I think the first three commits in this PR are from another branch.

jakeale:
> I think the first three commits in this PR are from another branch.

Yes I think I messed up a rebase. It looks like they can be safely dropped if you don't mind another rebase.

tests/analyses/test_api.py
```diff
@@ -235,6 +256,90 @@ async def test_remove(mocker, error, spawn_client, resp_is):
...
+        assert resp.status == 404
+
+    elif error == 422:
+        assert await resp_is.invalid_query(resp, {
+            "name": ["required field"]
+        })
+
+
+@pytest.mark.parametrize("exists", [True, False])
+async def test_download_file(exists, files, spawn_client, snapshot, tmpdir):
```
igboyes, line 314: Can you figure out how to test that the correct file is actually written?

virtool/analyses/api.py
```diff
@@ -174,6 +186,93 @@ async def remove(req: aiohttp.web.Request) -> aiohttp.web.Response:
...
+    if name_on_disk in files:
+        return conflict("File is already associated with analysis")
+
+    file_id = analysis_file["id"]
+    analysis_file_path = Path(req.app["settings"]["data_path"]) / "analyses" / name_on_disk
+
+    try:
+        size = await virtool.uploads.utils.naive_writer(req, analysis_file_path)
+    except asyncio.CancelledError:
+        logger.debug(f"Upload aborted: {file_id}")
```
igboyes, line 230: Change to "Analysis file upload" to differentiate from user uploads.

virtool/analyses/utils.py
```diff
@@ -10,6 +14,25 @@
 )
 
 
+async def attach_analysis_files(pg: AsyncEngine, analysis_id: str) -> List[Dict]:
```
igboyes, line 17: I don't think this is unit tested.

igboyes (changes requested):

virtool/analyses/api.py
```diff
@@ -174,6 +186,93 @@ async def remove(req: aiohttp.web.Request) -> aiohttp.web.Response:
...
+        return invalid_query(errors)
+
+    name = req.query.get("name")
+
+    if analysis_format not in ANALYSIS_FORMATS:
+        return bad_request("Unsupported analysis file format")
+
+    analysis_file = await virtool.analyses.db.create_row(pg, analysis_id, analysis_format, name)
+
+    name_on_disk = analysis_file["name_on_disk"]
```
igboyes, line 218: You need to refer to the files by their `id`, not `name_on_disk`.

virtool/analyses/api.py
```diff
@@ -174,6 +186,93 @@ async def remove(req: aiohttp.web.Request) -> aiohttp.web.Response:
...
+        size = await virtool.uploads.utils.naive_writer(req, analysis_file_path)
+    except asyncio.CancelledError:
+        logger.debug(f"Analysis file upload aborted: {file_id}")
+        await virtool.analyses.db.delete_row(pg, file_id)
+
+        return aiohttp.web.Response(status=499)
+
+    analysis_file = await virtool.uploads.db.finalize(pg, size, file_id, AnalysisFile)
+
+    if not analysis_file:
```
igboyes, line 237: If this happens it is a bug. We should not handle it and blame it on the client (`4**` status).

Clients should also not receive messages about implementation details of the application such as problems with the backing SQL database. The API is meant to abstract away these details.

virtool/analyses/api.py
```diff
@@ -174,6 +186,93 @@ async def remove(req: aiohttp.web.Request) -> aiohttp.web.Response:
...
+        await req.app["run_in_thread"](os.remove, analysis_file_path)
+        return not_found("Row not found in table after file upload")
+
+    files.append(name_on_disk)
+
+    await db.analyses.update_one({"_id": analysis_id}, {
+        "$set": {"files": files}
+    })
```
igboyes, line 245: Use `$push`. If there are two requests being handled at the same time (ie. the workflow is concurrently uploading two files), rewriting the `files` field could cause a serious bug.

virtool/analyses/db.py
```diff
@@ -208,6 +214,71 @@ async def create(
     return document
 
 
+async def create_row(pg: AsyncEngine, analysis_id: str, analysis_format: str, name: str) -> Dict[str, any]:
```
igboyes, line 217: The naming of these is going to clash with the `Analysis` model when it is eventually created.

I think we should move the analysis file code into a new module: `virtool.analyses.files`. Also, change these function names to something like `create_analysis_file`.

virtool/analyses/db.py
```diff
@@ -208,6 +214,71 @@ async def create(
...
+        await session.commit()
+
+
+async def get_row(pg: AsyncEngine, file_id: int) -> Optional[AnalysisFile]:
+    """
+    Get a row that represents an analysis result file by its `id`
+
+    :param pg: PostgreSQL AsyncEngine object
+    :param file_id: Row `id` to get
+    :return:
```
igboyes, line 271: Describe `:return:`.

virtool/analyses/models.py
```diff
@@ -0,0 +1,48 @@
...
+    tsv = "tsv"
+    json = "json"
+
+
+ANALYSIS_FORMATS = [*AnalysisFormat.to_list(), None]
+
+
+class AnalysisFile(Base):
+    """
+    SQL table to store new analysis files
```
igboyes, line 31: Is a model, not a table.

virtool/analyses/utils.py
```diff
@@ -10,6 +14,25 @@
...
+    :param analysis_id: An id for a specific analysis
+    :return: List of file details for each file associated with an analysis
+    """
+    files = []
+
+    async with AsyncSession(pg) as session:
+        results = (await session.execute(select(AnalysisFile).filter_by(analysis=analysis_id))).scalars().all()
+
+        for result in results:
+            files.append(result.to_dict())
```
igboyes, line 31:
```py
    results = (await session.execute(select(AnalysisFile).filter_by(analysis=analysis_id))).scalars().all()

return [result.to_dict() for result in results]
```

virtool/uploads/utils.py
```diff
@@ -28,6 +28,11 @@ async def naive_writer(req, file_path) -> int:
 
     size = 0
 
+    try:
+        file_path.parent.mkdir()
```
igboyes, line 32: Must run in thread to prevent filesystem blocking.

igboyes (changes requested):

igboyes (approved):

PR #2101, opened 2021-02-17
Index Deletion Endpoint 

BlakeASmith:
Adds a DELETE endpoint at /api/indexes/{index_id} which is only accessible to virtool jobs. This route is intendend to be used when a build_index job fails.

- Deletes the index with the given ID from the database.
- Resets any history database entries referencing that index.
	- "index.id" and "index.version" set to "unbuilt"
- Returns a 404 NOT FOUND response if the index does not exist.
- Returns a 204 NO CONTENT response otherwise.
- Adds a new test fixture spawn_job_client which provides an aiohttp client which can authenticate with the API as a Job. It accepts thesame arguments as the spawn_client fixture.

### DIFF
--- a/virtool/indexes/api.py
+++ b/virtool/indexes/api.py
@@ -14,7 +14,7 @@ import virtool.references.db
 import virtool.utils
 from virtool.api.response import bad_request, conflict, insufficient_rights, json_response, \
     not_found, no_content
-from virtool.db.core import DB
+from virtool.indexes.db import reset_history
 from virtool.jobs.utils import JobRights
 
 routes = virtool.http.routes.Routes()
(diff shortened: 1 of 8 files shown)


virtool/indexes/api.py
```diff
@@ -236,3 +239,44 @@ async def find_history(req):
...
+        }
+    })
+
+
+@routes.delete("/api/indexes/{index_id}", jobs_only=True)
+async def delete_index(req: aiohttp.web.Request):
+    """Delete the index with the given id and reset history relating to that index."""
+    index_id = req.match_info["index_id"]
+    db = req.app["db"]
+    indexes: Collection = db.indexes
```
igboyes, line 272: I think you are trying to improve type hinting by declaring `indexes` here instead of calling `db.indexes.delete_one`. I'd prefer foregoing nice type hinting in favour of avoiding unnecessary variable declarations.

We can improve the implementation of `virtool.db.core` once the workflow work is done to solve the typing problem.

virtool/indexes/api.py
```diff
@@ -236,3 +239,44 @@ async def find_history(req):
...
+    }
+
+    return await history.update_many(query, {
+        "$set": {
+            "index": {
+                "id": "unbuilt",
+                "version": "unbuilt"
+            }
+        }
+    })
```
igboyes, line 264: Move to `virtool.indexes.db`. API modules should only contain handlers, route tables, and possibly constants.

Please pass the function `db` instead of the specific collection object.

1. This is convention in Virtool.
2. If in the future this function needs access to other collections, the function signature will not change and usages will not need to be updated.
3. `db.collection.method` is a pattern reflected in the `motor` documentation and Mongo shell

igboyes (changes requested):
Very nice. Please see comments.

igboyes (approved):

PR #2102, opened 2021-02-18
Add subtraction upload endpoints

tianshengsui:
[ch375]


igboyes:
A file needs to be uploaded in this request. The `name` needs to be in the URL query, not a JSON body. See how this is done in `virtool.uploads`.

Please find a way to reuse code from the modules in `virtool.uploads`. Discuss with @jakeale so you stay in sync on this and any changes necessary to extend this code for subtraction uploads.

tianshengsui:
@igboyes Can I assume the upload file will be stored at `<data_path>/subtractions/:id/<file_name>`?

igboyes:
> @igboyes Can I assume the upload file will be stored at `<data_path>/subtractions/:id/<file_name>`?

Yes. And that there is only that strict set of allowed files names defined in the story.

virtool/subtractions/api.py
```diff
@@ -173,6 +185,69 @@ async def create(req):
...
+
+    file = {
+        "name": file_name,
+        "size": size,
+        "type": file_type
+    }
+
+    await db.subtraction.find_one_and_update({"_id": subtraction_id}, {
+        "$push": {
+            "files": file
```
igboyes, line 240: Store file IDs in this array. They will be used to join the SQL records to the subtraction when it is requested.

igboyes (changes requested):
Very good. One change required.

virtool/subtractions/api.py
tianshengsui: Only file IDs?

igboyes (approved):

PR #2107, opened 2021-02-19
Add `GET /download/hmms/profiles.hmm` to download HMM profiles

jakeale:


tests/downloads/test_api.py
```diff
@@ -62,6 +65,35 @@ async def test_all(get, missing, spawn_client):
...
+            await client.app["run_in_thread"](shutil.copy, test_file_path, file_path)
+            file_path = file_path / "profiles.hmm"
+            assert file_path.exists()
+
+    resp = await client.get("/download/hmms/profiles.hmm")
+
+    if data_exists and file_exists:
+        assert resp.status == 200
+    else:
+        assert resp.status == 404
```
igboyes, line 94: Does not test that correct file was downloaded. Find way to test.

virtool/downloads/api.py
```diff
@@ -58,6 +58,28 @@ async def download_analysis(req):
...
+    Download the HMM profiles file if HMM data is available.
+
+    """
+    file_path = Path(req.app["settings"]["data_path"]) / "hmm" / "profiles.hmm"
+
+    if not file_path.parent.is_dir():
+        return virtool.api.response.not_found("HMM data could not be found")
+
+    if not file_path.is_file():
+        return virtool.api.response.not_found("Profiles file could not be found")
```
igboyes, line 73: Not necessary to return different nessages. `HMM data could not be found` is not very informative. Use `Profiles file could not be found` in both cases.

Combine these checks and extract into function in `hmm.utils`. Then, run in thread to avoid blocking. Please double-check that a function like this doesn't already exist.

virtool/downloads/api.py
```diff
@@ -58,6 +58,28 @@ async def download_analysis(req):
...
+
+    headers = {
+        "Content-Length": virtool.utils.file_stats(file_path)["size"],
```
igboyes, line 76: I think `FileResponse` may calculate `Content-Length` automatically. Please check.

igboyes (changes requested):

virtool/hmm/utils.py
```diff
@@ -18,3 +20,10 @@ def format_hmm_release(updated, release, installed):
     )
 
     return formatted
+
+
+def hmm_data_exists(file_path: Path):
+    if not file_path.parent.is_dir() or not file_path.is_file():
+        return False
+
+    return True
```
igboyes, line 29: 1. Docstring
2. Simplify:
  ```py
  def hmm_data_exists(file_path: Path):
￼    return file_path.parent.is_dir() or file_path.is_file()
  ```

igboyes (changes requested):

igboyes (approved):

PR #2109, opened 2021-02-19
Add download endpoints for subtraction bowtie2 files

tianshengsui:
[ch372]


tests/downloads/test_api.py
```diff
@@ -134,3 +134,38 @@ async def test_download_cache_reads(error, paired, tmpdir, spawn_client, resp_is
...
+
+    await client.db.subtraction.insert_one(subtraction)
+
+    resp = await client.get("/download/subtraction/foo/subtraction.1.bt2")
+
+    if error == "404":
+        assert resp.status == 404
+        return
+
+    assert resp.status == 200
```
igboyes, line 170: Please figure out how to test that the correct file is downloaded. Check-in with @jakeale as I have asked him to do this also.

virtool/downloads/api.py
```diff
@@ -242,6 +242,36 @@ async def download_sequence(req):
...
+    suffix = req.match_info["suffix"]
+    file_name = f"subtraction.{reverse}{suffix}.bt2"
+
+    document = await db.subtraction.find_one(subtraction_id)
+
+    if document is None:
+        return virtool.api.response.not_found()
+
+    if not await db.subtraction.count_documents({"_id": subtraction_id, "files.name": file_name}):
+        return virtool.api.response.not_found("Bowtie2 file not found")
```
igboyes, line 263: I think this can just be reduced to:
```py
if not await db.subtraction.count_documents({"_id": subtraction_id, "files.name": file_name}):
    return virtool.api.response.not_found()
```

igboyes (changes requested):

igboyes (approved):

igboyes (approved):

PR #2113, opened 2021-02-23
Modify `DELETE /api/samples/:id` for use with the Jobs API

jakeale:
There seems to be an issue currently where two endpoints (one for the web API, one for the jobs API) cannot share the same route when the jobs API endpoint uses a `jobs_only` flag. Blake is aware of the issue and figured an implementation like this would work for now.


virtool/files/db.py
```diff
@@ -145,6 +145,20 @@ async def reserve(db, file_ids: list):
...
+    :param db: Application's MongoDB client
+    :param file_ids: List of `file_ids`
+    """
+    await db.files.update_many({"_id": {"$in": file_ids}}, {
+        "$set": {
+            "reserved": False
+        }
+    })
+
+
```
igboyes, line 161: This should be done in SQL (`Upload` model).

virtool/samples/api.py
```diff
@@ -485,12 +486,27 @@ async def remove(req):
...
     sample_id = req.match_info["sample_id"]
 
     try:
-        if not await virtool.samples.db.check_rights(db, sample_id, req["client"]):
-            return insufficient_rights()
+        if isinstance(client, JobClient):
+            ready = await virtool.db.utils.get_one_field(db.samples, "ready", sample_id)
+
+            if ready is None:
+                raise virtool.errors.DatabaseError("Sample does not exist")
```
igboyes, line 498: `404 Not found`

virtool/samples/api.py
```diff
@@ -485,12 +486,27 @@ async def remove(req):
...
+            file_ids = await virtool.db.utils.get_one_field(db.samples, "files", sample_id)
+
+            if file_ids:
+                await virtool.files.db.release(db, file_ids)
+        else:
+            if not await virtool.samples.db.check_rights(db, sample_id, client):
+                return insufficient_rights()
     except virtool.errors.DatabaseError as err:
         if "Sample does not exist" in str(err):
             return not_found()
```
igboyes, line 512: I think the best approach is to split this into two handlers: one for jobs and one for normal clients. Using conditionals is too flaky and will get very messy in some other handlers.

igboyes (changes requested):
I want to put this on ice until the job/non-job routing is cleared up. Leave changes until that is sorted.

jakeale:
I noticed that while running the tests that `id` in the `uploads` SQL table kept incrementing, and would only reset if I ran a test that used a regular `client` fixture. For these tests that only meant that I couldn't hard-code the expected file `id` to be `1`, but I wonder what kind of issues this might cause if multiple tests that use the `JobClient` and access the SQL database are run in succession.

virtool/uploads/db.py
```diff
@@ -146,15 +146,39 @@ async def delete(pg: AsyncEngine, upload_id: int) -> Optional[dict]:
...
+        query = (Upload.id.in_(upload_ids))
+
+    async with AsyncSession(pg) as session:
+        await session.execute(update(Upload).
+                              where(query).values(reserved=False).
+                              execution_options(synchronize_session="fetch")
+                              )
+
+
+async def reserve(pg: AsyncEngine, upload_ids: Union[int, List[int]]):
```
igboyes, line 168: Needs tests.

tests/samples/test_api.py
```diff
@@ -440,6 +442,40 @@ async def mock_remove_samples(*args, **kwargs):
...
+
+    resp = await client.delete("/api/samples/test")
+
+    if exists and not ready:
+        assert resp.status == 204
+    elif exists:
+        assert resp.status == 400
+    else:
+        assert resp.status == 404
+
```
igboyes, line 477: This test is not effective. Please think again about what behaviours need to be tested to ensure this function works in production.

Also, `remove_samples()` is a coroutine. The `mocker.patch` function only supports synchronous mocks. See https://docs.aiohttp.org/en/stable/testing.html#aiohttp.test_utils.make_mocked_coro.

igboyes (changes requested):

PR #2114, opened 2021-02-23
Add file upload endpoint for new indexes

tianshengsui:
[ch382]


tests/indexes/test_api.py
```diff
@@ -451,3 +455,68 @@ async def test_delete_index(spawn_job_client, error):
...
+    else:
+        url += "?name=reference.1.bt2"
+
+    resp = await client.post_form(url, data=files)
+
+    if error == "400_name":
+        assert await resp_is.bad_request(resp, "Unsupported index file name")
+        return
+
+    if error == "400_exists":
```
igboyes, line 501: Should be `409`.

https://stackoverflow.com/questions/9269040/which-http-response-code-for-this-email-is-already-registered/9270432#:~:text=409%20%3D%3E%20Conflict%20That%20mean.,exists%20in%20the%20parent%20location.

tests/indexes/test_api.py
```diff
@@ -451,3 +455,68 @@ async def test_delete_index(spawn_job_client, error):
...
+        return
+
+    if error == "422":
+        assert await resp_is.invalid_query(resp, {'name': ['required field']})
+        return
+
+    assert resp.status == 201
+    assert os.listdir(tmpdir / "references" / "bar" / "foo") == ["reference.1.bt2"]
+    snapshot.assert_match(await resp.json())
+    document = await client.db.indexes.find_one("foo")
```
igboyes, line 512: You can use a snapshot here as well.

tests/indexes/test_api.py
```diff
@@ -451,3 +455,68 @@ async def test_delete_index(spawn_job_client, error):
...
+    assert document == {
+        '_id': 'foo',
+        "reference": {
+            "id": "bar"
+        },
+        "user": {
+            "id": "test"
+        },
+        'files': [1]
+    }
```
igboyes, line 522: Need assertion on Postgres state.

igboyes (changes requested):
Most of the new code here has no tests.

igboyes (approved):

PR #2115, opened 2021-02-24
Remove dbs fixture

jakeale:
- Remove `dbs` fixture and associated usages
- Update some snapshots


igboyes (changes requested):
In some of the tests you modified the `dbi` fixtrue isn't actually used. Please clean those up and we'll merge.

igboyes (approved):

PR #2116, opened 2021-02-24
Set `--host` default to None for jobs API.

BlakeASmith:
A `None` host will configure the server to listen on all interfaces.

The current default, "localhost", causes the server to only listen on it's own localhost. This prevents the server from working with docker-compose port forwarding.

### DIFF
--- a/virtool/config.py
+++ b/virtool/config.py
@@ -2,10 +2,10 @@ import asyncio
 import json
 import logging
 import os
-import sys
 
 import click
 import psutil
+import sys
 import uvloop
 
 import virtool.app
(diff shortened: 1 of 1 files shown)


igboyes (commented):
I think a better approach is to keep `localhost` as the default and explicitly set `0.0.0.0` when you want to listen on all addresses.

BlakeASmith:
> I think a better approach is to keep `localhost` as the default and explicitly set `0.0.0.0` when you want to listen on all addresses.

Sounds good. I will set to 0.0.0.0 in the docker-compose config.

PR #2121, opened 2021-02-26
Fix client sentry issue and updated capture exception to be environment dependent

OfficialArms:
The use of raven is outdated and lead to runtime errors

### DIFF
--- a/client/src/js/index.js
+++ b/client/src/js/index.js
@@ -16,6 +16,8 @@ if (!window.virtool.dev) {
     window.Sentry = Sentry;
 }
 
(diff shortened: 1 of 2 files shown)


client/src/js/index.js
```diff
@@ -16,6 +16,8 @@ if (!window.virtool.dev) {
     window.Sentry = Sentry;
 }
 
+window.captureException = error => (window.virtool.dev ? console.error(error) : window.Sentry.captureException(error));
```
igboyes, line 19: What is gained by calling `console.error`?

PR #2126, opened 2021-02-27
Added download to subtraction detail page

ColeVoelpel:
I decided to keep the download consistent with other pages as an Icon.


igboyes (changes requested):
Please use an anchor element to do this. Point it directly at `/download/subtraction/:id`.

igboyes (changes requested):
Please see this recent PR from @ColeVoelpel. I think a text link works best for this. Yours should say _Download FASTA_.

https://github.com/virtool/virtool/pull/2150

igboyes (approved):

PR #2129, opened 2021-03-02
Add a task for storing existing NuVs files

tianshengsui:
[ch171]

- Add a task for storing existing NuVs analysis files to `<data_path>/analyses` and `analysis_files` SQL table
- Spawn the task on application start


virtool/analyses/db.py
```diff
@@ -34,6 +39,86 @@
...
+        async for analysis in db.analyses.find({"workflow": "nuvs"}):
+            analysis_id = analysis["_id"]
+            sample_id = analysis["sample"]["id"]
+            path = virtool.analyses.utils.join_analysis_path(settings["data_path"], analysis_id, sample_id)
+            target_path = os.path.join(settings["data_path"], "analyses", analysis_id)
+
+            try:
+                await self.app["run_in_thread"](os.mkdir, target_path)
+            except FileExistsError:
+                pass
```
igboyes, line 84: Could just use `os.makedirs` here and get rid of `make_analyses_directory`.

virtool/analyses/db.py
```diff
@@ -34,6 +39,86 @@
...
+                    if file == "hmm.tsv":
+                        await self.app["run_in_thread"](
+                            shutil.copy,
+                            os.path.join(path, "hmm.tsv"),
+                            os.path.join(target_path, "hmm.tsv")
+                        )
+                    else:
+                        await self.run_in_thread(virtool.utils.compress_file,
+                                                 os.path.join(path, file),
+                                                 os.path.join(target_path, f"{file}.gz"))
```
igboyes, line 97: This is a good candidate for a separate function.

igboyes (changes requested):
Some of your blocks are very hard to read because they are all jammed together. Try to use blank spaces to separate groups of related lines.

igboyes (changes requested):
* `isdir` check will pass even if only some of the files were moved in a interrupted previous run of the task
* Should use `analysis_files` table to check if files have been written for the analysis
* Write all rows to database in a single transaction **after** all files have been moved. This way, as long as at least one row matches the analysis ID, we can assume the files were written completely successfully for that analysis.

igboyes (approved):

PR #2130, opened 2021-03-03
Add  a Jobs API endpoint for deleting unfinalized subtractions

tianshengsui:
[ch473]

### DIFF
--- a/virtool/subtractions/api.py
+++ b/virtool/subtractions/api.py
@@ -348,7 +348,7 @@ async def job_remove(req: aiohttp.web.Request):
         return not_found()
 
     if "ready" in document and document["ready"]:
-        return bad_request("Only unfinalized subtractions can be deleted")
+        return conflict("Only unfinalized subtractions can be deleted")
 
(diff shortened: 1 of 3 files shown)


virtool/subtractions/api.py
```diff
@@ -332,3 +331,25 @@ async def finalize_subtraction(req: aiohttp.web.Request):
...
+    db = req.app["db"]
+    subtraction_id = req.match_info["subtraction_id"]
+
+    document = await db.subtraction.find_one(subtraction_id)
+
+    if document is None:
+        return not_found()
+
+    if "ready" in document and document["ready"]:
+        return bad_request("Only unfinalized subtractions can be deleted")
```
igboyes, line 351: Think this should be `409 Conflict` as the request conflicts with the subtraction's ready state.

igboyes (changes requested):
Great, although I think a `409` is more specific than `400` in this situation.

igboyes (approved):

PR #2131, opened 2021-03-03
Fix format of `quality` field set by sample finalization

jakeale:


igboyes (approved):
Wow. I totally missed this. Thanks.

PR #2135, opened 2021-03-05
Add Download Route For Index Files

BlakeASmith:


virtool/indexes/api.py
```diff
@@ -323,3 +323,27 @@ async def delete_index(req: aiohttp.web.Request):
...
+
+    reference_id = index_document["reference"]["id"]
+
+    path = Path(req.app["settings"]["data_path"]) / "references" / reference_id / index_id / filename
+
+    if path.exists():
+        return aiohttp.web.FileResponse(path)
+
+    if filename not in FILES:
+        return bad_request(f"{filename} must be one of {FILES}")
```
igboyes, line 348: Should be `404`.

igboyes (changes requested):
Thanks. Need to tweak status code when file name is invalid.

igboyes (approved):

PR #2136, opened 2021-03-06
Fix appearance of index list items

ColeVoelpel:
Added max-widths so the flex container cant evenly distribute the width to smaller components. The max-width % on the strong's styling allows for a larger range of window sizes over a static max-width.

### DIFF
--- a/client/src/js/indexes/components/Item.js
+++ b/client/src/js/indexes/components/Item.js
@@ -40,13 +40,13 @@ const IndexItemTop = styled.h3`
 
     > div {
         flex: 1 0 auto;
-        max-width: 70px;
     }
 `;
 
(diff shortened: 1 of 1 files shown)


client/src/js/indexes/components/Item.js
```diff
@@ -39,6 +40,7 @@ const IndexItemTop = styled.h3`
 
     > div {
         flex: 1 0 auto;
+        max-width: 70px;
```
igboyes, line 43: image

I believe this line is causing this. If you are trying to style `IndexItemIcon`, modify the styles for `StyledIndexItemIcon`.

igboyes (changes requested):

PR #2139, opened 2021-03-08
Add Jobs API upload endpoints for samples files

jakeale:
- Add `/api/samples:id/reads` to upload sample reads
- Add `/api/samples/:id/artifacts` for uploading artifacts created after sample creation
- Add `SQLEnum.to_list()` to work with all SQL enumerated types
- Move `postgres.py` to `pg` module and add simple `get_row()` and `delete_row()` functions for use with any table
- Automatically resolve interface name from SQL model in dispatcher


virtool/samples/files.py
```diff
@@ -0,0 +1,39 @@
...
+                uploaded_at=file["uploaded_at"]
+            )
+
+            reads.append(reads_file)
+
+        session.add_all(reads)
+        await session.flush()
+
+        if len(reads) == 2:
+            reads[0].paired, reads[1].paired = reads[1].id, reads[0].id
```
igboyes, line 63: Remove this value from the model. It is defined on the sample (single source of truth).

virtool/samples/models.py
```diff
@@ -0,0 +1,55 @@
+import enum
+
+from sqlalchemy import Column, Integer, String, DateTime, Enum
+
+from virtool.postgres import Base
+
+
+class ArtifactType(str, enum.Enum):
```
igboyes, line 8: What does multiple inheritance accomplish here?

virtool/samples/models.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    SQL model to store new sample reads files
+
+    """
+    __tablename__ = "sample_reads_files"
+
+    id = Column(Integer, primary_key=True)
+    sample = Column(String)
+    name_on_disk = Column(String)
+    paired = Column(Integer)
```
igboyes, line 49: Remove `paired` field.

virtool/postgres.py
```diff
@@ -63,5 +63,10 @@ def to_dict(self):
 
     return row
 
+@classmethod
+def to_list(cls):
+        return [e.value for e in cls.__members__.values()]
+
```
igboyes, line 69: Indentation problem at line 68.

The  classmethod decorator is better used in this case like so:
```py
Enum.to_list = classmethod(func)
```

Also, you should not modifiy `Enum` as it is imported from another package in the standard library. Subclass enum or find a better solution in the `enum` package documentation. Do a general Google search on a solution for this before writing a subclass of `Enum`.

virtool/postgres.py
```diff
@@ -63,5 +63,10 @@ def to_dict(self):
```
igboyes, line 63: Move this `postgres.py` module into `virtool.pg.utils` with your other code in this PR.

virtool/samples/api.py
```diff
@@ -680,3 +692,36 @@ async def analyze(req):
...
+
+        return aiohttp.web.Response(status=499)
+
+    if files is None:
+        return bad_request("One or more files are not compressed")
+
+    reads = await virtool.samples.files.create_reads_files(pg, sample_id, files)
+
+    headers = {
+        "Location": f"/api/samples/{sample_id}/reads"
```
igboyes, line 730: Should be the resource location:
```
/api/samples/{sample_id}/reads/{filename}
```

virtool/samples/models.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    __tablename__ = "sample_artifacts"
+
+    id = Column(Integer, primary_key=True)
+    sample = Column(String)
+    name = Column(String)
+    name_on_disk = Column(String)
+    size = Column(Integer)
+    type = Column(Enum(ArtifactType))
+    uploaded_at = Column(DateTime)
```
igboyes, line 36: Set column constraints where appropriate. Repeat for `SampleReadsFile`.

virtool/samples/models.py
```diff
@@ -0,0 +1,55 @@
...
+    sam = "sam"
+    bam = "bam"
+    fasta = "fasta"
+    fastq = "fastq"
+    csv = "csv"
+    tsv = "tsv"
+    json = "json"
+
+
+class SampleArtifacts(Base):
```
igboyes, line 23: `SampleArtifact`

virtool/uploads/utils.py
```diff
@@ -43,3 +59,45 @@ async def naive_writer(req, file_path) -> int:
...
+                if size == 0 and not is_gzip_compressed(chunk):
+                    return None
+
+                size += len(chunk)
+                await handle.write(chunk)
+
+            new_file["size"], new_file["uploaded_at"] = size, virtool.utils.timestamp()
+
+            files[filename] = new_file
+            file = await reader.next()
```
igboyes, line 103: Handle one file per request.

virtool/uploads/utils.py
```diff
@@ -43,3 +59,45 @@ async def naive_writer(req, file_path) -> int:
             await handle.write(chunk)
 
     return size
+
+
+async def naive_write_multiple(req: aiohttp.web.Request, file_path: pathlib.Path) -> Optional[Dict[str, int]]:
```
igboyes, line 66: `naive_writer_compressed`

igboyes (changes requested):

jakeale:
Really not sure why it's flagging that, the import was already [removed](https://github.com/jakeale/virtool/blob/de3458bff596e86d187c98e95f17f6510f3c6d4c/virtool/dispatcher/events.py#L2)

virtool/samples/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+class SampleReadsFile(Base):
+    """
+    SQL model to store new sample reads files
+
+    """
+    __tablename__ = "sample_reads_files"
+
+    id = Column(Integer, primary_key=True)
+    sample = Column(String, nullable=False)
```
igboyes, line 45: Should have `name` column. Will always be `reads_1.fq.gz` or `reads_2.fq.gz` at this point.

virtool/samples/models.py
```diff
@@ -0,0 +1,52 @@
...
+    name_on_disk = Column(String, nullable=False)
+    size = Column(Integer)
+    uploaded_at = Column(DateTime)
+
+    def __repr__(self):
+        return f"<SampleReadsFile(id={self.id}, sample={self.sample}, name_on_disk={self.name_on_disk}, " \
+               f"paired={self.paired}, size={self.size}, uploaded_at={self.uploaded_at})>"
```
igboyes, line 52: There is no attribute `paired` on `SampleReadsFile`.

virtool/uploads/utils.py
```diff
@@ -1,45 +1,81 @@
...
             await handle.write(chunk)
 
-    return size
+    new_file.update(
+        {
+            "name_on_disk": path.name,
+            "size": size,
+            "uploaded_at": virtool.utils.timestamp(),
+        }
+    )
```
igboyes, line 79: It is best to limit the scope of `naive_writer` to writing the file and returning values that are directly related to the writing of the file (`size`). Limit functions to one responsibility wherever possible.

You pass in the target `path` to `naive_writer` so that calling scope already knows the `path` and can calculate `name_on_disk` if it sees fit. The timestamp can also be easily calculated (or not) in the calling scope.

It is not a good idea to return a `dict` like you did above. They are a really fuzzy way of passing around data. There is no guarantee that the expected keys have been defined in the code that created the `dict` and it is impossible to type individual values. One solution is to gather your values in an object or dataclass that makes logical sense.

If you have a small number of values, return a tuple:
```
return name, size
```

This is much easier to type and an exception will be raised if a mismatch in the number of items in the tuple does not match. https://stackoverflow.com/questions/1479776/too-many-values-to-unpack-exception

virtool/samples/files.py
```diff
@@ -0,0 +1,63 @@
...
+
+        return artifacts_file
+
+
+async def create_reads_file(pg: AsyncEngine, file: Dict[str, any], sample_id: str):
+    """
+    Create one or two rows in the `samples_reads_files` SQL table that represents uploaded sample reads files.
+
+    :param pg: PostgreSQL AsyncEngine object
+    :param file: A dictionary containing info on an uploaded file
```
igboyes, line 42: Should have `name` arg.

virtool/samples/api.py
```diff
@@ -680,3 +695,92 @@ async def analyze(req):
...
+    pg = req.app["pg"]
+    sample_id = req.match_info["sample_id"]
+
+    reads_file_path = Path(req.app["settings"]["data_path"]) / "samples" / sample_id
+
+    if not await db.samples.find_one(sample_id):
+        return not_found()
+
+    try:
+        if {"reads_1.fq.gz", "reads_2.fq.gz"}.issubset(set(os.listdir(reads_file_path))):
```
igboyes, line 716: This check should be done against the database instead of the filesystem.

igboyes (changes requested):

virtool/samples/models.py
jakeale: Should I simply use `name` over `name_on_disk` since they would both be the same?

igboyes: No. The legacy type sample files are actually called `reads_1.fastq` etc.

jakeale: Oh I see. Right now it doesn't make use of a query, but should I set it up so `name` can be passed in (validated to be either `reads_1.fq.gz` or `reads_2.fq.gz`) and `name_on_disk` can be is the actual filename?

virtool/uploads/utils.py
jakeale: > You pass in the target `path` to `naive_writer` so that calling scope already knows the `path` and can calculate `name_on_disk` if it sees fit. The timestamp can also be easily calculated (or not) in the calling scope.

Right now it is set up to work with reads uploads (where `naive_writer` is only given the path to the folder) so the name of the file is determined by reading the multipart, then a row is created using that. 

If I were to use a required `name` query I can get away with not returning the name, but then `name` and `name_on_disk` would be the exact same (since `name_on_disk`  isn't something like `{id}-reads_1.fq.gz`).

virtool/samples/api.py
jakeale: I'm trying to test this properly but for some reason the database is persisting between tests. It's made evident in the [snapshots](https://github.com/virtool/virtool/pull/2139/files#diff-220a2ff3652084488071586d48831973fb1e59235e35ccbab7fb6f0cd8c55968R750-R775) as well, where the `id` should be 1 for each snapshot. It also creates situations like this when running the tests multiple times in succession:
```
>       assert value == snapshot
E       AssertionError: assert {'id': 12, 'n...: 'test', ...} == {'id': 5, 'na...: 'test', ...}
E         Omitting 5 identical items, use -vv to show
E         Differing items:
E         {'id': 12} != {'id': 5}
E         Use -v to get the full diff
```

igboyes (approved):

PR #2146, opened 2021-03-10
Add column constraints to Task model

tianshengsui:


virtool/tasks/models.py
```diff
@@ -8,15 +8,15 @@ class Task(Base):
     __tablename__ = 'tasks'
 
     id = Column(Integer, primary_key=True)
-    complete = Column(Boolean)
+    complete = Column(Boolean, default=False)
     context = Column(JSONB)
-    count = Column(Integer)
+    count = Column(Integer, default=0)
     created_at = Column(DateTime)
```
igboyes, line 14: Shouldn't this be `nullable=False` too?

igboyes (changes requested):

igboyes (approved):

PR #2148, opened 2021-03-10
Fix app hanging and user lockout on forced password reset

jakeale:
From looking at the middleware code it seems that it isn't a bug that `req["client"]` might resolve to `None`, but it seems like a few different endpoints are expecting it not to be (another example is [this one](https://app.clubhouse.io/virtool/story/445/put-500-error-thrown-on-successful-creation-of-first-user)), so I'm not sure how else this should be approached.


igboyes (approved):
Seems like good solution to me. Thanks for fixing this.

PR #2149, opened 2021-03-10
Rename hash field to key for cache documents

tianshengsui:
Add a migrate function for renaming `hash` field to `key` for existing cache documents


igboyes (changes requested):
I realized this morning that we need to change other uses of `hash` in the source code to `key` now.

Please search through code and update.

igboyes (approved):

PR #2151, opened 2021-03-11
Remove virtool.files modules

tianshengsui:
- Move `MigrateFilesTask` to `virtool.uploads.db`
- Update sample creation endpoint to use `Uploads` SQL table


virtool/uploads/db.py
```diff
@@ -4,12 +4,64 @@
...
+import virtool.tasks.task
 import virtool.uploads.utils
 import virtool.utils
+
+from virtool.types import App
 from virtool.uploads.models import Upload
 
 logger = logging.getLogger("uploads")
 
+PROJECTION = [
```
igboyes, line 16: Shouldn't be used anywhere as Mongo isn't used anymore for this data.

igboyes (changes requested):

virtool/uploads/db.py
tianshengsui: @igboyes We still need `projection` [here](https://github.com/virtool/virtool/blob/f74a5f35edb9947856618cabf3f00c931c071b0d/virtool/db/core.py#L327), the `MigrateFilesTask` still need to use `files` collection in Mongo.

igboyes (approved):

PR #2166, opened 2021-03-17
Fix amount of return values in `naive_writer()` based on given path

jakeale:


igboyes (commented):
Please add a comment on the PR for why you needed to do this.

jakeale:
If this function is used with a complete path (including filename) and `compressed=True` and the compression check fails it will return a `None, None` tuple, but on success it will return simply return `size`. This avoids a potential `ValueError: too many values to unpack (expected 1)` by returning `size` if it's successful and `None` if it fails.

PR #2167, opened 2021-03-17
Create download endpoint for sample files

tianshengsui:


virtool/samples/api.py
```diff
@@ -982,3 +985,78 @@ async def finalize_cache(req):
     processed = virtool.utils.base_processor(document)
 
     return json_response(processed)
+
+
+@routes.jobs_api.get("/api/samples/{sample_id}/reads/{suffix}")
```
igboyes, line 990: Please use these paths instead:
```
/api/samples/:id/reads/reads_1.fq.gz
/api/samples/:id/reads/reads_2.fq.gz
```

igboyes (changes requested):

igboyes (approved):

PR #2170, opened 2021-03-18
Remove CONFIG_PROJECTION and its usages

tianshengsui:
Add tests and docstrings for `virtool.settings` modules

### DIFF
--- a/virtool/settings/db.py
+++ b/virtool/settings/db.py
@@ -1,5 +1,7 @@
 import logging
 
+from typing import Any, Dict
+
 import virtool.settings.schema
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


virtool/settings/db.py
```diff
@@ -8,18 +8,16 @@
...
-    "proc",
-    "mem",
-    "lg_proc",
-    "lg_mem",
-    "sm_proc",
-    "sm_mem"
-)
-
 
 async def ensure(db):
```
igboyes, line 12: `-> Dict[str, Any]`

Please complete typing where required elsewhere.

igboyes (changes requested):

PR #2171, opened 2021-03-18
Use `PUT` instead of `POST` in endpoints that upload reads

jakeale:
* Use `PUT /api/samples/:id/reads/:filename`
* Use `PUT /api/samples/:id/caches/:key/reads/:filename`
* Refactor `naive_writer()` to accept `Callable` instead of `bool`, work only with complete paths
* Remove unused functions in `analyses.files`


igboyes (approved):
Looks good!

PR #2184, opened 2021-03-23
Update analysis projection and `attach_analysis_files`

jakeale:
- Change `subtraction` to `subtractions` in projection
- Remove `files` from projection and refactor where used
- Change `attach_analysis_files` to take a document parameter and return it with a `files` field instead of returning a list

[ch585]


jakeale:
ch[585]

igboyes (approved):

PR #2186, opened 2021-03-24
Add SQL relationship between uploads and sample reads

jakeale:
I decided on a one-to-many relationship since it's likely that a single upload will be associated with many different sample reads, but a sample reads entry will be only be associated with a single upload.
 
[ch604]


igboyes:
This looks great to me. Can you edit the opening comment with a brief rundown of your decision making?

igboyes (approved):

PR #2196, opened 2021-03-26
Update IndexFile model

tianshengsui:


virtool/indexes/models.py
```diff
@@ -24,10 +24,10 @@ class IndexFile(Base):
 
     id = Column(Integer, primary_key=True)
     name = Column(String)
-    reference = Column(String)
+    index = Column(String)
     type = Column(Enum(IndexType))
     size = Column(Integer)
```
igboyes, line 29: Let's also make `name` and `index` non-nullable.

igboyes (changes requested):

igboyes (approved):

PR #2198, opened 2021-03-29
Add task to move pre-SQL sample files records to SQL

jakeale:
* Change `uploads` size column to `BigInteger` to support large file sizes

### DIFF
--- a/tests/samples/test_db.py
+++ b/tests/samples/test_db.py
@@ -577,8 +577,10 @@ async def test_compress_samples_task(mocker, dbi, pg: AsyncEngine, pg_session, s
     ])
 
 
-@pytest.mark.parametrize("legacy, compressed", [(False, False), (True, True), (True, False)])
-async def test_move_sample_files_task(legacy, compressed, dbi, pg, pg_session, snapshot, static_time):
+@pytest.mark.parametrize("legacy", [True, False])
+@pytest.mark.parametrize("compressed", [True, False])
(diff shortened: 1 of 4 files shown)


tests/samples/test_db.py
```diff
@@ -572,3 +575,66 @@ async def compress_reads(app, sample):
...
+                "raw": True,
+                "size": 213889231,
+                "from": {
+                    "id": "vorbsrmz-17TFP120_S21_R1_001.fastq.gz",
+                    "name": "vorbsrmz-17TFP120_S21_R1_001.fastq.gz",
+                    "size": 239801249712,
+                    "uploaded_at": None,
+                },
+            }
+        ]
```
igboyes, line 606: Test that this works for paired samples (`files.length === 2`).

virtool/samples/db.py
```diff
@@ -396,3 +397,81 @@ async def compress_samples(self):
...
+    def __init__(self, app, process_id):
+        super().__init__(app, process_id)
+
+        self.steps = [
+            self.move_sample_files
+        ]
+
+    async def move_sample_files(self):
+        query = {
+            "files": {"$exists": True, "$ne": ""},
```
igboyes, line 451: What does `$ne` do here?

igboyes (changes requested):

igboyes (approved):

PR #2200, opened 2021-03-30
Return list of index files in index responses

tianshengsui:


tests/indexes/test_db.py
```diff
@@ -183,3 +184,49 @@ async def test_update_last_indexed_versions(dbi, test_otu, spawn_client):
...
+            },
+            {
+                'id': 2,
+                'name': 'reference.2.bt2',
+                'index': 'foo',
+                'type': 'bowtie2',
+                'size': 1234567
+            }
+        ]
+    }
```
igboyes, line 232: Use double quotes consistently.

igboyes (changes requested):

igboyes (approved):

PR #2201, opened 2021-03-30
Remove associated uploads when sample is finalized

jakeale:


virtool/samples/api.py
```diff
@@ -505,6 +506,15 @@ async def finalize(req):
...
     })
 
+    async with AsyncSession(pg) as session:
+        rows = (await session.execute(
+            select(SampleReads).filter(SampleReads.sample == sample_id, SampleReads.upload.isnot(None)))).scalars()
+
+    if rows:
+        uploads = {row.upload for row in rows}
+
+        await asyncio.gather(*[virtool.uploads.db.delete(req, pg, upload) for upload in uploads])
```
igboyes, line 516: You should be able to do this in a single query.

igboyes (changes requested):

igboyes (approved):
Great! Thanks for figuring that out.

PR #2207, opened 2021-04-01
Run `MoveSampleFilesTask` on application start

igboyes:
* Run `MoveSampleFilesTask` on application start and change `task_name` to `move_sample_files`. Change `process_id` to `task_id`.
* Log a warning instead of throwing an exception when a `Change` for an unknown dispatcher interface is received.
* Add `download_url` to records in `reads` field of sample response. Remove the legacy `files` field processing that was causing an exception when fetching a sample record.
* Remove some deprecated test snapshots

### DIFF
--- a/virtool/samples/api.py
+++ b/virtool/samples/api.py
@@ -185,7 +185,7 @@ async def get(req):
     document = await virtool.samples.db.attach_artifacts_and_reads(pg, document)
 
     if document["ready"]:
-        for index, file in enumerate(document["reads"]):
+        for file in document["reads"]:
             file.update({
                 "download_url": f"/api/samples/{sample_id}/reads/{file['name']}"
             })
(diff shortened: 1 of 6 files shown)


virtool/samples/api.py
```diff
@@ -180,21 +180,16 @@ async def get(req):
...
-                "download_url": file["download_url"].replace("reads_", f"{snake_case}_"),
-                "replace_url": f"/upload/samples/{sample_id}/files/{index + 1}"
-            })
-
     document = await virtool.subtractions.db.attach_subtractions(db, document)
     document = await virtool.samples.db.attach_labels(pg, document)
     document = await virtool.samples.db.attach_artifacts_and_reads(pg, document)
 
+    if document["ready"]:
+        for index, file in enumerate(document["reads"]):
```
jakeale, line 188:
```suggestion
        for file in document["reads"]:
```

igboyes: Good catch. And I did not know you could suggest changes...

PR #2208, opened 2021-04-01
Modify AddSubtractionFilesTask to write files to SQL

tianshengsui:


virtool/subtractions/files.py
```diff
@@ -34,6 +37,32 @@ async def create_subtraction_file(pg: AsyncEngine, subtraction_id: str, file_typ
         return subtraction_file
 
 
+async def create_subtraction_files(pg: AsyncEngine, subtraction_id: str, files: List[str], path: str):
+    """
+    Create multiple rows in th`subtraction_files` SQL table in a single transaction.
```
igboyes, line 42:
```suggestion
    Create multiple rows in the `subtraction_files` SQL table in a single transaction.
```

virtool/subtractions/db.py
```diff
@@ -79,14 +85,33 @@ async def add_files_field(self):
...
+                path
+            )
+
             await self.db.subtraction.update_one({"_id": subtraction["_id"]}, {
-                "$set": {
-                    "files": files
+                "$unset": {
+                    "files": ""
                 }
             })
```
igboyes, line 116: On second thought. Let's just leave the `files` field in case there is a problem with the SQL switch.

igboyes (changes requested):
Thanks. Minor tweaks needed.

igboyes (approved):

PR #2209, opened 2021-04-01
Split up virtool.app.on_shutdown

jakeale:
* Split up into separate tasks in `virtool.shutdown`
* Add tasks to `virtool.app.create_app`

### DIFF
--- a/virtool/shutdown.py
+++ b/virtool/shutdown.py
@@ -1,0 +1,78 @@
+import logging
+
+import aiohttp.web
+
+import virtool.startup
+
+logger = logging.getLogger(__name__)
+
+
+async def exit_client(app: aiohttp.web.Application):
+    """
+    Attempt to close the async HTTP client session.
+
+    :param app: The application object
+    """
+    try:
+        await app["client"].close()
+    except KeyError:
+        pass
+
+
+async def exit_dispatcher(app: aiohttp.web.Application):
+    """
+    Attempt to close the app's `Dispatcher` object.
+
+    :param app: The application object
+    """
+    try:
+        await app["dispatcher"].close()
+    except KeyError:
+        pass
+
+
+async def exit_executor(app: aiohttp.web.Application):
+    """
+    Attempt to close the app's `ThreadPoolExecutor`.
+
+    :param app: The application object
+    """
+    try:
+        app["executor"].shutdown(wait=True)
+    except KeyError:
+        pass
+
+
+async def exit_process_executor(app: aiohttp.web.Application):
+    """
+    Attempt to close the app's `ProcessPoolExecutor`.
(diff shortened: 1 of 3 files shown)


virtool/shutdown.py
```diff
@@ -0,0 +1,84 @@
...
+async def exit_process_executor(app: aiohttp.web.Application):
+    """
+    Attempt to close the app's `ProcessPoolExecutor`.
+
+    :param app: The application object
+    """
+    try:
+        app["process_executor"].shutdown(wait=True)
+    except KeyError:
+        pass
```
igboyes, line 59: Combine to mirror `init_executors`.
```suggestion
async def exit_executors(app: aiohttp.web.Application):
    """
    Attempt to close the `ThreadPoolExecutor` and `ProcessPoolExecutor`.

    :param app: the application object
    """
    try:
        app["executor"].shutdown(wait=True)
    except KeyError:
        pass
    try:
        app["process_executor"].shutdown(wait=True)
    except KeyError:
        pass
```

igboyes (changes requested):

jakeale:
Just took another look at `virtool.startup` and noticed `init_db` and `init_postgres`. Do you want these to be manually shut down and logged as well?

igboyes:
> Just took another look at `virtool.startup` and noticed `init_db` and `init_postgres`. Do you want these to be manually shut down and logged as well?

I'm not sure. Can you do some research on `motor` and `sqlalchemy` to see if this is really necessary?

jakeale:
> > Just took another look at `virtool.startup` and noticed `init_db` and `init_postgres`. Do you want these to be manually shut down and logged as well?
> 
> I'm not sure. Can you do some research on `motor` and `sqlalchemy` to see if this is really necessary?

I could not find the best practices with `motor` but I found this in the `sqlalchemy` docs:

> Valid use cases for calling Engine.dispose() include:
> 
> * When a program wants to release any remaining checked-in connections held by the connection pool and expects to no longer be connected to that database at all for any future operations.
> 
> * When a program uses multiprocessing or fork(), and an Engine object is copied to the child process, Engine.dispose() should be called so that the engine creates brand new database connections local to that fork. Database connections generally do not travel across process boundaries.
> 
> * Within test suites or multitenancy scenarios where many ad-hoc, short-lived Engine objects may be created and disposed.

If there are any open database connections at the time of garbage collection, the `Engine` will keep them alive. Can't immediately think of a case where this could happen.

https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=dispose#engine-disposal

igboyes (approved):

igboyes:
Thanks for checking into that @jakeale!

PR #2213, opened 2021-04-06
Stop writing `subtraction.files` when subtraction files are uploaded

jakeale:
* Rename `prepare_files_field` to `get_subtraction_files`. I can change this to `attach_subtraction_files` but currently, the function only returns a list of files instead of returning an updated dictionary.

### DIFF
--- a/virtool/subtractions/api.py
+++ b/virtool/subtractions/api.py
@@ -334,10 +334,11 @@ async def finalize_subtraction(req: aiohttp.web.Request):
     updated_document = await db.subtraction.find_one_and_update({"_id": subtraction_id}, {
         "$set": {
             "gc": data["gc"],
-            "ready": True,
-            "files": await virtool.subtractions.utils.get_subtraction_files(pg, subtraction_id)
+            "ready": True
         }
     })
+
+    updated_document["files"] = await virtool.subtractions.utils.get_subtraction_files(pg, subtraction_id)
 
     return json_response(virtool.utils.base_processor(updated_document))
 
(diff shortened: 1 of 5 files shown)


virtool/subtractions/api.py
```diff
@@ -341,7 +335,7 @@ async def finalize_subtraction(req: aiohttp.web.Request):
         "$set": {
             "gc": data["gc"],
             "ready": True,
-            "files": await virtool.subtractions.utils.prepare_files_field(pg, subtraction_id)
+            "files": await virtool.subtractions.utils.get_subtraction_files(pg, subtraction_id)
```
igboyes, line 338: In the story, I meant that the `files` field does not have to exist at all. We can leave all subtraction file data in SQL instead of storing anything in Mongo.

virtool/subtractions/api.py
```diff
@@ -296,7 +290,7 @@ async def edit(req):
         return not_found()
 
     document["linked_samples"] = await virtool.subtractions.db.get_linked_samples(db, subtraction_id)
-    document["files"] = await virtool.subtractions.utils.prepare_files_field(pg, subtraction_id)
+    document["files"] = await virtool.subtractions.utils.get_subtraction_files(pg, subtraction_id)
```
igboyes, line 293: Faster to make database queries concurrently:
```py
linked_samples, files = await gather(
  virtool.subtractions.db.get_linked_samples(db, subtraction_id),
  virtool.subtractions.utils.get_subtraction_files(pg, subtraction_id)
)

document.update({
  "files": files,
  "linked_samples": linked_samples
})
```

igboyes (changes requested):
Thanks for the comment. Please change `prepare_files_field` to

igboyes (approved):

PR #2217, opened 2021-04-06
Create fake analyses when using fake mode

tianshengsui:


virtool/dev/fake.py
```diff
@@ -160,9 +167,58 @@ async def analysis(app: App):
         "subtraction_2",
     ]
 
-    await virtool.analyses.db.create(
-        app, sample_id, ref_id, subtractions, USER_ID, "pathoscope"
-    )
+    file = await virtool.analyses.files.create_analysis_file(app["pg"], "analysis_1", "fasta", "result.fa", 123456)
```
igboyes, line 170: We should put the file with `analysis_2` because it is the finalized one.

```suggestion
    file = await virtool.analyses.files.create_analysis_file(app["pg"], "analysis_2", "fasta", "result.fa", 123456)
```

igboyes (changes requested):

igboyes (approved):

PR #2222, opened 2021-04-08
Raise an AttributeError when the dispatch interface is unknown

tianshengsui:
Decided to raise an `AttributeError` when the dispatch interface is unknown instead of catching it.

Currently, this will cause an error when running `AddSubtractionFilesTask` on application start because `subtraction_files` interface has not been added yet.

### DIFF
--- a/virtool/dispatcher/dispatcher.py
+++ b/virtool/dispatcher/dispatcher.py
@@ -143,22 +143,22 @@ class Dispatcher:
         :param change: the change to dispatch
 
         """
-        try:
-            fetcher = getattr(self._fetchers, change.interface)
-        except AttributeError:
-            logger.warning(f"Unknown dispatch interface: {change.interface}")
+        fetcher = getattr(self._fetchers, change.interface, None)
 
         if change.operation not in (DELETE, INSERT, UPDATE):
             raise ValueError(f"Unknown dispatch operation: {change.operation}")
 
-        async for connection, message in fetcher.fetch(change, self.authenticated_connections):
-            try:
-                await connection.send(message)
-            except RuntimeError as err:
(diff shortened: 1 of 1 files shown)


igboyes (changes requested):
This won't work because we don't want to dispatch for every SQL model. We also don't want to write fetchers for every possible model.

For example, subtraction files are only modified during subtraction creation and the clients don't need to receive an update about each file being added.

In the case that an unknown interface is encountered, please just emit the warning and don't dispatch anything.

tianshengsui:
> This won't work because we don't want to dispatch for every SQL model. We also don't want to write fetchers for every possible model.
> 
> For example, subtraction files are only modified during subtraction creation and the clients don't need to receive an update about each file being added.
> 
> In the case that an unknown interface is encountered, please just emit the warning and don't dispatch anything.

Yes, that makes sense.

virtool/dispatcher/dispatcher.py
```diff
@@ -143,22 +143,22 @@ async def _dispatch(self, change: Change):
         :param change: the change to dispatch
 
         """
-        try:
-            fetcher = getattr(self._fetchers, change.interface)
-        except AttributeError:
-            logger.warning(f"Unknown dispatch interface: {change.interface}")
+        fetcher = getattr(self._fetchers, change.interface, None)
```
igboyes, line 146: Better to just `return` early if we encounter `AttributeError`:
```py
except AttributeError:
  logger.warning(f"Unknown dispatch interface: {change.interface}")
  return
```
This keeps the error and the resulting behaviour in the same place instead of using `None` variables and conditionals.

igboyes (changes requested):

igboyes (approved):

PR #2223, opened 2021-04-08
Add dispatcher fetcher for subtraction files

tianshengsui:


igboyes (commented):
I think this should be closed. Please see my response on #2222.

https://github.com/virtool/virtool/pull/2222#pullrequestreview-631859056

PR #2224, opened 2021-04-08
Add `404` response for analysis document download endpoint

jakeale:


igboyes (approved):
Thanks for catching this!

PR #2225, opened 2021-04-08
Update virtool.subtractions.db.create function

tianshengsui:


tests/subtractions/test_db.py
```diff
@@ -139,14 +139,14 @@ async def test_unlink_default_subtractions(dbi):
     ]
 
 
-async def test_create(dbi, test_random_alphanumeric):
+async def test_create(dbi):
```
igboyes, line 142: Need to retain test of behaviour when `subtraction_id` is not provided.

igboyes (changes requested):

igboyes (approved):

PR #2234, opened 2021-04-12
Add a function for dropping fake Postgres database

tianshengsui:


igboyes (approved):

virtool/shutdown.py
```diff
@@ -75,3 +76,16 @@ async def exit_redis(app: aiohttp.web.Application):
...
+
+
+async def drop_fake_postgres(app: aiohttp.web.Application):
+    """
+    Drop a fake PostgreSQL database if the instance was run with the ``--fake`` option.
+
+    :param app: the application object
+
+    """
+    if app["config"]["fake"]:
```
igboyes, line 88: I want to be very sure we're deleting only fake databases:
```suggestion
    if app["config"]["fake"] and "fake_" in app["config"]["postgres_connection_string"]:
```

igboyes (changes requested):

igboyes (approved):

PR #2237, opened 2021-04-13
Create Fake Samples When In --fake Mode

BlakeASmith:
- Create a fake paired sample which is finalzied
- Create a fake unpaired sample which is finalized
- Create an unfinalized sample without files uploaded
- Add `run_in_thread` pytest fixture

### DIFF
--- a/tests/fixtures/core.py
+++ b/tests/fixtures/core.py
@@ -12,7 +12,6 @@ SAM_50_PATH = os.path.join(sys.path[0], "tests", "test_files", "sam_50.sam")
 
 
 class MockRequest:
-
     def __init__(self):
         self.app = dict()
         self._state = dict()
@@ -52,7 +51,6 @@ def test_files_path():
 @pytest.fixture
 def test_random_alphanumeric(mocker):
     class RandomAlphanumericTester:
-
         def __init__(self):
             self.choices = [
                 "aB67nm89jL56hj34AL90",
@@ -64,7 +62,7 @@ def test_random_alphanumeric(mocker):
                 "KfVw9vD27KGMly2qf45K",
                 "xjQVxIGHKsTQrVisJiKo",
                 "U3cuWAoQ3TDsy0wU7z0l",
-                "9PfsOM1B99KfaMz2Wu3C"
+                "9PfsOM1B99KfaMz2Wu3C",
             ]
 
             self.history = list()
@@ -91,7 +89,9 @@ def test_random_alphanumeric(mocker):
         def next_choice(self):
             return self.choices[-1]
 
-    return mocker.patch("virtool.utils.random_alphanumeric", new=RandomAlphanumericTester())
+    return mocker.patch(
+        "virtool.utils.random_alphanumeric", new=RandomAlphanumericTester()
+    )
 
 
 @pytest.fixture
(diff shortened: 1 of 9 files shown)


virtool/samples/fake.py
```diff
@@ -0,0 +1,147 @@
...
+        db=db,
+        name=" ".join(fake.words(2)),
+        host="Vine",
+        isolate="Isolate A1",
+        locale="",
+        subtractions=subtraction_ids,
+        files=files,
+        notes=fake.text(50),
+        library_type="normal",
+        labels=fake.words(3),
```
igboyes, line 124: Just leave as empty list. The `labels` field is a list of integer label `id`s assigned to the sample. Labels are in SQL.

tests/samples/test_fake.py
```diff
@@ -0,0 +1,53 @@
...
+    if "reads" in fake_sample:
+        for f in fake_sample["reads"]:
+            del f["uploaded_at"]
+
+    return fake_sample
+
+
+@pytest.mark.parametrize("paired", [True, False])
+@pytest.mark.parametrize("finalized", [True, False])
+async def test_create_fake_unpaired(paired, finalized, app, snapshot):
```
igboyes, line 32: Using `static_time` fixture will make all calls to `virtool.utils.timestamp` return the same value and support snapshotting without have to strip keys from the sample document.

virtool/samples/fake.py
```diff
@@ -0,0 +1,147 @@
...
+    return await create_reads_file(
+        pg,
+        name=name,
+        size=size,
+        name_on_disk=path.name,
+        sample_id=sample_id,
+    )
+
+
+def _create_fake_composition(fake: FakerWrapper):
```
igboyes, line 37:
```suggestion
def create_fake_composition(fake: FakerWrapper):
```

virtool/samples/fake.py
```diff
@@ -0,0 +1,147 @@
...
+                    sample_id,
+                    name=f"read_{n}.fq.gz",
+                ) for n in (1, 2)
+            ])
+        else:
+            files.append(await create_fake_read_file(pg,
+                                                     READ_FILES_PATH /
+                                                     "single.fq.gz",
+                                                     sample_id,
+                                                     name="reads.fq.gz"))
```
igboyes, line 111: Don't write `files` field. Sample reads are in SQL now (`sample_reads`) and are "joined" to a sample based on their `sample` column on request.

Just need to create sample read records here.

igboyes (changes requested):
I don't think the sample read files are not copied in a sample data directory in this code. I think we should merge and complete the work in another PR.

tests/fixtures/core.py
```diff
@@ -130,3 +130,11 @@ def sam_line(request):
...
 def example_path():
     return Path(__file__).parent.parent.parent / "example"
+
+
+@pytest.fixture
+def run_in_thread(loop):
+    async def _run_in_thread(func, *args, **kwargs):
+        return loop.run_in_executor(func, *args, **kwargs)
+
+    return _run_in_thread
```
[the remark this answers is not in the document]
BlakeASmith: You are correct, it would not have worked.. I've corrected it now. 

Thanks!

tests/samples/test_fake.py
BlakeASmith: Very nice! That saves alot of pain.

igboyes (approved):

PR #2241, opened 2021-04-15
Don't write a files field in index documents

tianshengsui:


igboyes (changes requested):
This block in `virtool.indexes.api` is not going to work now.
```py
if upload_id in document.get("files", []):
  return conflict("File name already exists")
```
Please fix and also add test for this case as it should have been caught after your changes.

virtool/indexes/utils.py
```diff
@@ -12,3 +21,22 @@ def check_index_file_type(file_name: str) -> str:
...
+    :param filename: the name of the index file
+    :param index: the id of the index
+
+    :return: `None` or a row from `IndexFile` table
+
+    """
+    async with AsyncSession(pg) as session:
+        exists = (await session.execute(
+            select(IndexFile).filter_by(name=filename, index=index)
+        )).scalar()
```
igboyes, line 40: Is there more efficient way than fetching entire row? It is not super important here, but I would like to have an example of how to do this.

virtool/indexes/utils.py
```diff
@@ -12,3 +21,22 @@ def check_index_file_type(file_name: str) -> str:
         return "json"
     else:
         return "bowtie2"
+
+
+async def check_file_exists(pg: AsyncEngine, filename: str, index: str) -> Optional[Base]:
```
igboyes, line 26: It makes more sense for this to return a boolean.

igboyes (changes requested):

virtool/indexes/utils.py
tianshengsui: I haven't found a better solution for this.
There is an [exists()](https://docs.sqlalchemy.org/en/14/orm/query.html?highlight=exists#sqlalchemy.orm.Query.exists) method in SQLAlchemy, but can't work with `AsyncSession`

igboyes: Okay. Thanks for checking.

virtool/indexes/utils.py
```diff
@@ -23,20 +20,20 @@ def check_index_file_type(file_name: str) -> str:
...
+async def check_file_exists(pg: AsyncEngine, filename: str, index: str) -> bool:
     """
     Check if a index file already exists in `IndexFile` SQL table.
 
     :param pg: the PostgreSQL AsyncEngine object
     :param filename: the name of the index file
     :param index: the id of the index
 
-    :return: `None` or a row from `IndexFile` table
+    :return: `True` is file already exists, otherwise return `False`
```
igboyes, line 31:
```suggestion
    :return: `True` if file already exists, otherwise return `False`
```

virtool/indexes/utils.py
```diff
@@ -23,20 +20,20 @@ def check_index_file_type(file_name: str) -> str:
...
 
     """
     async with AsyncSession(pg) as session:
         exists = (await session.execute(
             select(IndexFile).filter_by(name=filename, index=index)
         )).scalar()
 
-    return exists
+    return False if exists is None else True
```
igboyes, line 39:
```suggestion
    return exists is not None
```

igboyes (changes requested):

igboyes (approved):

PR #2242, opened 2021-04-15
Add generic `__repr__` for `Base` model

jakeale:
* Add custom `Base` class with `__repr__` and `to_dict`
* Remove `__repr__` from every individual table definition
* Update snapshots


igboyes (approved):
Great!

PR #2245, opened 2021-04-16
Change index file uploads endpoint from `POST` to `PUT`

jakeale:
* Make endpoint accessible only through Jobs API
* Remove `name` query validators

### DIFF
--- a/virtool/indexes/api.py
+++ b/virtool/indexes/api.py
@@ -252,7 +252,7 @@ async def create(req):
     return json_response(virtool.utils.base_processor(document), status=201, headers=headers)
 
 
-@routes.post("/api/indexes/{index_id}/files")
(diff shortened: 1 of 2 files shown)


virtool/indexes/api.py
```diff
@@ -262,37 +262,31 @@ async def upload(req):
...
     pg = req.app["pg"]
     index_id = req.match_info["index_id"]
+    name = req.match_info["name"]
 
-    errors = virtool.uploads.utils.naive_validator(req)
-
-    if errors:
-        return invalid_query(errors)
+    if name not in FILES:
+        return bad_request("Unsupported index file name")
```
igboyes, line 268: Should be `404`.

igboyes (changes requested):

igboyes (approved):

PR #2249, opened 2021-04-16
Create unique constraints for sample reads uploads

jakeale:
* Add `UniqueConstraint` between `sample` and `name` in `SampleReads` model
* Catch resulting error in upload endpoint and return conflict response

### DIFF
--- a/virtool/samples/api.py
+++ b/virtool/samples/api.py
@@ -7,7 +7,7 @@ import aiohttp.web
 import pymongo.errors
 from aiohttp.web_fileresponse import FileResponse
 from cerberus import Validator
-from sqlalchemy import select
+from sqlalchemy import select, exc
 from sqlalchemy.ext.asyncio import AsyncSession
 
 import virtool.analyses.db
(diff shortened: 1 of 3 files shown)


virtool/samples/models.py
```diff
@@ -41,6 +41,7 @@ class SampleReads(Base):
 
     """
     __tablename__ = "sample_reads"
+    __table_args__ = (UniqueConstraint('sample', 'name'),)
```
igboyes, line 44:
```suggestion
    __table_args__ = (UniqueConstraint("sample", "name"),)
```

virtool/samples/api.py
```diff
@@ -855,20 +855,17 @@ async def upload_reads(req):
...
         return bad_request("File is not compressed")
     except asyncio.CancelledError:
         logger.debug(f"Reads file upload aborted for {sample_id}")
         return aiohttp.web.Response(status=499)
-
-    reads = await virtool.samples.files.create_reads_file(pg, size, name, name, sample_id, upload_id=upload)
+    try:
+        reads = await virtool.samples.files.create_reads_file(pg, size, name, name, sample_id, upload_id=upload)
+    except exc.IntegrityError:
+        return conflict("Reads file is already associated with this sample")
```
igboyes, line 868: Let's be explicit that it is the name that is the problem.
```suggestion
        return conflict("Reads file name is already uploaded for this sample")
```

igboyes (changes requested):

igboyes (approved):

igboyes (approved):

PR #2253, opened 2021-04-19
Create unique constraints for sample reads cache uploads

jakeale:
* Add `UniqueConstraint` between `sample` and `name` in `SampleReadsCache` model

### DIFF
--- a/virtool/samples/api.py
+++ b/virtool/samples/api.py
@@ -7,7 +7,7 @@ import aiohttp.web
 import pymongo.errors
 from aiohttp.web_fileresponse import FileResponse
 from cerberus import Validator
-from sqlalchemy import select
+from sqlalchemy import select, exc
 from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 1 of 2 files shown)


virtool/samples/api.py
```diff
@@ -985,15 +985,12 @@ async def upload_reads_cache(req):
...
-
-    if name in existing_reads:
-        return conflict("Reads file is already associated with this sample")
-
     try:
         size = await virtool.uploads.utils.naive_writer(req, cache_path, is_gzip_compressed)
     except OSError:
         return bad_request("File is not compressed")
+    except exc.IntegrityError:
+        return conflict("Reads file is already associated with this sample")
```
igboyes, line 993: Agree with this tweak? It is not a sample.

```suggestion
        return conflict("File name is already uploaded for this cache")
```

igboyes (changes requested):

igboyes (approved):

PR #2261, opened 2021-04-21
Setup Sentry performance monitoring

jakeale:
* Set `traces_sample_rate` to 0.2 to sample 20% of transactions


igboyes (approved):
Thanks! Seems to be working.

PR #2267, opened 2021-04-22
Move all tasks in db modules to new tasks modules

tianshengsui:
Move all `Task` sub-classes and corresponding tests to dedicated modules


igboyes (commented):
Thanks

igboyes (approved):

PR #2268, opened 2021-04-22
Add query to labels find endpoint to find labels by name

jakeale:


virtool/labels/api.py
```diff
@@ -21,12 +22,11 @@ async def find(req):
     Get a list of all label documents in the database.
 
     """
-    documents = list()
-    async with AsyncSession(req.app["pg"]) as session:
-        result = await session.execute(select(Label))
-        labels = result.scalars().all()
-        for label in labels:
-            documents.append(label.to_dict())
+    find = req.query.get("find")
```
igboyes, line 25: Clashes with the function name. Change to `term` to match other similar handlers (eg. `samples.api.find`)

```suggestion
    term = req.query.get("find")
```

virtool/labels/api.py
```diff
@@ -21,12 +22,11 @@ async def find(req):
...
+
+    labels = await get_rows(req.app["pg"], find, Label)
+
+    documents = [label.to_dict() for label in labels]
 
     documents = await asyncio.gather(*[virtool.labels.db.attach_sample_count(req.app["db"], d) for d in documents])
```
igboyes, line 31: Suggestion: combine so `documents` is iterated only once.

Do `from virtool.labels.db import attach_sample_count` to make line more readable.

virtool/pg/utils.py
```diff
@@ -102,15 +102,12 @@ async def get_row(
...
-        if not row:
-            return None
-
     return row
 
 
 async def get_rows(
     pg: AsyncEngine,
-    query: Union[str, int, bool, SQLEnum],
+    query: Optional[Union[str, int, bool, SQLEnum]],
```
igboyes, line 110: In this case it makes sense to make `query` a keyword argument.

igboyes (changes requested):

igboyes (approved):

igboyes (approved):

PR #2272, opened 2021-04-23
Create reusable `example_path`

jakeale:
* Create `example_path` constant in `virtool.example`
* Reuse constant where its currently being generated
* Change `example_path` test fixture to use constant

### DIFF
--- a/virtool/example/paths.py
+++ b/virtool/example/paths.py
@@ -1,3 +1,0 @@
-from pathlib import Path
-
-example_path = Path(__file__).parent.parent.parent / "example"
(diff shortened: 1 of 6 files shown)


virtool/example/paths.py
```diff
@@ -0,0 +1,3 @@
+from pathlib import Path
+
+example_path = Path(__file__).parent.parent.parent / "example"
```
igboyes, line 3: I think this should be moved to a module at `virtool/example.py`. No need to have another directory for a single constant.

igboyes (changes requested):
One idea on locating the constant. Great otherwise.

igboyes (approved):

PR #2275, opened 2021-04-26
Move OTU FASTA download endpoint

tianshengsui:
Move OTU FASTA download endpoint to `/api/otus/:id.fa`


virtool/downloads/db.py
```diff
@@ -64,7 +64,7 @@ async def generate_sequence_fasta(db, sequence_id: str) -> Tuple[str, str]:
     return virtool.downloads.utils.format_fasta_filename(otu_name, isolate_name, sequence["_id"]), fasta
 
 
-async def generate_otu_fasta(db, otu_id: str) -> Tuple[str, str]:
+async def generate_otu_fasta(db, otu_id: str) -> Tuple[Optional[str], Optional[str]]:
```
igboyes, line 67: This should probably be moved to `virtool.otus.db`. Is it only used in `virtool.otus`?

I want to eventually drop `virtool.downloads`.

tianshengsui: Yes, I can do that.

tests/otus/test_db.py
```diff
@@ -142,3 +142,29 @@ async def test_update_last_indexed_version(dbi, test_otu):
 
     assert otu_2["version"] == 5
     assert otu_2["last_indexed_version"] == 5
+
+
+async def test_generate_virus_fasta(dbi, test_otu, test_sequence):
```
igboyes, line 147:
```suggestion
async def test_generate_otu_fasta(dbi, test_otu, test_sequence):
```

igboyes (changes requested):

igboyes (approved):

PR #2277, opened 2021-04-27
Replace `create_sample_reads_record()` with `create_reads_file()`

tianshengsui:
- Add an option for copying files to data path in `create_reads_file()`
- Replace usage of `create_sample_reads_record()` with `create_reads_file()`
- Fix wrong reads filenames for fake samples
- Add tests

### DIFF
--- a/virtool/samples/files.py
+++ b/virtool/samples/files.py
@@ -1,3 +1,5 @@
+import shutil
+from pathlib import Path
 from typing import Dict, List
 
 from sqlalchemy import select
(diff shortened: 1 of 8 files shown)


virtool/samples/files.py
```diff
@@ -63,28 +66,33 @@ async def create_artifact_file(
 
 
 async def create_reads_file(
```
igboyes, line 68: This function is becoming overly complex. A function should ideally do one thing well. Keep this function SQL-focussed.

I suggest writing another function in `virtool.samples.fake` dedicated to copying the example reads to the sample directory. Call this in `create_fake_sample()` before creating a SQL record for the reads file using `create_reads_file()`.

igboyes (changes requested):
This `create_reads_file` has too many

igboyes (approved):

PR #2278, opened 2021-04-27
Make sample cache file download endpoints available in `caches.api`

jakeale:
* Move sample artifact cache download endpoint to `GET /api/caches/:key/artifact/:filename`
* Make sample reads cache download endpoint to `GET /api/caches/:key/reads/:filename`
* Add `key` column to sample cache file tables and use for lookups


jakeale:
Is this safe to rebase?

igboyes:
Replaced by #2292.

PR #2279, opened 2021-04-27
List downloadable files associated with a subtraction

OfficialArms:
Subtraction files are now individually listed and downloadable from the subtraction detail view

### DIFF
--- a/client/src/js/subtraction/components/Detail.js
+++ b/client/src/js/subtraction/components/Detail.js
@@ -61,7 +61,7 @@ export class SubtractionDetail extends React.Component {
 
         const detail = this.props.detail;
 
-        const hasFiles = get(this.props, "state.files.length", 0) > 0;
+        const hasFiles = get(this.props, "files.length", 0) > 0;
 
         if (!detail.ready) {
             return <LoadingPlaceholder message="Subtraction is still being imported" />;
--- a/client/src/js/subtraction/components/Detail.js
+++ b/client/src/js/subtraction/components/Detail.js
@@ -4,6 +4,7 @@ import React from "react";
 import { connect } from "react-redux";
 import styled from "styled-components";
 import { pushState } from "../../app/actions";
+import { fontWeight } from "../../app/theme";
 import {
     BoxGroup,
     BoxGroupHeader,
@@ -16,7 +17,7 @@ import {
     ViewHeaderIcons,
     ViewHeaderTitle
 } from "../../base";
-import { checkAdminOrPermission } from "../../utils/utils";
+import { byteSize, checkAdminOrPermission } from "../../utils/utils";
 import { getSubtraction } from "../actions";
 import EditSubtraction from "./Edit";
 import RemoveSubtraction from "./Remove";
@@ -27,9 +28,9 @@ const StyledBoxGroupSection = styled(BoxGroupSection)`
 
     a {
         margin-right: auto;
+        font-weight: ${fontWeight.thick};
     }
 `;
-
 const calculateGC = nucleotides => numbro(1 - nucleotides.a - nucleotides.t - nucleotides.n).format("0.000");
 
 export class SubtractionDetail extends React.Component {
@@ -43,7 +44,6 @@ export class SubtractionDetail extends React.Component {
 
     componentDidMount() {
         this.props.onGet(this.props.match.params.subtractionId);
-        console.log("state = ", this.props.state);
     }
 
     handleHide = () => {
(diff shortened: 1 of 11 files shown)


client/src/js/subtraction/components/Detail.js
```diff
@@ -86,18 +108,32 @@ export class SubtractionDetail extends React.Component {
...
+                            <p>Data files available to workflows using this subtraction</p>
+                        </BoxGroupHeader>
+                        {this.props.files.map(file => (
+                            <StyledBoxGroupSection key={file.id}>
+                                <a href={file.download_url}>{file.name}</a>
+                                <strong>{byteSize(file.size)}</strong>
+                            </StyledBoxGroupSection>
+                        ))}
+                    </BoxGroup>
+                )}
```
igboyes, line 124: Break this up into modules.
1. Add module with connected component for listing subtraction files.
2. Add module with list item component only.
3. Make a new directory at `subtraction/components/Detail` and put all detail related component modules in here. Relevant tests also have to move.

client/src/js/subtraction/components/Detail.js
```diff
@@ -39,18 +61,19 @@ export class SubtractionDetail extends React.Component {
 
         const detail = this.props.detail;
 
+        const hasFiles = get(this.props, "files.length", 0) > 0;
```
igboyes, line 64: Instead, render the files `BoxGroup` but show a none found message using the `NoneFound` base components.

Did you have a situation when `files` was undefined?

igboyes (changes requested):

client/src/js/subtraction/components/Detail.js
OfficialArms: `files` was undefined was during testing when a shallow copy of `SubtractionDetail` is being rendered. Using `get` prevented that test from failing and works as a more defensive solution to directly access the file's length.

igboyes: Provide a list of fake file objects during the test. Having an undefined `files` field is an error condition and we need that to be thrown and reported to Sentry if it ever happens. Using `get` makes that potential bug silent.

igboyes:
@OfficialArms There are ESLint/Prettier errors on this. Make sure the output of Webpack is clear.

client/src/js/subtraction/components/Detail/Files.js
```diff
@@ -0,0 +1,31 @@
...
+            ))}
+        </BoxGroup>
+    );
+};
+
+const mapStateToProps = state => {
+    return {
+        files: get(state, "subtraction.detail.files", [])
+    };
+};
```
igboyes, line 29: Return statement can be removed. Reduced example:
```js
=> ({ files: get() }}
```

client/src/js/subtraction/components/Detail/File.js
```diff
@@ -0,0 +1,24 @@
...
+`;
+
+export const SubtractionFile = ({ file }) => {
+    return (
+        <StyledBoxGroupSection>
+            <a href={file.download_url}>{file.name}</a>
+            <strong>{byteSize(file.size)}</strong>
+        </StyledBoxGroupSection>
+    );
+};
```
igboyes, line 24: * Can remove `return` statement.
* Take the individual required props instead of the file object.

client/src/js/subtraction/components/Detail/__tests__/File.test.js
```diff
@@ -0,0 +1,25 @@
...
+    beforeEach(() => {
+        props.file = {
+            download_url: "/api/subtractions/xl8faqqz/files/subtraction.fa.gz",
+            id: 1,
+            name: "foo",
+            size: 36461731,
+            subtraction: "xl8faqqz",
+            type: "fasta"
+        };
+    });
```
igboyes, line 18: A blank line after `beforeEach()`.

client/src/js/subtraction/components/Detail/__tests__/File.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+    beforeEach(() => {
+        props.file = {
+            download_url: "/api/subtractions/xl8faqqz/files/subtraction.fa.gz",
+            id: 1,
+            name: "foo",
+            size: 36461731,
+            subtraction: "xl8faqqz",
+            type: "fasta"
+        };
```
igboyes, line 17: Please follow pattern used in rest of Virtool tests.
```js
// Declare variable without value.
let props;
```

In `beforeEach()`, declare a value for `prop`:
```js
beforeEach(() => {
    props = {
        file: {}
    }
});
```

Of course, this has to change because you are going to use a more parametrized approach for `SubtractionFile`.

igboyes (changes requested):

client/src/js/subtraction/components/Detail/Files.js
```diff
@@ -0,0 +1,27 @@
...
+                <p>Data files available to workflows using this subtraction.</p>
+            </BoxGroupHeader>
+            {files.map(file => (
+                <SubtractionFile file={file} key={file.id} />
+            ))}
+        </BoxGroup>
+    );
+};
+
+const mapStateToProps = state => ({ files: get(state, "subtraction.detail.files", []) });
```
igboyes, line 25: Need to write test for `mapStateToProps`.

igboyes (changes requested):

igboyes (approved):

PR #2283, opened 2021-04-29
Sample Integration

BlakeASmith:
Changes to support the `virtool_workflow` integration test introduced by [#146](https://github.com/virtool/virtool-workflow/pull/146)

### DIFF
--- a/virtool/samples/fake.py
+++ b/virtool/samples/fake.py
@@ -36,8 +36,8 @@ async def create_fake_samples(app: App) -> List[dict]:
     )
     samples.append(await create_fake_sample(
         app,
-        USER_ID,
         SAMPLE_ID_UNPAIRED,
+        USER_ID,
         paired=False,
         finalized=False)
     )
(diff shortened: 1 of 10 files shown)


virtool/samples/fake.py
```diff
@@ -11,14 +11,43 @@
...
+        app,
+        SAMPLE_ID_PAIRED_FINALIZED,
+        USER_ID,
+        paired=True,
+        finalized=True)
+    )
+    samples.append(await create_fake_sample(
+        app,
+        USER_ID,
+        SAMPLE_ID_UNPAIRED,
```
igboyes, line 40: This correct?

igboyes (changes requested):

virtool/samples/fake.py
```diff
@@ -11,14 +11,43 @@
 EXAMPLE_FILES_PATH = Path(__file__).parent.parent.parent / "example"
```
igboyes, line 11: See `virtool.example`. There is an `example_path` constant there.

igboyes (changes requested):

igboyes (approved):

PR #2284, opened 2021-04-29
Remove usage of `sample["files"]` in sample remove endpoint

jakeale:
I'm not sure if this is the correct strategy here, but I assumed that the document in `jobs` wouldn't persist beyond `create_sample` so I did not have it check for `ids` in `task_args`.


igboyes (approved):
Thanks for finding and fixing that!

PR #2285, opened 2021-04-29
Only check OTU existence in OTU FASTA download endpoint

tianshengsui:


igboyes (approved):
Good improvement using `async for`. Thanks.

PR #2312, opened 2021-06-16
Moreiraj/vir 794 race condition when loading settings in

OfficialArms:
`State.account.data` was previously not considered to determine the value of the ready prop. This lead to issues where other components that depended on that data, like the sample creation page, would attempt to access it before it had been loaded.


igboyes:
This needs to be based of the `develop` branch and contain only the commits related to the issue. I suggest:
1. Create new branch from `develop`.
2. Cherry pick e9f0f59144cf3fd9e091e0c9ac9ef9ef0a505b4d from existing branch to new branch.
3. Open new PR.

PR #2313, opened 2021-06-16
Updated `ready` prop to be dependent on `state.settings.data`

OfficialArms:
`State.account.data` was previously not considered to determine the value of the ready prop. This lead to issues where other components that depended on that data, like the sample creation page, would attempt to access it before it had been loaded.

### DIFF
--- a/client/src/js/app/Main.js
+++ b/client/src/js/app/Main.js
@@ -80,8 +80,8 @@ export const Main = ({ ready, onLoad }) => {
     return <LoadingPlaceholder />;
 };
 
-export const mapStateToProps = ({ account: { ready }, settings: { data } }) => ({
-    ready: ready && Boolean(data)
+export const mapStateToProps = state => ({
+    ready: state.account.ready && Boolean(state.settings.data)
 });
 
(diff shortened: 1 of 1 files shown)


client/src/js/app/Main.js
```diff
@@ -74,15 +73,15 @@ export const Main = ({ ready, onLoad }) => {
...
     }
 
     return <LoadingPlaceholder />;
 };
 
-export const mapStateToProps = state => ({
-    ready: state.account.ready && Boolean(keys(state.settings).length)
+export const mapStateToProps = ({ account: { ready }, settings: { data } }) => ({
+    ready: ready && Boolean(data)
 });
```
igboyes, line 85: I don't find this readable. Stick to the signature of `mapStateToProps` used everywhere else in Virtool.
```suggestion
export const mapStateToProps = state => ({
    ready: state.account.ready && Boolean(state.settings.data)
});
```

igboyes (changes requested):

igboyes (approved):

PR #2314, opened 2021-06-18
Allow Zero or More Subtractions in Sample Creation View

OfficialArms:
Creating a sample now supports 0 to N subtractions using the `MultiSelector` component


client/src/js/samples/components/Create/Create.js
```diff
@@ -27,7 +27,7 @@ import ReadSelector from "./ReadSelector";
 import { SampleUserGroup } from "./UserGroup";
 
 const CreateSampleFields = styled.div`
-    display: grid;
+    display: inline;
```
igboyes, line 30: Default is `block`, which is what you actually want here.
```suggestion
```

igboyes (changes requested):

igboyes (approved):

PR #2319, opened 2021-07-13
Remove create_subtraction.py job module

eroberts9789:
Remove `create_sample.py` as jobs are now developed and run outside of Virtool.


igboyes (changes requested):
From Linear issue:
> Remove imports only used in this module

The function `calculate_fasta_gc()` was imported and used in this module. It is the only place it was used other than its test function.

Please remove the function and tests.

igboyes (approved):

PR #2324, opened 2021-07-16
Remove migrate.py jobs module

eroberts9789:
Remove migrate.py module as jobs are developed and run outside of Virtool now.
* Remove related tests and unused code


igboyes (changes requested):
Request handlers shouldn't be removed. Remember those have a decorator like:
```py
@routes.patch("/api/jobs/{job_id}")
```

igboyes (approved):

PR #2335, opened 2021-07-20
Remove dead code from virtool.virtool

eroberts9789:
* Use static code analysis tools `Vulture` and `Pyflakes` to detect unused code
* Remove functions only used in test modules and not in `virtool.virtool` project code

### DIFF
--- a/virtool/app.py
+++ b/virtool/app.py
@@ -1,8 +1,10 @@
+import asyncio
 import logging
 
 import aiohttp.web
 import aiojobs
 import aiojobs.aiohttp
+from virtool.process_utils import create_app_runner, wait_for_restart, wait_for_shutdown
 
 import virtool.db.core
 import virtool.db.migrate
@@ -83,3 +85,21 @@ def create_app(config):
     ])
 
     return app
+
+
+async def run_app(config):
+    app = create_app(config)
+
+    runner = await create_app_runner(
+        app,
+        config["host"],
+        config["port"]
+    )
+
+    _, pending = await asyncio.wait(
+        [wait_for_restart(runner, app["events"]), wait_for_shutdown(runner, app["events"])],
+        return_when=asyncio.FIRST_COMPLETED
+    )
+
+    for task in pending:
+        task.cancel()
--- a/virtool/app.py
+++ b/virtool/app.py
@@ -1,4 +1,3 @@
-import asyncio
 import logging
 
 import aiohttp.web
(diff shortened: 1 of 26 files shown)


virtool/config.py
```diff
@@ -155,57 +107,6 @@ def cli(ctx, data_path, db_connection_string, db_name, dev, force_version, no_se
...
-        "no_check_db": no_check_db,
-        "no_check_files": no_check_files,
-        "no_client": no_client,
-        "no_fetching": no_fetching
-    }
-
-    logger.info("Starting in server mode")
-    asyncio.get_event_loop().run_until_complete(virtool.app.run_app(config))
-
-
```
igboyes, line 208: I don't think this should be removed.

virtool/dev/fake.py
```diff
@@ -177,29 +175,6 @@ async def create_fake_analysis(app: App):
...
-        user_id=USER_ID,
-        rights=rights,
-        job_id="sample_integration_test_job",
-    )
-
-
-async def create_fake_jobs(app: App):
-    await create_sample_integration_test_job(app)
-
-
```
igboyes, line 202: Can this be removed @BlakeASmith?

virtool/logs.py
```diff
@@ -41,32 +41,6 @@ def configure_base_logger(dev: bool, verbose: bool) -> Logger:
...
-        "server.log", maxBytes=1000000, backupCount=3
-    )
-
-    handler.setFormatter(logging.Formatter(get_log_format(verbose), style="{"))
-
-    logger.addHandler(handler)
-
-    return logger
-
-
```
igboyes, line 69: This is called in `config.py`.

virtool/app.py
```diff
@@ -85,21 +83,3 @@ def create_app(config):
...
-        config["port"]
-    )
-
-    _, pending = await asyncio.wait(
-        [wait_for_restart(runner, app["events"]), wait_for_shutdown(runner, app["events"])],
-        return_when=asyncio.FIRST_COMPLETED
-    )
-
-    for task in pending:
-        task.cancel()
```
igboyes, line 105: Called in `config.py`.

igboyes (changes requested):
Some things that I believe shouldn't be removed.

I have requested input from @BlakeASmith on an bit of integration test code.

virtool/dev/fake.py
BlakeASmith: Yes. The test job is now created in `virtool.fake.test_cases.sample_integration`

BlakeASmith (approved):

igboyes (approved):

PR #2340, opened 2021-07-22
Accept None user group during sample creation

eroberts9789:
* When force_group_choice is enabled, accept "" as a valid user group
* Update tests

### DIFF
--- a/tests/samples/test_api.py
+++ b/tests/samples/test_api.py
@@ -298,7 +298,7 @@ class TestCreate:
 
         assert await resp_is.bad_request(resp, "Sample name is already in use")
 
-    @pytest.mark.parametrize("group", [None, "", "diagnostics"])
+    @pytest.mark.parametrize("group", ["", "diagnostics", None])
     async def test_force_choice(self, spawn_client, pg: AsyncEngine, resp_is, group):
         """
         Test that when ``force_choice`` is enabled, a request with no group field passed results in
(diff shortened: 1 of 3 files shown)


tests/samples/test_api.py
```diff
@@ -320,13 +321,22 @@ async def test_force_choice(self, spawn_client, pg: AsyncEngine, resp_is):
...
             "files": [1],
             "subtractions": ["apple"]
-        })
+        }
+        await client.db.groups.insert_one(
+            {"_id": "diagnostics"},
+        )
 
-        assert await resp_is.bad_request(resp, "Group value required for sample creation")
+        if group:
```
BlakeASmith, line 333: If I understand correctly, 

In the case where `group` is the empty string `""` we would want to include it in the request data.  

Here `if group` (equivalent to `if ""`) will evaluate `False`, since `""` is falsy in Python. 

This means that the group `""` would not be added to the request data and the result would be identical to the `None` case.  The test will still pass, but it is not testing that a group of `""` is allowed, instead we are essentially testing the `None` case twice. 

The fix for this is to test `if group is None` instead of just `if group`. This is a common gotcha in python since `""`, `[]`, `{}`, `set()`, 0 all evaluate to False in an if statement.

BlakeASmith (changes requested):

BlakeASmith (approved):

PR #2343, opened 2021-07-27
Lilly/vir 874 Remove is_host on backend

eroberts9789:
* Remove `is_host` key from subtraction records
* Update related tests


BlakeASmith (approved):
Looks good to me.

PR #2344, opened 2021-07-28
Fix Sample Creation API Call

OfficialArms:
Incorrect values are being sent on sample creation. This pull request addresses that issue and supplies new tests to help prevent such issues in the future.

### DIFF
--- a/client/src/js/analyses/components/Create/Create.js
+++ b/client/src/js/analyses/components/Create/Create.js
@@ -46,7 +46,7 @@ export const CreateAnalysis = ({
     const {
         errors,
         references,
-        subtraction,
+        subtractions,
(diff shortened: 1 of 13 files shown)


client/src/js/analyses/components/Create/Create.js
```diff
@@ -32,7 +32,7 @@ export const CreateAnalysis = ({
     hasHmm,
     sampleId,
     show,
-    subtractions,
+    subtractionList,
```
igboyes, line 35: This is not a very descriptive variable name. Let's do something like `allSubtractions` or come up with something better.

igboyes (changes requested):

PR #2346, opened 2021-08-04
Fixed Error on Sample Detail Page Refresh 

OfficialArms:
Added loading placeholder when the results can't yet be retrieved from useFuse


BlakeASmith (approved):
Looks good to me.

PR #2348, opened 2021-08-11
Use aiohttp's Web Exceptions

eroberts9789:
Update `virtool.http.errors.middleware` to return a `web.Response` object with a JSON body 
* "id" value is a snake-case message produced from the "reason" field of the exception. For example the HTTPNotFound error would have "Not Found" as its reason, which would be converted to "not_found".
* "message" is the text passed in when the error is raised. For example when helper functions are used, the message was passed in as `return not_found("Release not found")` and now the text will be given as `raise HTTPNotFound(text="Release not found")`.

### DIFF
--- a/virtool/http/errors.py
+++ b/virtool/http/errors.py
@@ -1,6 +1,6 @@
 from typing import Callable
+
 from aiohttp import web
-from aiohttp.web_exceptions import HTTPNotFound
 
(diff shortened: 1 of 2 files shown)


igboyes (changes requested):
API and non-API requests need to be treated differently.

igboyes (approved):

PR #2358, opened 2021-08-16
Rework RespIs class to improve pytest introspection

eroberts9789:
* Rename fixtures.response to fixtures.test_response so module is included in test suite and assertions are re-written before testing. Rewritten assert statements put introspection information into the assertion failure message
* Add no_content, bad_gateway, and bad_request tests in RespIs class 
* Remove AssertRespIs class to avoid redundancy


igboyes (approved):

igboyes (changes requested):
You shouldn't name the file `test_response.py` that way because it doesn't contain tests.

Files containing fixtures you want available for use should just be imported in `conftest.py`.

igboyes (approved):

PR #2360, opened 2021-08-17
Replace insufficient_rights() helper function with HTTPForbidden exception

eroberts9789:
* Update RespIs.insufficient_rights helper function
* Update tests

### DIFF
--- a/virtool/analyses/api.py
+++ b/virtool/analyses/api.py
@@ -9,6 +9,7 @@ from typing import Any, Dict, Union
 import aiohttp.web
 import aiojobs.aiohttp
 from aiohttp.web import HTTPNoContent, HTTPBadRequest, HTTPNotModified
+from aiohttp.web_exceptions import HTTPForbidden
 
 import virtool.analyses.format
 import virtool.bio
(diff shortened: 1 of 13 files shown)


virtool/analyses/api.py
```diff
@@ -110,7 +110,7 @@ async def get(req: aiohttp.web.Request) -> aiohttp.web.Response:
     read, _ = get_sample_rights(sample, req["client"])
 
     if not read:
-        return insufficient_rights()
+        raise HTTPForbidden(text="Insufficient rights", reason="Insufficient rights")
```
igboyes, line 113: Reuse code by subclassing `HTTPForbidden`.

igboyes (changes requested):

virtool/api/response.py
```diff
@@ -22,6 +23,15 @@ def json_response(data: object, status: int = 200, headers: Optional[dict] = Non
     return resp
 
 
+class HTTPInsufficientRights(HTTPForbidden):
+    def __init__(self, message):
+        super().__init__(text=message, reason="insufficient_rights")
+
+
+def insufficient_rights(message="Insufficient rights"):
+    return HTTPInsufficientRights(message)
```
igboyes: Can we not just have:
```py
class HTTPInsufficientRights(HTTPForbidden):
    def __init__(self, message = "Insufficient rights"):
        super().__init__(text=message, reason="insufficient_rights")
```
Then raise with:
```py
raise HTTPInsufficientRights("Non-default message")
```

Is the helper function necessary?

eroberts9789: Hadn't thought of that, I'll give it a go.

igboyes (approved):

PR #2363, opened 2021-08-17
Replace not_found helper function with aiohttp.web.HTTPNotFound exception

eroberts9789:
* Update tests

### DIFF
--- a/virtool/account/api.py
+++ b/virtool/account/api.py
@@ -7,14 +7,14 @@ requests.
 """
 import aiohttp.web
 from aiohttp.web import HTTPNoContent
-from aiohttp.web_exceptions import HTTPBadRequest
+from aiohttp.web_exceptions import HTTPBadRequest, HTTPNotFound
 
 import virtool.account.db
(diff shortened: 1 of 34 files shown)


virtool/account/api.py
```diff
@@ -182,7 +182,7 @@ async def get_api_key(req: aiohttp.web.Request) -> aiohttp.web.Response:
     document = await db.keys.find_one({"id": key_id, "user.id": user_id}, API_KEY_PROJECTION)
 
     if document is None:
-        return not_found()
+        raise HTTPNotFound(text="Not found")
```
igboyes, line 185: I think we should subclass the aiohttp `HTTPNotFound` exception to make our own `HTTPNotFound` that sets the default value for `text`.

igboyes (changes requested):

virtool/account/api.py
eroberts9789: Sounds good!

igboyes (approved):

PR #2367, opened 2021-08-18
Change all usages of task to workflow

igboyes:
* Types of jobs are now called workflows. Better put, jobs are instances of workflows.
* The front end has been updated to clarify the concepts around workflows and jobs.


BlakeASmith (approved):
Fantastic

PR #2368, opened 2021-08-18
Add InvalidQuery subclass of HTTPUnprocessableEntity exception to replace invalid_query() helper function

eroberts9789:
* Update middleware to add "errors" item to json body of InvalidQuery errors
* Update tests

### DIFF
--- a/virtool/samples/api.py
+++ b/virtool/samples/api.py
@@ -982,7 +982,7 @@ async def upload_cache_artifact(req):
     errors = virtool.uploads.utils.naive_validator(req)
 
     if errors:
-        return InvalidQuery(errors)
+        raise InvalidQuery(errors)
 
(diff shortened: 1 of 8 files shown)


virtool/samples/api.py
```diff
@@ -982,7 +982,7 @@ async def upload_cache_artifact(req):
     errors = virtool.uploads.utils.naive_validator(req)
 
     if errors:
-        return invalid_query(errors)
+        return InvalidQuery(errors)
```
igboyes, line 985:
```suggestion
        raise InvalidQuery(errors)
```
This should be `raise` right?

igboyes (changes requested):

igboyes (approved):

PR #2372, opened 2021-08-19
Only assign first two available subtractions for fake samples

igboyes:
This is more realistic and better supports frontend testing.

Is this going to screw up your integration tests @BlakeASmith?


BlakeASmith (approved):
This shouldn't cause any issues with the integration tests. All the sample integration test does is instantiate the `sample` and `read` fixtures, so it shouldn't matter how many subtractions are attached.

PR #2373, opened 2021-08-19
Added Default Subtraction Editing to Sample Detail View

OfficialArms:


igboyes (changes requested):
Please get rid of the label-style border around the subtraction text. The subtraction options should just have default text styling.

image

PR #2377, opened 2021-08-20
Update error message for email formatting in cerberus schema check

eroberts9789:
Instead of writing separate custom validator, check for email formatting error in virtool.http.schema and update error message

### DIFF
--- a/virtool/validators.py
+++ b/virtool/validators.py
@@ -1,9 +1,10 @@
 import re
 
+from email_validator import validate_email, EmailSyntaxError
+
 import virtool.users.utils
 
 RE_HEX_COLOR = re.compile("^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$")
-RE_EMAIL = re.compile("^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$")
 
 
 def strip(value: str) -> str:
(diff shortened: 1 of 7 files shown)


igboyes:
I think I may have misdirected you on this one. Please see the `check_with` rule in Cerberus and the existing module: https://github.com/virtool/virtool/blob/develop/virtool/validators.py.

This is the preferred solution for this issue.

virtool/validators.py
```diff
@@ -3,6 +3,7 @@
 import virtool.users.utils
```
igboyes, line 3: Can you please update this module with the following:
- tests for all validators
- docstrings and typing
- `from` imports

virtool/validators.py
```diff
@@ -3,6 +3,7 @@
...
 
 RE_HEX_COLOR = re.compile("^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$")
+RE_EMAIL = re.compile("^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$")
```
igboyes, line 6: Did you review the `email-validator` package?

igboyes (changes requested):

virtool/validators.py
eroberts9789: I did, I wasn't sure if you would prefer me to use that or not though

igboyes (approved):

PR #2379, opened 2021-08-20
Update `aioredis` version

BlakeASmith:


igboyes (commented):
Hmm. Something going wrong with resolving the `aioredis^=2.0.0` dep.

BlakeASmith:
PubSub support in `aioredis==2.0.0` leaves some to be desired. The dispatcher relies on channels which have been removed in `2.0.0`. 

For now it is better to stay on version `1.3.1`

PR #2380, opened 2021-08-24
Support index file downloads on web API

eroberts9789:
* Rename download_index_file -> download_index_for_jobs for jobs API
* Add new request download_index_file handler for web requests
* Check requesting user has read access to parent reference

### DIFF
--- a/virtool/indexes/api.py
+++ b/virtool/indexes/api.py
@@ -26,6 +26,7 @@ from virtool.jobs.utils import JobRights
 from virtool.pg.utils import delete_row, get_rows
 from virtool.uploads.utils import naive_writer
 from virtool.utils import compress_json_with_gzip
+from virtool.references.db import check_right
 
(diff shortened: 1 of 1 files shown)


virtool/indexes/api.py
```diff
@@ -160,9 +161,39 @@ async def download_otus_json(req):
...
+
+    index_document = await req.app["db"].indexes.find_one(index_id)
+
+    if index_document is None:
+        raise NotFound()
+
+    # check the requesting user has read access to the parent reference
+    has_read_access = await check_right(req, index_document["reference"], "read")
+    if not has_read_access:
+        raise InsufficientRights()
```
igboyes, line 183:
```suggestion
    if not await check_right(req, index_document["reference"], "read"):
        raise InsufficientRights()
```

igboyes (changes requested):

PR #2382, opened 2021-09-07
Create AddIndexFilesTask to populate index files field for index documents on startup

eroberts9789:
* Use a task to go through all indexes and populate their files fields on startup
* Add tests

### DIFF
--- a/tests/indexes/test_tasks.py
+++ b/tests/indexes/test_tasks.py
@@ -1,11 +1,19 @@
-from virtool.indexes.models import IndexFile
+import pytest
 from sqlalchemy import select
 
+from virtool.indexes.models import IndexFile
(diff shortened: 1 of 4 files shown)


tests/indexes/test_tasks.py
```diff
@@ -44,23 +62,27 @@ async def test_add_index_files(spawn_client, pg_session, static_time, tmp_path):
...
+                "id": 2,
+                "name": "reference.fa.gz",
+                "index": "foo",
+                "type": "fasta",
+                "size": 4096
+            }
+        ]
+
+    else:
+        assert rows == []
```
igboyes: Replace with `snapshot` test fixture.

igboyes (changes requested):

igboyes (approved):

PR #2387, opened 2021-09-13
Add JWT refresh tokens

eroberts9789:
Add JWT to Virtool API and write tests

Upon login create and assign a JWT access token and refresh token. The access token is created in `virtool.jwt.create_access_token()` and includes all info that would normally be stored in the session i.e. admin status, user groups, IP address etc. The access token is stored in an HttpOnly cookie and sent with each request to or response from the API. The access token will be decoded in aiohttp middlewares and supply info to route handlers instead of having to query the database for user information stored in sessions. This access token has a lifetime of 5 minutes and relies on the refresh token to create new access tokens and avoid forcing the user to logout. 

The refresh token only contains the user id, the expiration time and the creation time. It doesn't contain session information but keeps track of the longer term timeline we want the user to be authorized for before forcing a logout. The refresh token is encoded using a secret that is salted with the user's hashed password. This secret + password combo is also what is used to decode the refresh token. When the access token expires, `virtool.jwt.refresh_tokens()` will be called where the refresh token's validity is checked. To check if the refresh token is still valid, the hashed password from the database is added to the refresh secret which means that if the database password has been updated since the token's creation, the validity check will fail and the user will be logged out. This is a quick way to maintain security because if someone's account is breached, all they have to do is change their password to invalidate all remaining access and refresh tokens. If the validity check passes then the user is issued another access token with a lifetime of 5 minutes. 

The idea of salting the refresh secret with the user password is inspired by this video: https://www.youtube.com/watch?v=UA0AIkjI85c. The code referenced in this video is available at https://github.com/benawad/graphql-express-template/tree/22_advanced_jwt_auth.

I decided to store the refresh tokens in the user document because I don't believe they should be sent in each request with the access token. If there is a token capture attack, the attacker shouldn't have access to both tokens. Querying the database every 5 minutes to fetch and validate the refresh token to produce a new access token is still much more efficient than querying the database for each request.

When it comes to interacting with the API it should be very easy to perform permissions and rights checks from here on out by adding middlewares where the access tokens are verified, replaced if they've expired, and are decoded to gather permissions, rights etc. Then the permissions are checked depending on the` req.path` field for each API route handler and errors can be raised if correct permissions aren't included

### DIFF
--- a/tests/users/test_jwt.py
+++ b/tests/users/test_jwt.py
@@ -1,31 +1,9 @@
 import pytest
-from aiohttp.test_utils import make_mocked_request
-from aiohttp.web import Response
-from jwt import encode, decode
+from jwt import decode
 
 from virtool.users.jwt import create_access_token, ACCESS_SECRET, JWT_ALGORITHM, create_refresh_token, REFRESH_SECRET, \
     refresh_tokens
 from virtool.users.utils import hash_password
-
-
-async def handler(request, auth):
-    assert request.headers["AUTHORIZATION"] == auth
-    return Response(headers={"AUTHORIZATION": auth})
-
-
-async def test_auth_header():
-    """
-    Test that jwt in authorization header remains unchanged when sent/returned in a request and response
-    """
-    jwt = encode({"user": "info"}, "secret", algorithm=JWT_ALGORITHM)
-    auth = f"Bearer {jwt}"
-    headers = {
-        "AUTHORIZATION": auth
-    }
-    req = make_mocked_request("GET", "/", headers=headers)
-
-    resp = await handler(req, auth)
-    assert resp.headers["AUTHORIZATION"] == auth
 
 
 @pytest.fixture()
@@ -87,7 +65,7 @@ async def test_login_with_jwt(spawn_client, mocker, iat_time):
     })
 
     assert resp.status == 201
-    encoded = resp.headers["AUTHORIZATION"].split(" ")[1]
+    encoded = resp.cookies.get("access_token").value
     decoded = decode(encoded, ACCESS_SECRET, algorithms=JWT_ALGORITHM)
 
     assert isinstance(decoded.pop("exp"), int)
--- a/tests/users/test_jwt.py
+++ b/tests/users/test_jwt.py
@@ -1,0 +1,20 @@
+from aiohttp.web import Response
+from aiohttp.test_utils import make_mocked_request
+from jwt import encode
+
+
+async def handler(request, auth):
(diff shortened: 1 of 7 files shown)


tests/users/test_jwt.py
```diff
@@ -0,0 +1,166 @@
...
+    if password_change:
+        assert new_token is None
+    else:
+        decoded = decode(encoded, ACCESS_SECRET, algorithms=JWT_ALGORITHM)
+
+        decoded.pop("iat")
+        decoded.pop("exp")
+        decoded.pop("ip")
+
+        assert decoded == {
```
igboyes, line 161: Use double quotes.

virtool/users/jwt.py
```diff
@@ -0,0 +1,115 @@
...
+from jwt import encode, decode, ExpiredSignatureError, InvalidTokenError
+
+from virtool.utils import timestamp
+
+ACCESS_SECRET = "access_secret"
+REFRESH_SECRET = "refresh_secret"
+JWT_ALGORITHM = "HS256"
+
+
+async def create_reset_code_with_jwt():
```
igboyes, line 12: What is this for? Should be removed if unused.

virtool/users/jwt.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    access_token_payload = decode(access_token, ACCESS_SECRET, algorithms=JWT_ALGORITHM, verify_exp=False)
+    user_id = access_token_payload["user"]["id"]
+
+    user_document = await db.users.find_one(user_id)
+    password = user_document["password"]
+
+    try:
+        refresh_token = user_document["refresh_token"]
+        decode(refresh_token, REFRESH_SECRET + str(password), algorithms=JWT_ALGORITHM)
```
igboyes, line 109:
```suggestion
        decode(refresh_token, f"{REFRESH_SECRET}{str(password)}", algorithms=JWT_ALGORITHM)
```

igboyes (changes requested):

virtool/users/jwt.py
```diff
@@ -0,0 +1,115 @@
...
+
+    :param db: The application database client
+    :param ip: The user's IP address
+    :param user_id: The user's username
+    """
+    payload = {
+        "user": {
+            "id": user_id
+        },
+        "ip": ip
```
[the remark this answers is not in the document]
eroberts9789: I can look into it further if that would be helpful but I was just including the same info in the access tokens as is kept in the session. I'm not really sure what it's used for in the sessions atm but I didn't want to rule it out.

igboyes:
Closing for now.

PR #2397, opened 2021-09-28
Fixed subtraction and reference multiSelect lists rendering improperly when the list is empty

ReeceHoffmann:
Changes:

1. After a few revisions ended up fixing VIR-948 by having the SubtractionSelector component directly return a NoneFoundBox if there are no subtractions to be listed. 
2. Added similar code to ReferenceSelector to create a properly formed error message
3. NoneFoundBox was modified to accept and directly render props.children to permit inclusion of links to the correct pages. Assumes that the child will be an inline element otherwise formatting will be off
4. Changed the content checking ternary operator in MultiSelector so that it checks the truthiness of children.length rather than children

### DIFF
--- a/client/src/js/subtraction/components/Selector.js
+++ b/client/src/js/subtraction/components/Selector.js
@@ -13,7 +13,7 @@ export const SampleSubtractionSelector = ({ name, noun, selected, subtractions, 
     ));
     if (!subtractions.length) {
         return (
-            <NoneFoundBox noun={"subtractions"}>
+            <NoneFoundBox noun="subtractions">
                 <BrowserRouter>
                     <Link to="/subtraction"> Create one</Link>.
                 </BrowserRouter>
--- a/client/src/js/subtraction/components/Selector.js
+++ b/client/src/js/subtraction/components/Selector.js
@@ -2,7 +2,7 @@ import { map } from "lodash-es";
 import PropTypes from "prop-types";
 import React from "react";
 import { MultiSelector, MultiSelectorItem } from "../../base/MultiSelector";
-import { Link, BrowserRouter } from "react-router-dom";
+import { Link } from "react-router-dom";
 import { NoneFoundBox } from "../../base/NoneFound";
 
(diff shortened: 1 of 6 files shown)


client/src/js/subtraction/components/Selector.js
```diff
@@ -2,14 +2,24 @@ import { map } from "lodash-es";
...
     const subtractionComponents = map(subtractions, ({ name, id }) => (
         <MultiSelectorItem key={id} name={name} value={id} id={id}>
             <span>{name}</span>
         </MultiSelectorItem>
     ));
-
+    if (!subtractions.length) {
+        return (
+            <NoneFoundBox noun={"subtractions"}>
+                <BrowserRouter>
```
igboyes, line 17: Why did you need `BrowserRouter` here?

client/src/js/subtraction/components/Selector.js
```diff
@@ -2,14 +2,24 @@ import { map } from "lodash-es";
...
 export const SampleSubtractionSelector = ({ name, noun, selected, subtractions, onChange }) => {
     const subtractionComponents = map(subtractions, ({ name, id }) => (
         <MultiSelectorItem key={id} name={name} value={id} id={id}>
             <span>{name}</span>
         </MultiSelectorItem>
     ));
-
+    if (!subtractions.length) {
+        return (
+            <NoneFoundBox noun={"subtractions"}>
```
igboyes, line 16:
```suggestion
            <NoneFoundBox noun="subtractions">
```

igboyes (changes requested):

client/src/js/subtraction/components/Selector.js
ReeceHoffmann: Turns out I don't need it there. Jest was complaining that the Link wasn't wrapped while running test because it lacked the normal context of the router from the parent components. Fixed in latest version by wrapping the whole component in the test rather than adding a new router to the production code.

igboyes (approved):

PR #2398, opened 2021-09-29
Make base SelectBox component accessible

ReeceHoffmann:
Changed select box to render as a button. Slight tweaks to the styling were required to keep the look consistent with the old div version. A helper function was also added to enable the default passing of a default button type to prevent the form from submitting automatically when an option is selected.

### DIFF
--- a/client/src/js/base/SelectBox.js
+++ b/client/src/js/base/SelectBox.js
@@ -2,7 +2,10 @@ import styled from "styled-components";
 import { Box } from "./Box";
 import React from "react";
 
-export const StyledSelectBox = styled(Box)`
+export const SelectBox = styled(Box).attrs(props => ({
+    type: "button",
+    as: "button"
+}))`
     border: 1px ${props => (props.active ? props.theme.color.blue : props.theme.color.greyLight)} solid;
     border-radius: ${props => props.theme.borderRadius.sm};
     box-shadow: ${props => (props.active ? props.theme.boxShadow.inset : "none")};
@@ -18,11 +21,3 @@ export const StyledSelectBox = styled(Box)`
         padding-top: 5px;
     }
 `;
-
-export const SelectBox = props => {
-    return (
-        <StyledSelectBox {...props} type="button" as="button">
-            {props.children}
-        </StyledSelectBox>
-    );
-};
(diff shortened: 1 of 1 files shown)


client/src/js/base/SelectBox.js
```diff
@@ -1,11 +1,13 @@
...
 import { Box } from "./Box";
+import React from "react";
 
-export const SelectBox = styled(Box)`
+export const StyledSelectBox = styled(Box)`
     border: 1px ${props => (props.active ? props.theme.color.blue : props.theme.color.greyLight)} solid;
     border-radius: ${props => props.theme.borderRadius.sm};
     box-shadow: ${props => (props.active ? props.theme.boxShadow.inset : "none")};
-
+    background: white;
```
igboyes, line 9:
```suggestion
    background-color: ${props => props.theme.color.white};
```
1. `background-color` is the more specific property here.
2. Pull color from Styled Components theme.

client/src/js/base/SelectBox.js
```diff
@@ -1,11 +1,13 @@
...
+    text-align: left;
```
igboyes, line 10: Why?

client/src/js/base/SelectBox.js
```diff
@@ -16,3 +18,12 @@ export const SelectBox = styled(Box)`
...
 `;
+
+export const SelectBox = props => {
+    const buttonType = props.type ? props.type : "button";
+    return (
+        <StyledSelectBox {...props} type={buttonType} as="button">
+            {props.children}
+        </StyledSelectBox>
+    );
+};
```
igboyes, line 29: Why can't `button` type just be hardcoded?

igboyes (changes requested):

client/src/js/base/SelectBox.js
ReeceHoffmann: Changes in formatting between the default styling of button and div. Without those lines the buttons look like this:
Screenshot from 2021-10-01 09-16-02

I like how the original looks but I am happy to leave the button styling if preferred.

client/src/js/base/SelectBox.js
ReeceHoffmann: Fixed

client/src/js/base/SelectBox.js
ReeceHoffmann: In retrospect, I think it can be in this case.  Seems unlikely there would be a use case where we want to use a select box to submit. Committed a new version that removes the helper function and has a hardcoded type.

igboyes (approved):

igboyes:
Looks great! Thanks.

PR #2399, opened 2021-10-01
Fix for user page not correctly showing error message when there are no groups

ReeceHoffmann:
Super small fix. Changed the ternary operator in Groups.js to check the truthiness of the length rather than the array itself. Previously implemented errors message now renders as expected.


igboyes (approved):

igboyes:
Test is failing.

ReeceHoffmann:
Failing test was supposed to built to confirm that the component would render an errors message when no groups were present, but the snapshot it was using did not itself contain a NoneFound section. Fixed by updating the snapshot. Screenshot comparing the old and new snapshots below:
Screenshot from 2021-10-04 09-11-38

igboyes (approved):

PR #2437, opened 2021-11-02
Add deterministically generated custom user icons

ReeceHoffmann:
Changes:

- Created new base component `InitialIcon` which when passed the user handle and a size renders a circle of deterministically generated hue with the first two letters of the users name within.
    - `InitialIcon` Expects to be handed the users handle, but id is substituted where handle is not accessible
    - Only color is determined using hsl type where only hue is generated per user. saturation and lightness are kept constant prevent stylistic conflicts with color scheme of the page
- Integrated `InitialIcon` where `identicon` was used previously. 
  - `IntitialIcon` now renders with every instance of the `Attribution` base where a user handle is passed
  - Added `InitialIcon` to `indexes/components/Contributors`
- Updated tests as needed to accommodate the new page element

### DIFF
--- a/client/src/js/base/InitialIcon.js
+++ b/client/src/js/base/InitialIcon.js
@@ -1,0 +1,39 @@
+import { reduce } from "lodash-es";
+import React, { useMemo } from "react";
+import styled from "styled-components";
+import { getColor, getFontSize, getFontWeight, theme } from "../app/theme";
+
+const iconSize = {
+    xs: "20px",
+    sm: "24px",
+    md: "28px",
+    lg: "32px",
+    xl: "48px",
+    xxl: "64px"
+};
+const getIconSize = size => iconSize[size];
+
+export const StyledInitialIcon = styled.span`
+    border-radius: 50%;
+    display: flex;
+    flex: 0 0 auto;
+    justify-content: center;
+    align-items: center;
+    height: ${props => getIconSize(props.size)};
+    width: ${props => getIconSize(props.size)};
+    font-size: ${props => getFontSize(props.size)};
+    font-weight: ${getFontWeight("thick")};
+    color: ${getColor({ color: "white", theme })};
+    background: ${props => `hsl(${props.hash}, 83%, 21%);`};
+`;
+
+const colorHash = (hash, newChar) => (hash << 5) - newChar.charCodeAt(0);
+
+export const InitialIcon = ({ handle, size }) => {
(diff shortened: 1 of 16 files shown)


client/src/js/base/InitialIcon.js
```diff
@@ -0,0 +1,30 @@
...
+    flex: 0 0 auto;
+    justify-content: center;
+    align-items: center;
+    height: ${props => `${props.size}px`};
+    width: ${props => `${props.size}px`};
+    background: ${props => `hsl(${props.hash}, 83%, 21%);`}
+    color: ${getColor({ color: "white", theme })};
+    font-weight: ${getFontWeight("thick")};
+    font-size: ${props => `${props.size / 1.95}px`};
+`;
```
igboyes, line 17: Let's tighten this up.

1. Allow predefined sizes instead of taking `size` in `px` (eg. `sm`, `md`, `lg`). This should fit well in whatever `fontSize` theme value they match with. In other words, a small `InitialIcon` should fit well next to text with `sm` font size.
2. Focus on `font-size` for sizing these. Maybe just use the themed font sizes.
3. Don't use `flex`. Just set varying padding on the `span` depending on `span` to get to the final circle size.

Let me know if you need to discuss.

igboyes (changes requested):

client/src/js/base/InitialIcon.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const InitialIcon = ({ handle, size }) => {
+    let hash = 0;
+    for (let i = 0; i < handle.length; i++) {
+        hash = (hash << 5) - handle.charCodeAt(i);
+        hash = hash & hash;
+    }
```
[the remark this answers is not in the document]
ReeceHoffmann: Fixed. I am used useMemo instead of use callback so that I could also calculate the mod only once, but otherwise it should be functionally identical to use callback. This highlighted code has now been reduced to two lines by using reduce.

client/src/js/base/InitialIcon.js
ReeceHoffmann: Updated to use the same string sizes as used in theme rather than integers. As discussed, new version still uses flexbox due to consistent sizing problems of the padding approach. As suggested in point 2 the font sizes used are the same as those in the theme. Most of the time this looks good, but in the `Account/Profile` view a larger version seemed more fitting.

igboyes (approved):

PR #2438, opened 2021-11-03
Split configuration from settings

eroberts9789:
Separate the `click` configuration information and settings stored in the database and store separately in `app["config"]` and `app["settings"]`. Create a data class for config information and a data class for settings. Update all references to settings and config to use new data classes.

### DIFF
--- a/tests/fixtures/settings.py
+++ b/tests/fixtures/settings.py
@@ -1,5 +1,7 @@
 import pytest
 
+from virtool.settings.db import Settings
+
 
 @pytest.fixture()
 async def test_settings(dbi):
@@ -24,3 +26,8 @@ async def test_settings(dbi):
     }
 
     await dbi.settings.insert_one(settings)
+
+
+@pytest.fixture()
+def settings():
+    return Settings()
--- a/tests/history/test_db.py
+++ b/tests/history/test_db.py
@@ -1,19 +1,18 @@
 import datetime
-from aiohttp.test_utils import make_mocked_coro
 
 import pytest
+from aiohttp.test_utils import make_mocked_coro
 
 import virtool.history.db
 
 
 class TestAdd:
 
-    async def test(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path):
+    async def test(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path, config):
+
         app = {
             "db": dbi,
-            "settings": {
-                "data_path": tmp_path
-            }
+            "config": config
         }
 
         old, new = test_otu_edit
@@ -32,12 +31,10 @@ class TestAdd:
         snapshot.assert_match(change, "change")
         snapshot.assert_match(document, "document")
 
-    async def test_create(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path):
+    async def test_create(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path, config):
         app = {
             "db": dbi,
-            "settings": {
-                "data_path": tmp_path
-            }
+            "config": config
         }
 
         # There is no old document because this is a change document for a otu creation operation.
@@ -61,16 +58,14 @@ class TestAdd:
         snapshot.assert_match(change)
         snapshot.assert_match(document)
 
-    async def test_remove(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path):
+    async def test_remove(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path, config):
         """
         Test that the addition of a change due to otu removal inserts the expected change document.
 
         """
         app = {
             "db": dbi,
-            "settings": {
-                "data_path": tmp_path
-            }
+            "config": config
         }
 
         # There is no new document because this is a change document for a otu removal operation.
@@ -96,7 +91,7 @@ class TestAdd:
 
 
 @pytest.mark.parametrize("file", [True, False])
-async def test_get(file, mocker, snapshot, dbi, tmp_path):
+async def test_get(file, mocker, snapshot, dbi, tmp_path, config):
     await dbi.history.insert_one({
         "_id": "baz.2",
         "diff": "file" if file else {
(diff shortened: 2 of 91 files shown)


pyproject.toml
```diff
@@ -45,6 +45,7 @@ aiojobs = "^0.3.0"
 Cerberus = "^1.3.4"
 PyYAML = "^5.4.1"
 email-validator = "^1.1.3"
+mock = "^4.0.3"
```
igboyes, line 48: 1. Remove this. The same functionality is provided by `pytest-mocker`, which is already a development dependency.
2. Poetry distinguishes between development and non-development dependencies. Since `mock` (if we were going to use it) is just used for testing it is a development dependency.
  
   You can flag development dependencies using [`poetry add --dev`](https://python-poetry.org/docs/cli/#add).

tests/fixtures/settings.py
```diff
@@ -24,3 +25,24 @@ async def test_settings(dbi):
     }
 
     await dbi.settings.insert_one(settings)
+
+
+@pytest.fixture()
+def mock_settings():
```
igboyes, line 31: Just call this `settings`. Why not just return a real `Settings` object?

tests/history/test_db.py
```diff
@@ -1,19 +1,19 @@
...
+from aiohttp.test_utils import make_mocked_coro
 
 import virtool.history.db
 
 
 class TestAdd:
 
-    async def test(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path):
+    async def test(self, snapshot, dbi, static_time, test_otu_edit, test_change, tmp_path, mock_config):
+        config = mock_config
```
igboyes, line 12: See, what is the point of calling the fixture `mock_config`?

More useable to just call it `config`. We know it is a `pytest` fixture because it is an argument in a test method.

igboyes (changes requested):
Can you deal with these changes for now and then I'll re-review?

virtool/samples/db.py
```diff
@@ -328,7 +329,7 @@ async def recalculate_workflow_tags(db, sample_id: str) -> dict:
 
 
 async def remove_samples(
-        db, settings: Dict[str, Any], id_list: List[str]
+        db, config, id_list: List[str]
```
igboyes, line 332: Add typing.

virtool/samples/fake.py
```diff
@@ -102,6 +103,14 @@ async def create_fake_sample(
                     sample_id,
                 )
 
+    settings = Settings()
+    settings.app = app
+    settings.db = db
```
igboyes, line 108: Why is this being done?

virtool/samples/utils.py
```diff
@@ -96,16 +96,16 @@ def join_legacy_read_path(sample_path: Path, suffix: int) -> Path:
     return sample_path / f"reads_{suffix}.fastq"
 
 
-def join_legacy_read_paths(settings: dict, sample):
+def join_legacy_read_paths(config, sample):
```
igboyes, line 99: Add typing.

virtool/settings/db.py
```diff
@@ -1,15 +1,43 @@
...
+    enable_sentry: bool = True
+
+    # Software Updates
+    software_channel: str = "stable"
+
+    # Accounts
+    minimum_password_length: int = 8
+
+    # Reference Settings
+    default_source_types: list = field(default_factory=lambda: ["isolate", "strain"])
```
igboyes, line 38: Remove comments. No necessary and incorrect in the case of `enable_api`.

virtool/subtractions/api.py
```diff
@@ -383,7 +383,7 @@ async def download_subtraction_files(req: aiohttp.web.Request):
     file = result.to_dict()
 
     file_path = (
-        virtool.subtractions.utils.join_subtraction_path(req.app["settings"], subtraction_id)
+        virtool.subtractions.utils.join_subtraction_path(req.app["config"], subtraction_id)
```
igboyes, line 386: Use `from` import.

igboyes (changes requested):
Overall things to fix:
1. Use `from` imports in code you have changed.
2. Fix typing missing for `config` and `settings`.

virtool/samples/fake.py
eroberts9789: I just updates the code that was already there, before the "app" and "db" values were assigned to setting s like this:
`    sample = await create_sample(
        _id=sample_id,
        db=db,
        name=f"Fake {sample_id.upper()}",
        host="Vine",
        isolate="Isolate A1",
        locale="",
        subtractions=subtraction_ids,
        notes=fake.text(50),
        library_type="normal",
        labels=[],
        user_id=user_id,
        group="none",
        settings={
            "app": app,
            "db": db,
            "sample_group_read": True,
            "sample_group_write": True,
            "sample_all_read": True,
            "sample_all_write": True,
        },
    )
`

igboyes: OK. Doesn't make sense to me. Can you check if it is necessary and remove if it isn't?

tests/settings/test_db.py
```diff
@@ -5,6 +5,7 @@ async def test_ensure(dbi, test_settings):
     settings = await virtool.settings.db.ensure(dbi)
 
     assert settings == {
```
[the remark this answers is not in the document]
eroberts9789: Removing the _id in Settings is tricky because in ensure and update I make a Settings object by unpacking the combined existing settings data and new settings data using Settings(**existing_and_new_info) so Settings needs to be able to accept an id parameter because it will be included in the existing settings document.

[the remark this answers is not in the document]
igboyes: Solutions:
* https://docs.python.org/3/library/stdtypes.html#dict.pop
* https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/#suppress-_id-field
   (use `False` for Python)

virtool/samples/fake.py
eroberts9789: I just checked and the "app" and "db" values aren't used so I removed those two lines

igboyes (approved):

PR #2440, opened 2021-11-08
Return dev mode value in root API response

eroberts9789:
Update documentation links in virtool.http.root. Add test to ensure that dev is included in the response from the GET /api request handler.


igboyes (changes requested):
This is a `feat` commit as a new value is being returned by the API.

**refactor**
> To rewrite existing source code in order to improve its readability, reusability or structure without affecting its meaning or behaviour

igboyes (approved):

PR #2444, opened 2021-11-10
Rework UserClient class

eroberts9789:
Add authenticated, force_reset and administrator properties to AbstractClient.

Remove ip property from AbstractClient.

Stop setting req["client"] to None, create a UserClient object where authenticated attribute is False instead.

Update imports.


virtool/http/client.py
```diff
@@ -137,23 +152,27 @@ async def has_right_on_upload(self, upload_id: str, right: Right) -> bool:
...
     @property
-    def ip(self):
-        return self._ip
-
-    def has_permission(self, permission) -> bool:
+    def force_reset(self) -> bool:
         return False
 
+    def has_permission(self, permission: str) -> bool:
+        return self.permissions.get(permission, False)
```
igboyes, line 174: This should just return `False`. Jobs don't have permissions, only rights on specific resources. Right?

igboyes (changes requested):
Looks good. One change though.

igboyes (approved):

PR #2448, opened 2021-11-17
Support login with Azure AD B2C

eroberts9789:
When server is started with --use-b2c option, prompt user login or account creation with B2C.

Retrieve ID token for user once they have logged in via B2C and refresh tokens when they've expired.

Once tokens are retrieved from B2C, create corresponding user document in virtool database.

Add new tests and error handling for B2C operations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,9 @@ aiojobs = "^0.3.0"
 Cerberus = "^1.3.4"
 PyYAML = "^5.4.1"
 email-validator = "^1.1.3"
+PyJWT = "^2.3.0"
+msal = "^1.16.0"
+python-jose = {extras = ["cryptography"], version = "^3.3.0"}
 
 
 [tool.poetry.dev-dependencies]
--- a/virtool/config.py
+++ b/virtool/config.py
@@ -8,8 +8,8 @@ import uvloop
 
 import virtool.app
 import virtool.jobs.main
-from virtool.logs import configure_logs
 from virtool.configuration.config import Config
+from virtool.logs import configure_logs
 
 logger = logging.getLogger(__name__)
 
@@ -136,8 +136,46 @@ def cli(ctx, data_path, db_connection_string, db_name, dev, force_version, no_se
     help="Start with automatic fetching disabled",
     is_flag=True
 )
+@click.option(
+    "--b2c-client-id",
+    help="Azure AD B2C client ID, required for --use-b2c",
+    type=str
+)
+@click.option(
+    "--b2c-client-secret",
+    help="Azure AD B2C client secret, required for --use-b2c",
+    type=str
+)
+@click.option(
+    "--b2c-tenant",
+    help="Azure AD B2C tenant name",
+    type=str
+)
+@click.option(
+    "--b2c-user-flow",
+    help="Azure AD B2C signupsignin user flow, required for --use-b2c",
+    type=str,
+)
+@click.option(
+    "--use-b2c",
+    help="Use Azure AD B2C for authentication",
+    is_flag=True
+)
 @click.pass_context
-def start_server(ctx, host, port, no_check_db, no_check_files, no_client, no_fetching):
+def start_server(
+        ctx,
+        host,
+        port,
+        no_check_db,
+        no_check_files,
+        no_client,
+        no_fetching,
+        b2c_client_id,
+        b2c_client_secret,
+        b2c_tenant,
+        b2c_user_flow,
+        use_b2c
+):
     configure_logs(ctx.obj["dev"], ctx.obj["verbose"])
 
     config = Config(
@@ -156,7 +194,12 @@ def start_server(ctx, host, port, no_check_db, no_check_files, no_client, no_fet
         proxy=ctx.obj["proxy"],
         host=host,
         no_client=no_client,
-        port=port
+        port=port,
+        b2c_client_id=b2c_client_id,
+        b2c_client_secret=b2c_client_secret,
+        b2c_tenant=b2c_tenant,
+        b2c_user_flow=b2c_user_flow,
+        use_b2c=use_b2c
     )
 
     logger.info("Starting in server mode")
--- a/virtool/configuration/config.py
+++ b/virtool/configuration/config.py
@@ -22,3 +22,8 @@ class Config(object):
     port: int = 9950
     fake: bool = False
     fake_path: Path = None
+    b2c_client_id: str = None
+    b2c_client_secret: str = None
+    b2c_tenant: str = None
+    b2c_user_flow: str = None
+    use_b2c: bool = False
--- a/virtool/http/auth.py
+++ b/virtool/http/auth.py
@@ -4,20 +4,33 @@ import aiofiles
 import jinja2
 from aiohttp import BasicAuth
 from aiohttp import web
-from aiohttp.web import Request, Response
+from aiohttp.web import Request, Response, HTTPFound
 from aiohttp.web_exceptions import HTTPUnauthorized
+from jose import ExpiredSignatureError
+from jose.exceptions import JWTClaimsError, JWTError
 
 import virtool.errors
+import virtool.utils
 from virtool.http.client import UserClient
 from virtool.http.utils import set_session_id_cookie
+from virtool.oidc.utils import validate_token, update_jwk_args
+from virtool.users.db import B2CUserAttributes, find_or_create_b2c_user
 from virtool.users.sessions import get_session, create_session, clear_reset_code
-from virtool.utils import hash_key
+from virtool.utils import random_alphanumeric
 
 AUTHORIZATION_PROJECTION = [
     "user",
     "administrator",
     "groups",
     "permissions"
+]
+
+PUBLIC_ROUTES = [
+    "/api/oidc/acquire_tokens",
+    "/api/oidc/refresh_tokens",
+    "/api/oidc/delete_tokens",
+    "/api/account/login",
+    "/api/account/logout"
 ]
 
 
@@ -62,7 +75,7 @@ def decode_authorization(authorization: str) -> Tuple[str, str]:
     return auth.login, auth.password
 
 
-async def authenticate_with_key(req: Request, handler: Callable):
+async def authenticate_with_key(req: Request, handler: Callable) -> Response:
     """
     Authenticate the request with an API key or job key.
 
@@ -78,7 +91,7 @@ async def authenticate_with_key(req: Request, handler: Callable):
     return await authenticate_with_api_key(req, handler, holder_id, key)
 
 
-async def authenticate_with_api_key(req: Request, handler: Callable, user_id: str, key: str):
+async def authenticate_with_api_key(req: Request, handler: Callable, user_id: str, key: str) -> Response:
     db = req.app["db"]
 
     document = await db.keys.find_one({
@@ -102,11 +115,75 @@ async def authenticate_with_api_key(req: Request, handler: Callable, user_id: st
     return await handler(req)
 
 
+async def authenticate_with_b2c(req: Request, handler: Callable) -> Response:
+    """
+    Authenticate requests when req.app["config"].use_b2c is True.
+
+    If no id_token cookie is attached to request, redirect to /api/acquire_tokens
+
+    If id_token cookie is found, attempt to validate to gather user information from claims. If token is expired,
+    redirect to /api/refresh_tokens. If token is invalid for some other reason, redirect to /api/delete_tokens
+
+    find or create user based on token claims, then populate req["cient"] with user information and return the
+    response from the handler.
+
+    :param req: the request to handle
+    :param handler: the handler to call with the request if it is authenticated successfully
+
+    :return: the response
+    """
+    if req.cookies.get("id_token") is None:
+        auth_code_flow = await req.app["run_in_thread"](
+            req.app["b2c"].msal.initiate_auth_code_flow,
+            ["email"],
+            f"http://localhost:9950/api/oidc/acquire_tokens",
+            random_alphanumeric(8)
+        )
+        req.app["b2c"].auth_code_flow = auth_code_flow
+        return HTTPFound(auth_code_flow["auth_uri"])
+
+    try:
+        token_claims = await validate_token(req.app, req.cookies.get("id_token"))
+    except ExpiredSignatureError:
+        return HTTPFound("/api/oidc/refresh_tokens")
+    except (JWTClaimsError, JWTError):
+        await update_jwk_args(req.app, req.cookies.get("id_token"))
+        return HTTPFound("/api/oidc/delete_tokens")
+
+    user_attributes = B2CUserAttributes(
+        display_name=token_claims["name"],
+        given_name=token_claims["given_name"],
+        family_name=token_claims["family_name"],
+        oid=token_claims["oid"]
+    )
+
+    user_document = await find_or_create_b2c_user(req.app["db"], user_attributes)
+
+    req["client"] = UserClient(
+        db=req.app["db"],
+        administrator=user_document["administrator"],
+        force_reset=False,
+        groups=user_document["groups"],
+        permissions=user_document["permissions"],
+        user_id=user_document["_id"],
+        authenticated=True
+    )
+
+    return await handler(req)
+
+
 @web.middleware
-async def middleware(req, handler):
+async def middleware(req, handler) -> Response:
+    """
+    Handle requests based on client type and authentication status.
+
+    :param req: the request to handle
+    :param handler: the handler to call with the request if it is authenticated successfully
+    :return: the response
+    """
     db = req.app["db"]
 
-    if req.path == "/api/account/login" or req.path == "/api/account/logout":
+    if req.path in PUBLIC_ROUTES:
         req["client"] = UserClient(
             db=db,
             administrator=False,
@@ -122,6 +199,9 @@ async def middleware(req, handler):
     if req.headers.get("AUTHORIZATION"):
         if can_use_key(req):
             return await authenticate_with_key(req, handler)
+
+    if req.app["config"].use_b2c:
+        return await authenticate_with_b2c(req, handler)
 
     # Get session information from cookies.
     session_id = req.cookies.get("session_id")
@@ -186,7 +266,7 @@ async def index_handler(req: Request) -> Response:
     """
     requires_first_user = not await req.app["db"].users.count_documents({})
 
-    requires_login = req["client"].authenticated is False
+    requires_login = not req.app["config"].use_b2c and not req["client"].authenticated
 
     path = req.app["client_path"] / "index.html"
 
--- a/virtool/http/routes.py
+++ b/virtool/http/routes.py
@@ -5,6 +5,7 @@ from aiohttp.web_exceptions import HTTPUnauthorized
 from aiohttp.web_routedef import RouteTableDef
 
 from virtool.api.response import json_response
+from virtool.http.auth import PUBLIC_ROUTES
 from virtool.users.utils import PERMISSIONS
 
 
@@ -42,10 +43,10 @@ def protect(
         async def wrapped(req):
             client = req["client"]
 
-            if not public and not client.authenticated:
+            if not public and not client.authenticated and req.path not in PUBLIC_ROUTES:
                 raise HTTPUnauthorized(text="Requires authorization")
 
-            if client is None or not client.administrator:
+            if not client.authenticated or not client.administrator:
                 if admin:
                     return json_response({
                         "id": "not_permitted",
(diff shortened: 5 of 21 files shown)


virtool/config.py
```diff
@@ -159,6 +203,31 @@ def start_server(ctx, host, port, no_check_db, no_check_files, no_client, no_fet
         port=port
     )
 
+    if use_b2c:
+        if config.port != 9950:
+            logger.fatal("registered endpoints for Azure AD B2C require server to run on port 9950")
+            sys.exit(1)
```
igboyes, line 209: Here, you are tying implementation in Virtool to the configuration used in our specific tenant. This is not portable. In reality, the registered endpoints will be a FQDN like `https://app.virtool.ca/token`, not at IP + port combination.

virtool/config.py
```diff
@@ -159,6 +203,31 @@ def start_server(ctx, host, port, no_check_db, no_check_files, no_client, no_fet
...
+
+        msal_app = msal.ClientApplication(
+            client_id=b2c_client_id,
+            authority=f"https://{b2c_tenant}.b2clogin.com/{b2c_tenant}.onmicrosoft.com/{b2c_user_flow}",
+            client_credential=b2c_client_secret,
+        )
+
+        os.environ["B2C_CLIENT_ID"] = b2c_client_id
+        os.environ["B2C_CLIENT_SECRET"] = b2c_client_secret
+        os.environ["B2C_AUTHORITY"] = f"https://{b2c_tenant}.b2clogin.com/{b2c_tenant}.onmicrosoft.com/{b2c_user_flow}"
```
igboyes, line 226: Why? This seems a bad practice to me. An application shouldn't be overriding environmental variables potentially set in the shell.

virtool/config.py
```diff
@@ -159,6 +203,31 @@ def start_server(ctx, host, port, no_check_db, no_check_files, no_client, no_fet
...
+
+        config.use_b2c = True
```
igboyes, line 228: `Config` represents values provided to the application at runtime. You shouldn't be mutating it.

I'm not sure what is being accomplished here either...

virtool/config.py
```diff
@@ -159,6 +203,31 @@ def start_server(ctx, host, port, no_check_db, no_check_files, no_client, no_fet
...
+        config.msal_app = msal_app
```
igboyes, line 229: Again, not a configuration value. Probably should be at `app["msal"]`

virtool/config.py
```diff
@@ -136,8 +139,49 @@ def cli(ctx, data_path, db_connection_string, db_name, dev, force_version, no_se
...
+    help="Keycloak client secret, required for --use-oidc"
+)
+@click.option(
+    "--b2c-user-flow",
+    type=str,
+    envvar="B2C_USER_FLOW"
+)
 @click.pass_context
-def start_server(ctx, host, port, no_check_db, no_check_files, no_client, no_fetching):
+def start_server(
```
igboyes, line 171: All of the setting up b2c stuff in here should go in a aiohttp startup function.

igboyes (changes requested):
First set of changes and comments.

Additionally, any code the involves making a request needs to be asyncio-ified, otherwise these requests will block the entire application while they resolve. Easiest way is to `run_in_thread`.

virtool/config.py
eroberts9789: OK, I just thought it would be an easy way to update these values on startup every time, I'll take this out and just check that those values are still there without passing them in.

virtool/config.py
eroberts9789: I wanted to add some sort of flag to the config values to indicate when the server is running in "b2c mode". I suppose I could remove this and just check for app["msal"] once I make that change.

igboyes: Isn't that going to come from:
```py
@click.option(
    "--use-b2c",
    help="Use OIDC for authentication",
    is_flag=True
)
```

eroberts9789: Yes, but I'm saving it in config because I reference it later to check that we are in fact supporting b2c login to avoid "force_login" in `http.auth.index_handler()`. Isn't it also a value provided to the application at runtime?

virtool/config.py
igboyes: Yes. Remove those lines permanently. The deployer is expected to configure this all properly.

pyproject.toml
```diff
@@ -45,6 +45,10 @@ aiojobs = "^0.3.0"
 Cerberus = "^1.3.4"
 PyYAML = "^5.4.1"
 email-validator = "^1.1.3"
+requests = "^2.26.0"
```
igboyes, line 48: Can this be removed?

virtool/configuration/config.py
```diff
@@ -21,4 +23,6 @@ class Config(object):
     no_client: bool = False
     port: int = 9950
     fake: bool = False
-    fake_path: Path = None
\ No newline at end of file
+    fake_path: Path = None
+    use_b2c: bool = False
+    msal_app: ClientApplication = None
```
igboyes, line 28: This should be removed, correct?

virtool/http/auth.py
```diff
@@ -20,6 +26,14 @@
     "permissions"
 ]
 
+UNAUTHENTICATED_ROUTES = [
+    "/api/account/login",
+    "/api/account/logout",
+    "/api/acquire_tokens",
+    "/api/refresh_tokens",
+    "/api/delete_tokens"
+]
```
igboyes, line 35: Can be replaced with `public` kwarg on route decorators. See comment further down.

BTW: Use `tuple` for constants that are [sequences](https://docs.python.org/3/library/stdtypes.html). They are immutable, unlike lists.

virtool/http/auth.py
```diff
@@ -102,11 +116,76 @@ async def authenticate_with_api_key(req: Request, handler: Callable, user_id: st
...
+        auth_code_flow = await req.app["run_in_thread"](
+            req.app["msal"].initiate_auth_code_flow,
+            ["email"],
+            f"http://localhost:9950/api/acquire_tokens",
+            random_alphanumeric(8)
+        )
+        req.app["auth_code_flow"] = auth_code_flow
+        return HTTPFound(auth_code_flow["auth_uri"])
+
+    else:
```
igboyes, line 147: Not necessary to use `else` here.

virtool/oidc/api.py
```diff
@@ -0,0 +1,90 @@
...
+import jwt
+from aiohttp.web import Request, Response
+from aiohttp.web_exceptions import HTTPFound
+
+from virtool.http.routes import Routes
+
+routes = Routes()
+
+
+@routes.get("/api/acquire_tokens")
```
igboyes, line 13: Use the `public` kwarg instead of storing a list of "unauthenticated routes".
```suggestion
@routes.get("/api/acquire_tokens", public=True)
```

virtool/http/auth.py
```diff
@@ -102,11 +116,76 @@ async def authenticate_with_api_key(req: Request, handler: Callable, user_id: st
...
+    Handle requests based on client type and authentication status.
+
+    :param req: the request to handle
+    :param handler: the handler to call with the request if it is authenticated successfully
+    :return: the response
+    """
     db = req.app["db"]
 
-    if req.path == "/api/account/login" or req.path == "/api/account/logout":
+    if req.path in UNAUTHENTICATED_ROUTES:
```
igboyes, line 188: This should be doable with `public` as well. I'm not sure why explicit checking of `req.path` was done here.

virtool/http/routes.py
```diff
@@ -42,7 +43,7 @@ def decorator(handler):
         async def wrapped(req):
             client = req["client"]
 
-            if not public and not client.authenticated:
+            if not public and not client.authenticated and req.path not in UNAUTHENTICATED_ROUTES:
```
igboyes, line 46: Use `public` only. Not `UNAUTHENTICATED_ROUTES`.

virtool/startup.py
```diff
@@ -349,6 +351,26 @@ async def init_version(app: typing.Union[dict, Application]):
     app["version"] = version
 
 
+async def init_b2c(app: Application):
+    b2c_client_id = os.environ.get("B2C_CLIENT_ID")
+    b2c_client_secret = os.environ.get("B2C_CLIENT_SECRET")
+    b2c_tenant = os.environ.get("B2C_TENANT")
+    b2c_user_flow = os.environ.get("B2C_USER_FLOW")
+    b2c_authority = os.environ.get("B2C_AUTHORITY")
```
igboyes, line 359: Why are we still using environmental variables here?

virtool/oidc/utils.py
```diff
@@ -0,0 +1,65 @@
...
+async def update_jwk_args(token: jwt, app: Application) -> JWKArgs:
+    """
+    Update jwk_args and store in environment variable
+
+    :param token: JWT token to validate and decode
+    :param app: the app object
+    :return: jwk_args
+    """
+    header = await app["run_in_thread"](get_unverified_header, token)
+    authority = os.environ.get("B2C_AUTHORITY")
```
igboyes, line 49: Why are we still using an environmental variable here?

virtool/oidc/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    try:
+        jwk_args = app["jwk_args"]
+    except KeyError:
+        jwk_args = await update_jwk_args(token, app)
+
+    return decode(
+        token,
+        asdict(jwk_args),
+        algorithms=["RS256"],
+        audience=os.environ.get("B2C_CLIENT_ID")
```
igboyes, line 36: Why are we still using an environmental variable here?

virtool/oidc/utils.py
```diff
@@ -0,0 +1,65 @@
...
+@dataclass
+class JWKArgs:
+    kty: str
+    kid: str
+    use: str
+    n: str
+    e: str
+
+
+async def validate_token(token: jwt, app: Application) -> dict:
```
igboyes, line 19: Convention in Virtool:
```suggestion
async def validate_token(app: Application, token: jwt) -> dict:
```

virtool/oidc/api.py
```diff
@@ -0,0 +1,90 @@
...
+        payload = jwt.decode(req.cookies.get("id_token"), options={"verify_signature": False})
+
+        user_account = [account for account in accounts if payload["oid"] == account["local_account_id"]][0]
+
+        if user_account:
+            result = await req.app["run_in_thread"](
+                req.app["msal"].acquire_token_silent,
+                [],
+                user_account,
+                os.environ.get("B2C_AUTHORITY")
```
igboyes, line 69: Why are we still using an environmental variable here?

virtool/oidc/api.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Once tokens are acquired, redirect user back to Virtool homepage.
+
+    If error occurs during token retrieval, delete tokens to restart process.
+
+    :param req: the request to handle
+    :return: the response
+    """
+    auth_response = {key: value[0] for key, value in parse_qs(req.url.query_string).items()}
+    try:
```
igboyes, line 26:
```suggestion
    auth_response = {key: value[0] for key, value in parse_qs(req.url.query_string).items()}
    
    try:
```

virtool/oidc/api.py
```diff
@@ -0,0 +1,90 @@
...
+            req.app["msal"].acquire_token_by_auth_code_flow,
+            req.app["auth_code_flow"],
+            auth_response
+        )
+    except ValueError:
+        resp = HTTPFound("/api/delete_tokens")
+        return resp
+
+    resp = HTTPFound("/")
+    if "id_token" in result:
```
igboyes, line 37:
```suggestion
    resp = HTTPFound("/")
    
    if "id_token" in result:
```

virtool/http/auth.py
```diff
@@ -186,7 +268,9 @@ async def index_handler(req: Request) -> Response:
     """
     requires_first_user = not await req.app["db"].users.count_documents({})
 
-    requires_login = req["client"].authenticated is False
+    requires_login = False
+    if req.app["config"].use_b2c is False and req["client"].authenticated is False:
+        requires_login = True
```
igboyes, line 273:
```suggestion
    requires_login = not req.app["config"].use_b2c and not req["client"].authenticated
```

virtool/oidc/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    :param token: JWT token to validate and decode
+    :param app: the app object
+    :return: jwk_args
+    """
+    header = await app["run_in_thread"](get_unverified_header, token)
+    authority = os.environ.get("B2C_AUTHORITY")
+    resp = await app["client"].get(f"{authority}/discovery/v2.0/keys", allow_redirects=True)
+    jwks_keys = json.loads(await resp.text())
+
+    jwk = [key for key in jwks_keys["keys"] if key["kid"] == header["kid"]][0]
```
igboyes, line 53: This is not very efficient.

You are iterating through the entire sequence even if you get a `kid` match in the first item. Move code into separate function that iterates through `jwks_keys` and returns the first matching key.

Also, is it supposed to be `jwk_keys`?

virtool/startup.py
```diff
@@ -349,6 +351,26 @@ async def init_version(app: typing.Union[dict, Application]):
...
+
+    if not b2c_client_id or not b2c_client_secret or not b2c_tenant or not b2c_user_flow or not b2c_authority:
```
igboyes, line 361: https://docs.python.org/3/library/functions.html#all
```suggestion
    if not all([b2c_client_id, b2c_client_secret, b2c_tenant, b2c_user_flow, b2c_authority]):
```

igboyes (changes requested):

virtool/startup.py
eroberts9789: I thought you meant that it was fine to feed in environment variables, just not overwrite them to maintain unidirectional flow. These are taken from the shell. Should I only take them as click options?

virtool/oidc/utils.py
eroberts9789: I'll move to a separate function. JWKS stands for Json Web Key Set so I think I should just replace with `jwks` alone.

virtool/http/auth.py
```diff
@@ -20,6 +26,8 @@
     "permissions"
 ]
 
+UNAUTHENTICATED_ROUTES = [("GET", "/api/oidc")]
```
igboyes, line 29:
```suggestion
PUBLIC_ROUTES = [("GET", "/api/oidc")]
```

virtool/oidc/api.py
```diff
@@ -0,0 +1,91 @@
...
+
+    try:
+        result = await req.app["run_in_thread"](
+            req.app["msal"].acquire_token_by_auth_code_flow,
+            req.app["auth_code_flow"],
+            auth_response
+        )
+    except ValueError:
+        resp = HTTPFound("/api/oidc/delete_tokens")
+        return resp
```
igboyes, line 34:
```suggestion
        return HTTPFound("/api/oidc/delete_tokens")
```
or `raise`.

virtool/oidc/api.py
```diff
@@ -0,0 +1,91 @@
...
+            if "id_token" in result:
+                resp = HTTPFound("/")
+                resp.set_cookie("id_token", result.get("id_token"), httponly=True)
+                return resp
+
+    return HTTPFound("/api/oidc/delete_tokens")
+
+
+@routes.get("/api/oidc/delete_tokens", public=True)
+async def delete_tokens(req: Request) -> Response:
```
igboyes, line 82: Is this tested?

virtool/oidc/utils.py
```diff
@@ -0,0 +1,78 @@
...
+@dataclass
+class JWKArgs:
+    kty: str
+    kid: str
+    use: str
+    n: str
+    e: str
+
+
+async def validate_token(app: Application, token: jwt) -> dict:
```
igboyes, line 18: Is this tested/testable?

virtool/oidc/utils.py
```diff
@@ -0,0 +1,78 @@
...
+        n=jwk["n"],
+        e=jwk["e"]
+    )
+
+    app["jwk_args"] = jwk_args
+
+    return jwk_args
+
+
+async def fetch_jwk(jwks: dict, kid: str) -> dict:
```
igboyes, line 67: Naming is not accurate. Nothing is fetched. `get_matching_jwk()`?

igboyes (changes requested):
There are a lot of new keys being added to `app`. I think it might be better to create a class with these related variables as attributes.

```py
# Example
app["b2c"].msal
app["b2c"].authority
```

virtool/oidc/utils.py
eroberts9789: I'll add a test.

virtool/oidc/api.py
eroberts9789: I'll add a test

virtool/http/auth.py
```diff
@@ -186,7 +263,9 @@ async def index_handler(req: Request) -> Response:
     """
     requires_first_user = not await req.app["db"].users.count_documents({})
 
-    requires_login = req["client"].authenticated is False
+    requires_login = False
+    if not req.app["config"].use_b2c and not req["client"].authenticated:
+        requires_login = True
```
igboyes, line 268:
```suggestion
    requires_login = not req.app["config"].use_b2c and not req["client"].authenticated
```

virtool/startup.py
```diff
@@ -47,6 +49,12 @@
 logger = logging.getLogger("startup")
 
 
+@dataclass
+class B2CInfo:
```
igboyes, line 53: "Info" is not accurate. `ClientApplication` is a collection of methods and attributes abstracting access to the B2C API.

```suggestion
class B2C:
```

virtool/oidc/utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+    jwk_args = JWKArgs(
+        kty=jwk["kty"],
+        kid=jwk["kid"],
+        use=jwk["use"],
+        n=jwk["n"],
+        e=jwk["e"]
+    )
+
+    app["jwk_args"] = jwk_args
```
igboyes, line 62: Put in `B2C`.

igboyes (changes requested):
Hopefully, this will be the last review.

One more thing that needs to be taken care of is _b2c_ vs _ad_. Lets settle on only using only _b2c_ when talking about the MS authentication support.

Examples of what I mean:
```py
# Should be `b2c_oid` and `b2c_user_attributes`.
document.update({
  "ad_oid": ad_user_attributes.oid
})

# Should be `B2CUserAttributes`.
class ADUserAttributes:
  pass
```

virtool/http/auth.py
```diff
@@ -102,11 +110,77 @@ async def authenticate_with_api_key(req: Request, handler: Callable, user_id: st
...
+    Handle requests based on client type and authentication status.
+
+    :param req: the request to handle
+    :param handler: the handler to call with the request if it is authenticated successfully
+    :return: the response
+    """
     db = req.app["db"]
 
-    if req.path == "/api/account/login" or req.path == "/api/account/logout":
+    if (req.method, os.path.split(req.path)[0]) in PUBLIC_ROUTES \
```
igboyes, line 181: What is this doing?
```
os.path.split(req.path)[0]
```

virtool/http/auth.py
```diff
@@ -102,11 +110,77 @@ async def authenticate_with_api_key(req: Request, handler: Callable, user_id: st
...
+    :return: the response
+    """
+    if req.cookies.get("id_token") is None:
+        auth_code_flow = await req.app["run_in_thread"](
+            req.app["b2c"].msal.initiate_auth_code_flow,
+            ["email"],
+            f"http://localhost:9950/api/oidc/acquire_tokens",
+            random_alphanumeric(8)
+        )
+        req.app["auth_code_flow"] = auth_code_flow
```
igboyes, line 137: Put in `B2C`.

virtool/oidc/api.py
```diff
@@ -0,0 +1,90 @@
...
+
+    :param req: the request to handle
+    :return: the response
+    """
+    auth_response = {key: value[0] for key, value in parse_qs(req.url.query_string).items()}
+
+    try:
+        result = await req.app["run_in_thread"](
+            req.app["b2c"].msal.acquire_token_by_auth_code_flow,
+            req.app["auth_code_flow"],
```
igboyes, line 29: Put in `B2C`.

igboyes (changes requested):

virtool/http/auth.py
eroberts9789: I wrote it this way to mimic the check for `PUBLIC_ROUTES` in `virtool.jobs.auth` this way it just checks that the first piece of the `req.path` is `/api/oidc`

igboyes: There is too much potential here for unintentionally making routes public that should be protected. Stick to a static list of method-route combos.

igboyes (approved):

PR #2455, opened 2021-11-26
Rework naming for init and exit functions

eroberts9789:
Rename "init" -> "startup".

Rename "exit" -> "shutdown".

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -4,7 +4,7 @@ import pytest
 from aiohttp.client import ClientSession
 from aiohttp.web import Application
 
-from virtool.startup import init_executors, init_http_client
+from virtool.startup import startup_executors, startup_http_client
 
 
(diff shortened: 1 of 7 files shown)


igboyes (approved):

virtool/tasks/task.py
```diff
@@ -29,12 +29,12 @@ def __init__(self, app, task_id):
         self.errored = False
         self.temp_dir = virtool.utils.get_temp_dir()
 
-    async def init_db(self):
+    async def startup_db(self):
```
igboyes, line 32: Not an `aiohttp` startup function. Revert change, please.

igboyes (changes requested):

virtool/tasks/task.py
eroberts9789: Done!

PR #2473, opened 2021-12-06
chore: use aiohttp==3.8.1

eroberts9789:
Update aiohtttp


igboyes (changes requested):
This should be a `fix` as it resolves security issues caused by the previous version of the library.

igboyes (approved):

PR #2478, opened 2021-12-08
Allow configuration of root url

eroberts9789:
Add --base-url option to server command.

Include base_url in Location header values.

Update tests.


virtool/http/headers.py
```diff
@@ -12,3 +14,16 @@ async def headers_middleware(req, handler):
     resp.headers["Server"] = "Virtool"
 
     return resp
+
+
+async def format_location(request: Request, response: Response):
```
igboyes, line 19: Change to `on_prepare_location` so we known this is a signal handler elsewhere in code.

virtool/http/headers.py
```diff
@@ -12,3 +14,16 @@ async def headers_middleware(req, handler):
...
+    """Format location header to include base_url"""
+    try:
+        location = response.headers.get("Location")
+        base_url = request.app["config"].base_url
+
+        if location and base_url not in location:
+            response.headers["Location"] = f"{base_url}{location}"
+
+    except AttributeError:
+        pass
```
igboyes, line 29: Why is this required?

tests/fixtures/client.py
```diff
@@ -106,13 +108,18 @@ async def func(
             enable_api=False,
             groups=None,
             permissions=None,
-            use_b2c=False
+            use_b2c=False,
+            base_url="https://app.virtool.ca",
+            add_route_table: RouteTableDef = None
```
igboyes, line 113: Should be `addon_route_table`. Add is a verb which makes this sound like it should be a function.

igboyes, line 113:
```suggestion
            add_route_table: Optional[RouteTableDef] = None
```

igboyes (changes requested):
Thanks. Please resolve comments and I will merge.

virtool/http/headers.py
```diff
@@ -12,3 +14,12 @@ async def headers_middleware(req, handler):
     resp.headers["Server"] = "Virtool"
 
     return resp
+
+
+async def on_prepare_location(request: Request, response: Response):
```
igboyes, line 19: Convention elsewhere in VT is `req` and `resp`.

igboyes (changes requested):

PR #2483, opened 2021-12-10
Use Generated ID For Fake User

BlakeASmith:


igboyes (changes requested):
We have to focus on the result of `fix` and `feat` commits now that they are used to generate release notes.

Can you squash the last two with a message like `fix: prevent error during fake user generation` and include the details in other lines in the message.

BlakeASmith:
> We have to focus on the result of `fix` and `feat` commits now that they are used to generate release notes.
> 
> Can you squash the last two with a message like `fix: prevent error during fake user generation` and include the details in other lines in the message.

Yes, it has been done

PR #2484, opened 2021-12-11
Use aiojobs==1.0.0

eroberts9789:


igboyes (approved):

igboyes (changes requested):
Commit message is not appropriate for release notes.

This PR resolves unhandled errors on server shutdown. Please verify that the process shuts down without error and fix the commit message.

PR #2486, opened 2021-12-13
Stop storing download URLs in MongoDB or Postgres

eroberts9789:
Update tests.

### DIFF
--- a/virtool/samples/db.py
+++ b/virtool/samples/db.py
@@ -74,12 +74,13 @@ RIGHTS_PROJECTION = {
 }
 
 
-async def attach_artifacts_and_reads(pg: AsyncEngine, document: dict) -> dict:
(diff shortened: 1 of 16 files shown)


igboyes (changes requested):
Please make sure the endpoints still return the URL in the response. It should be computed on request and include the `backend_url`.

virtool/samples/db.py
```diff
@@ -74,12 +74,13 @@
 }
 
 
-async def attach_artifacts_and_reads(pg: AsyncEngine, document: dict) -> dict:
+async def attach_artifacts_and_reads(pg: AsyncEngine, document: dict, no_download_urls: bool = False) -> dict:
```
igboyes, line 77: Why is `no_download_urls` necessary?

They should be included in all responses, but not stored in the database. Maybe I'm missing something.

igboyes (changes requested):

PR #2489, opened 2021-12-15
Set created_at field during subtraction creation

eroberts9789:
Update tests.


igboyes (approved):
Looks good. Thanks for adding more tests!

PR #2496, opened 2021-12-22
Resolve KeyError during API authentication

eroberts9789:
Administrator, force_reset and groups values aren't stored in API key documents. Assign values to create UserClient object for API key users instead of trying to gather from API key document.


igboyes (changes requested):
Yes, `force_reset` should be `False`, but let's get the other two values from the user document associated with the key.

igboyes (approved):

igboyes:
@eroberts9789 Please improve commit message.

Something like _prevent error when authenticating with API key_. Need to mention the error is specific to API keys.

igboyes (approved):

PR #2497, opened 2021-12-23
Sort subtraction find results alphabetically by name

eroberts9789:


igboyes (changes requested):
There is a much better way to do this. Review this block.
```py
    data = await paginate(
        db.subtraction,
        db_query,
        req.query,
        base_query=BASE_QUERY,
        sort="_id",
        projection=PROJECTION
    )
```

virtool/subtractions/api.py
```diff
@@ -60,14 +61,14 @@ async def find(req):
         async for document in db.subtraction.find({**db_query, **BASE_QUERY}, ["name", "ready"]):
             documents.append(base_processor(document))
 
-        return json_response(documents)
+        return json_response(await sort_alphabetically(documents))
```
igboyes, line 64: https://motor.readthedocs.io/en/stable/tutorial-asyncio.html#querying-for-more-than-one-document

igboyes (changes requested):

PR #2512, opened 2022-02-01
Restore support for exporting analysis results as csv and xlsx

ReeceHoffmann:
Small fixes. Seems likely that the `document` dict changed shape at some point and this was missed in the updates. Also changed the the formatted response to be sent in body rather than text, as the text field does not accept the byte format used for the .xlsx files. 

Final csv file (in excel)
image

Final xlsx file (in excel)
image

(Input is a pathoscope run on reads_1.fq.gz and reads_2.fq.gz combined)


igboyes (approved):
Ah. That's what it was.

PR #2545, opened 2022-04-27
Update otu sequence add/edit api to accept empty values for segment

ReeceHoffmann:
Changes the otu add/edit APIs to accept a null value as a valid option for the segment parameter. Prevents a possible problem where it would be impossible to create or edit sequences where a segment is not defined.

### DIFF
--- a/tests/otus/__snapshots__/test_api.ambr
+++ b/tests/otus/__snapshots__/test_api.ambr
@@ -2849,10 +2849,10 @@
     'version': 0,
   }
 ---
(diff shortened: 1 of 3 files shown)


ReeceHoffmann:
Not sure exactly what is going on, but the test that failed yesterday ran perfectly fine today with no changes to the code. Test runs with no issues when run locally every time that I have tried it. Not sure if this is a github fluke or if the test itself can give variable results with the same code.

igboyes (approved):

PR #2557, opened 2022-05-11
chore: Replace app[run_in_thread]

ryanfang5:


tests/test_app.py
```diff
@@ -35,7 +36,7 @@ async def test_startup_executors():
     def func(*args):
         return sum(args)
 
-    result = await app["run_in_thread"](func, 1, 5, 6, 2)
+    result = await virtool.utils.run_in_thread(func, 1, 5, 6, 2)
```
igboyes, line 39: This test doesn't make sense anymore as you remove `app["run_in_thread"]`.

Add a test for your new `run_in_thread` utility function that ensure the following:
* It is _awaitable_.
* It takes `args` and `kwargs` as expected and passes them to the `func`.
* The result of `func` is returned by `run_in_thread`.

igboyes (changes requested):
In addition the single comment, please change imports to use `from`:

_Old_
```py
import virtool.utils

virtool.utils.run_in_thread()
```
_New_
```py
from virtool.utils import run_in_thread

run_in_thread()
```
We previously used the old style of import to support mocking in `pytest`. This is no longer necessary here.

igboyes (approved):
Excellent!

PR #2563, opened 2022-05-17
Attempt restoring dispatcher connection when disconnected

ReeceHoffmann:
Changes:
  - Adds a new function to redis.py which accepts the channel string and a redis object and attempts to create a subscription to the specified channel. 
  - Modifies dispatch listener's `__anext__` method to handle silently attempt to resubscribe on disconnects and terminate the server if re-subscription is not possible after 2 minutes.
  - Minor changes to tests as needed


virtool/redis.py
```diff
@@ -74,3 +74,27 @@ async def check_version(redis: Redis) -> Optional[str]:
             return version
 
     return None
+
+
+async def resubscribe(redis: Redis, redis_channel_string: str) -> Channel:
```
igboyes, line 79: Don't put types in variable names: ``redis_channel_string`` to ``redis_channel_name``.

https://docs.quantifiedcode.com/python-anti-patterns/readability/putting_type_information_in_a_variable_name.html

igboyes (changes requested):

igboyes (approved):
Perfecto!

PR #2564, opened 2022-05-19
feat: support job archival on API

ryanfang5:
Changed ready test in uploads/api slightly as there may have been a bug

### DIFF
--- a/virtool/jobs/data.py
+++ b/virtool/jobs/data.py
@@ -92,7 +92,11 @@ class JobsData:
         }
         """
         state = query.get("state")
+        archive = query.get("archived")
         term = query.get("find")
+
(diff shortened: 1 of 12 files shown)


virtool/jobs/data.py
```diff
@@ -223,6 +230,32 @@ async def acquire(self, job_id: str):
 
         return await processor(self._db, {**document, "key": key})
 
+    async def archive(self, job_id: str):
```
igboyes, line 233: Add return value type annotation using `->`.

igboyes (changes requested):
Almost perfect. Excellent work.

igboyes (approved):

PR #2571, opened 2022-05-31
feat: add virtool data model and group module

ryanfang5:


virtool/data_model/group.py
```diff
@@ -0,0 +1,9 @@
+from virtool_core.models.group import Group
+
+VirtoolGroup = Group
+
+
+class VTGroup(VirtoolGroup):
+    """
+    Internal model for group used in Virtool.
+    """
```
igboyes, line 9:
```suggestion
VirtoolGroup = Group
"""Internal model for group used in Virtool."""
```

igboyes (changes requested):
I think my example was a bit confusing. The suggested change in my review should work.

PR #2573, opened 2022-06-01
feat: plan api auto documentation

ryanfang5:


virtool/groups/oas.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+from pydantic import BaseModel, Field
+
+
+class SamplePermissions(BaseModel):
```
igboyes, line 5: Why does this start with `Sample`? Change name to `EditGroupSchema` as noted in`api.py`.

With this schema, we need to make sure that the request doesn't set some permissions to `None` in the database. In other words, if our request body is `{modify_hmm: true}`, are all the other permissions going to stay as they are or be set to `None`?

virtool/groups/oas.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class SampleGroupID(BaseModel):
+    """
+    Group id for creating a new Group.
+    """
+    group_id: str = Field(description="a unique id and display name for the group")
+
+
+class SamplePermission(BaseModel):
```
igboyes, line 26: Change name to `EditGroupSchema` as noted in `api.py`.

igboyes (changes requested):
I think we can safely call the request models `SomethingSchema`. This will give us more naming possibilities and jives with the fact that we are essentially defining a JSON schema with these models. Agree?

virtool/groups/oas.py
ryanfang5: Using permissions = data.permissions.dict(exclude_unset=True) will create a dict that only contains the permissions we want to update, excluding the rest.

ryanfang5:
Changed group_id in CreateGroupSchema to type constr which allows you to strip white-space, set to lowercase, and be non-empty.

igboyes (approved):

PR #2582, opened 2022-06-10
feat: support admin and permission configuration in class based views

ryanfang5:


ryanfang5:
Needed to add temporary attributes to req in virtool/http/routes.py to make tests pass since they are checked in the new middleware regardless even if they arent converted to class based views yet.

igboyes (approved):

PR #2593, opened 2022-06-17
Aman260/vir 1374 auto doc history api endpoints

buddy326:


buddy326:
I'm not sure why the tests are failing immediately because of not being able to import the history module from virtool core. I had no issues with this locally so not sure what needs to be changed.

igboyes (changes requested):

igboyes (approved):

PR #2599, opened 2022-07-04
chore: add auto-doc for HMM API endpoints

buddy326:


virtool/hmm/api.py
```diff
@@ -27,167 +34,206 @@
...
+        return json_response(data)
 
-@routes.get("/hmms/status")
-async def get_status(req):
-    """
-    Get the status of the HMM data. Contains the following fields:
+    @permissions(Permission.modify_hmm)
+    async def delete(self) -> Union[r204, r403]:
+        """
+        Delete all unreferenced HMMs and hide the rest.
```
igboyes, line 72: Let's update this since it is going to face users of the API in the documentation.

Right now it is describing an implementation detail instead of the behaviour as it affects the user.
```
Delete all installed HMM data.

This won't break analyses that reference the installed HMM data.
```

igboyes (changes requested):

PR #2600, opened 2022-07-04
chore: move OTU file download endpoints

buddy326:


igboyes:
> Are there tests for the old request handlers?

@buddy326 I've check. There are tests. Can you handle this so we can merge?

igboyes (approved):

PR #2602, opened 2022-07-04
chore: use HistoryMethod enum instead of string literals

buddy326:


igboyes (changes requested):
Can you think of a way to improve this by slightly modifying `virtool.history.db.add()`?

igboyes (approved):

PR #2613, opened 2022-07-13
fix: return correct job created_at time

ReeceHoffmann:
**Changes:**
  - Modifies _find_beta to define `created_at` using the first status's timestamp rather than the last. 
  - Updates the relevant snapshots


igboyes (changes requested):
Ah-ha! Good solve.

PR #2617, opened 2022-07-19
ci: update ci to use custom oas

ryanfang5:


igboyes:
Should be `ci` commit type.

igboyes (approved):

PR #2620, opened 2022-07-20
feat: ensure name is returned in group API responses

buddy326:


virtool/groups/oas.py
```diff
@@ -19,17 +20,15 @@ class EditPermissionsSchema(BaseModel):
 
 class CreateGroupSchema(BaseModel):
     """
-    Group id for creating a new Group.
+    Group id and name for creating a new Group.
```
igboyes, line 23:
```suggestion
    A schema for requests to create groups.
```

igboyes (changes requested):

igboyes (approved):

PR #2625, opened 2022-07-21
chore: add auto-doc for uploads API endpoints

buddy326:


virtool/uploads/oas.py
```diff
@@ -0,0 +1,75 @@
...
+                        "administrator": True,
+                        "handle": "mrott",
+                        "id": "ihvze2u9",
+                    },
+                }
+            ]
+        }
+
+
+class CreateUploadResponse(UploadMinimal):
```
igboyes, line 30: I don't think this should be minimal...

Once you fix that (unless I am missing something), I believe this is exactly the same as `UploadResponse`.

igboyes (changes requested):

virtool/uploads/api.py
```diff
@@ -12,89 +15,158 @@
...
-        {
-            "documents": await apply_transforms(
-                uploads, [AttachUserTransform(req.app["db"])]
-            )
-        }
-    )
+
+@routes.view("/uploads/{id}")
+class UploadView(PydanticView):
+    async def get(self) -> Union[r200[UploadResponse], r404]:
```
igboyes, line 114: Typing doesn't match what is actually returned by the endpoint.

igboyes (changes requested):

PR #2629, opened 2022-07-21
feat: add support for updating group name in PATCH endpoint

buddy326:


igboyes (changes requested):

virtool/groups/data.py
```diff
@@ -63,12 +63,12 @@ async def create(self, group_id: str) -> Group:
 
         return Group(**base_processor(document), users=[])
 
-    async def update(self, group_id: str, permissions_update: Dict[str, str]) -> Group:
+    async def update(self, group_id: str, data: Dict[str, Union[str, dict]]) -> VTGroup:
```
igboyes, line 66: I want the data layer methods to take the Pydantic objects. That way we have typing on the `data` parameter.

virtool/groups/data.py
```diff
@@ -78,10 +78,22 @@ async def update(self, group_id: str, permissions_update: Dict[str, str]) -> Gro
         if not permissions:
             raise ResourceNotFoundError
 
+        permissions_update = data["permissions"]
+
+        name = group_id
+
+        if "name" in data:
+            name = data["name"]
```
igboyes, line 86: If name is not in data, just do nothing. The response model already populates `name` with `id` if no name is provided.

Please ask me if this is not clear. It is important.

virtool/groups/oas.py
```diff
@@ -79,8 +79,14 @@ class EditGroupSchema(BaseModel):
         default={},
     )
 
+    name: Optional[constr(min_length=1)] = Field(
+        description="a name for the group", default=None
```
igboyes, line 83: Remove `default` value. If a name is not provided, it will be excluded from the update using `exclude_unset`.

virtool/groups/oas.py
```diff
@@ -79,8 +79,14 @@ class EditGroupSchema(BaseModel):
...
+    )
+
     class Config:
-        schema_extra = {"example": {"permissions": {"create_ref": True}}}
+        schema_extra = {
+            "example": {"permissions": {"create_ref": True}, "name": "Gobblers"}
```
igboyes, line 88: This is not appropriate for public API documentation... 🤔
```suggestion
            "example": {"permissions": {"create_ref": True}, "name": "Technicians"}
```

igboyes (changes requested):

virtool/groups/data.py
buddy326: So permissions update should not be passed in? Instead it should be data: `EditGroupsSchema`?

igboyes: `data: EditGroupsSchema`

Yes, that's it.

PR #2635, opened 2022-08-02
Auto-doc samples API endpoints

buddy326:


igboyes:
The commit dd8f62c should not be in this PR.

virtool/samples/oas.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+class EditSampleSchema(BaseModel):
+    name: Optional[constr(strip_whitespace=True, min_length=1)] = None
+    host: Optional[constr(strip_whitespace=True)] = None
+    isolate: Optional[constr(strip_whitespace=True)] = None
+    locale: Optional[constr(strip_whitespace=True)] = None
+    notes: Optional[constr(strip_whitespace=True)] = None
+    labels: Optional[list] = None
+    subtractions: Optional[list] = None
```
igboyes, line 225: No default values. Remember to use `exclude_unset`. This will be compatible with the existing logic:
```py
if "labels" in data:
  non_existent_labels = await check_labels(pg, data["labels"])

  if non_existent_labels:
    return bad_labels_response(non_existent_labels)
```
Data will just be `dict` you get from calling `EditSampleSchema.dict(exclude_unset=True)` on the request data object.

virtool/samples/oas.py
```diff
@@ -0,0 +1,469 @@
...
+
+    class Config:
+        schema_extra = {
+            "example": {
+                "name": "foo",
+                "host": "virus",
+                "labels": [1, 5, 6],
```
igboyes, line 232:
```suggestion
                "name": "Tobacco mosaic virus",
                "host": "Tobacco",
                "labels": [1, 5, 6],
```

virtool/samples/oas.py
```diff
@@ -0,0 +1,469 @@
...
+        }
+
+
+class CreateAnalysisJobSchema(BaseModel):
+    ref_id: str
+    subtractions: Optional[list] = None
+    workflow: QuickAnalyzeWorkflow
+
+
+class CreateAnalysisJobResponse(AnalysisMinimal):
```
igboyes, line 453:
```suggestion
class CreateAnalysisResponse(AnalysisMinimal):
```

virtool/samples/oas.py
```diff
@@ -0,0 +1,469 @@
...
+                        "handle": "mrott",
+                        "id": "ihvze2u9",
+                    },
+                    "workflow": "pathoscope_bowtie",
+                }
+            ]
+        }
+
+
+class CreateAnalysisJobSchema(BaseModel):
```
igboyes, line 447:
```suggestion
class CreateAnalysisSchema(BaseModel):
```

igboyes (changes requested):

virtool/samples/api.py
```diff
@@ -75,319 +90,343 @@
...
 
-    term = req.query.get("find")
+        if term:
+            queries.append(compose_regex_query(term, ["name", "user.id"]))
 
-    if term:
-        queries.append(compose_regex_query(term, ["name", "user.id"]))
+        if "label" in self.request.query:
+            labels = self.request.query.getall("label")
+            labels = [int(label) if label.isdigit() else label for label in labels]
```
igboyes, line 120: Label IDs are always integers, so there is no point including non-digits in the query. We can filter the list comprehension like so:
```py
labels = [int(label) for label in labels if if label.isdigit()]
```

Anything that isn't truthy for `is_digit()` will be excluded.

https://realpython.com/lessons/filtering-elements-list-comprehensions/

virtool/samples/oas.py
```diff
@@ -0,0 +1,469 @@
...
+                "user": {"administrator": True, "handle": "mrott", "id": "ihvze2u9"},
+            }
+        }
+
+
+class CreateSampleSchema(BaseModel):
+    name: constr(strip_whitespace=True, min_length=1)
+    host: constr(strip_whitespace=True) = ""
+    isolate: constr(strip_whitespace=True) = ""
+    group: str = ""
```
igboyes, line 125: Another thing I want to be consistent about is using `null` when a string field is not set and is not actually typed by the user.

Example:

* `host` is typed out by the user so it should be `""` if they haven't entered anything.
* `group` is the ID of one of the user groups on the instance (not manually typed by user) and should be `null` if unset.

I've added this to the discussion https://github.com/virtool/virtool/discussions/2644.

```suggestion
    group: Optional[str] = None
```

igboyes (changes requested):

virtool/samples/oas.py
```diff
@@ -0,0 +1,469 @@
...
+    locale: constr(strip_whitespace=True) = ""
+    library_type: LibraryType = "normal"
```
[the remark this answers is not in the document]
igboyes: @buddy326 This is not resolved in your latest push.

igboyes (approved):

PR #2636, opened 2022-08-02
chore: add data layer component for analyses

buddy326:


virtool/analyses/data.py
```diff
@@ -1,24 +1,179 @@
...
         if document["ready"]:
-            return await virtool.analyses.format.format_analysis(self._app, document)
+            document = await virtool.analyses.format.format_analysis(
+                request.app, document
+            )
+
+        headers = {
+            "Cache-Control": "no-cache",
+            "Last-Modified": isoformat(document["created_at"]),
+        }
```
igboyes, line 117: This belongs in the request handler.

virtool/analyses/data.py
```diff
@@ -1,24 +1,179 @@
...
+
+        document = await processor(self._db, document)
+
+        if document["workflow"] == "nuvs":
+            document = await apply_transforms(
+                document,
+                [AttachNuVsBLAST(self._pg)],
+            )
+
+        return [document, headers]
```
igboyes, line 127: You are going to remove this, but this will implicitly return a `tuple` and is preferred Python style in this use case:
```py
return documents, headers
```
Correct this elsewhere.

virtool/analyses/data.py
```diff
@@ -1,24 +1,179 @@
...
+
+    async def delete_analysis(self, request: Request):
```
igboyes, line 129:
```suggestion
    async def delete(self, request: Request):
```

virtool/analyses/data.py
```diff
@@ -1,24 +1,179 @@
...
+
+        try:
+            await run_in_thread(rm, path, True)
+        except FileNotFoundError:
+            pass
+
+        await recalculate_workflow_tags(self._db, sample_id)
 
-        return document
+    async def get_jobs_api(self):
```
igboyes, line 176: What is this doing?

igboyes (changes requested):
Deal with these when you resume work on this.

virtool/analyses/api.py
```diff
@@ -491,26 +336,16 @@ async def put(self) -> Union[r200[Response], r400, r403, r404, r409]:
...
-    analysis_document = await analyses.find_one({"_id": analysis_id}, ["ready"])
+    analysis_id = req.match_info["analysis_id"]
+    data = await req.json()
 
-    if not analysis_document:
+    try:
+        document = await get_data_from_req(req).analyses.finalize(
+            analysis_id, data["results"]
+        )
+    except ResourceNotFoundError:
```
igboyes, line 346: We need this data layer method to return a model, not a dictionary.

igboyes (changes requested):

virtool/analyses/data.py
```diff
@@ -1,24 +1,350 @@
...
+        if document["workflow"] == "nuvs":
+            document = await apply_transforms(
+                document,
+                [AttachNuVsBLAST(self._pg)],
+            )
+
+        if "index" not in document:
+            return document, created_at
+
+        return Analysis(**document), created_at
```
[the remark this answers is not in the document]
buddy326: So the if statement pasted below takes out the index, job, ready, reference, sample and files fields and therefore Analysis(**document) throws a bunch of pydantic validation errors hence why if index is not in the document, I just return the document. Any ideas on how to improve this?       

if document["ready"]:
            document = await virtool.analyses.format.format_analysis(
                self._config, self._db, document
            )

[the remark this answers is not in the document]
igboyes: We need `format_analysis()` to return those fields. Is that an easy fix?

Generally, if the model validation is failing, we need to fix what we are putting into the model. The whole point of this is making our API responses reliably consistent. We should not be bypassing the model to make tests pass.

buddy326:
Wasnt sure if you wanted me to change the lazy formatting string issue as so many places in virtool do it like that?

igboyes:
> Wasnt sure if you wanted me to change the lazy formatting string issue as so many places in virtool do it like that?

No. Don't bother.

tests/fixtures/data.py
```diff
@@ -15,9 +15,9 @@
 
 
 @pytest.fixture
-def data_layer(dbi, config, mocker, pg: AsyncEngine):
+def data_layer(dbi, config, mocker, pg: AsyncEngine, tasks):
     return DataLayer(
-        AnalysisData({"db": dbi, "pg": pg}),
+        mocker.Mock(spec=AnalysisData),
```
igboyes, line 20: I don't think a `Mock` should be used here. Do you have a justification?

virtool/analyses/api.py
```diff
@@ -57,52 +46,25 @@
...
             200: Successful operation
         """
-        db = self.request.app["db"]
 
-        db_query = {}
-
-        data = await paginate(
-            db.analyses,
-            db_query,
+        documents = await get_data_from_req(self.request).analyses.find(
```
igboyes, line 57:
```suggestion
        search_result = await get_data_from_req(self.request).analyses.find(
```

virtool/analyses/api.py
```diff
@@ -57,52 +46,25 @@
 
 @routes.view("/analyses")
 class AnalysesView(PydanticView):
-
-    async def get(self) -> r200[List[AnalysisMinimal]]:
+    async def get(self) -> r200[List[GetAnalysisResponse]]:
```
igboyes, line 49: Typing not correct.

virtool/analyses/api.py
```diff
@@ -328,32 +207,25 @@ async def upload(req: Request) -> Response:
...
-    analysis_file_path = (
-        req.app["config"].data_path / "analyses" / analysis_file["name_on_disk"]
-    )
-
     try:
-        size = await naive_writer(req, analysis_file_path)
-    except CancelledError:
-        logger.debug(f"Analysis file upload aborted: {upload_id}")
-        await delete_row(pg, upload_id, AnalysisFile)
+        result = await get_data_from_req(req).analyses.upload(
```
igboyes, line 211:
```suggestion
        analysis_file = await get_data_from_req(req).analyses.upload(
```

virtool/analyses/api.py
```diff
@@ -328,32 +207,25 @@ async def upload(req: Request) -> Response:
...
+            await req.multipart(), analysis_id, analysis_format, name
+        )
+    except ResourceNotFoundError:
+        raise NotFound()
 
+    if result is None:
         return Response(status=499)
 
-    analysis_file = await virtool.uploads.db.finalize(pg, size, upload_id, AnalysisFile)
+    analysis_file = result
```
igboyes, line 220: This can be removed if you are naming it properly at line 211.
```suggestion
```

virtool/analyses/data.py
```diff
@@ -1,23 +1,391 @@
...
+
+        documents = await apply_transforms(
+            documents,
+            [AttachUserTransform(self._db), AttachSubtractionTransform(self._db)],
+        )
+
+        data["documents"] = documents
 
-    async def get_by_id(self, analysis_id: str) -> Optional[Document]:
+        return AnalysisSearchResult(**data)
```
igboyes, line 94: I think this will work:
```py
return AnalysisSearchResult(**data, document=documents)
```

virtool/analyses/data.py
```diff
@@ -1,23 +1,391 @@
...
+
+    async def get(
+        self, analysis_id: str, if_modified_since: Optional[datetime]
+    ) -> Analysis:
+        """
+        Get a single analysis by its ID.
+
+        :param analysis_id: the analysis ID
+        :param if_modified_since: the date tbe document should have been last modified
```
igboyes, line 103: Spelling.

> raise ResourceNotModifiedError if the analysis has not been changed since this date

virtool/analyses/data.py
```diff
@@ -1,23 +1,391 @@
...
+        path = self._config.data_path / "samples" / sample_id / "analysis" / analysis_id
+
+        try:
+            await run_in_thread(rm, path, True)
+        except FileNotFoundError:
+            pass
+
+        await recalculate_workflow_tags(self._db, sample_id)
+
+    async def upload(
```
igboyes, line 214: This method name makes it sound like it is for uploading an analysis. It is actually for uploading a file that is a child resource of the analysis.
```suggestion
    async def upload_file(
```

virtool/analyses/data.py
```diff
@@ -1,23 +1,391 @@
...
+        """
+
+        analysis_file = await get_row_by_id(self._pg, AnalysisFile, upload_id)
+
+        if analysis_file:
+            return analysis_file.name_on_disk
+
+        raise ResourceNotFoundError()
+
+    async def download_analysis(
```
igboyes, line 271: This method is on the analysis data layer so it is not necessary to be so detailed with this name.
```suggestion
    async def download(
```

igboyes (changes requested):

virtool/analyses/data.py
buddy326: I tried it and it wasnt working so just left it as is.

PR #2638, opened 2022-08-03
feat: return users in group response

ryanfang5:


tests/groups/test_api.py
```diff
@@ -12,19 +13,27 @@ async def test_find(spawn_client, all_permissions, no_permissions):
...
+                "permissions": all_permissions,
+                "users": [
+                    {"administrator": True, "handle": "igboyes", "id": "igboyes"}
+                ],
+            },
+            {
+                "_id": "limited",
+                "permissions": no_permissions,
+                "users": [{"administrator": False, "handle": "bob", "id": "test"}],
+            },
```
igboyes, line 28: * The group document does not contain a `users` field. It is populated from the users collection in MongoDB.
* Don't need to worry about `users` at the find endpoint as the field is not included in the minimal model.

virtool/groups/data.py
```diff
@@ -49,10 +50,17 @@ async def create(self, group_id: str) -> VTGroup:
         :return: the group
         """
         try:
+            user_list = []
+            async for user in self._db.users.find():
+                if group_id in user["groups"]:
+                    user_minimal = get_user_minimal(user)
+                    user_list.append(user_minimal)
```
igboyes, line 57: Use `$in` MongoDB operator.

https://www.mongodb.com/docs/manual/reference/operator/query/in/

virtool/groups/data.py
```diff
@@ -49,10 +50,17 @@ async def create(self, group_id: str) -> VTGroup:
...
+
             document = await self._db.groups.insert_one(
                 {
                     "_id": group_id,
                     "permissions": generate_base_permissions(),
+                    "users": user_list,
```
igboyes, line 63: Don't include `users` in group document. The response field is derived from the users collection.

igboyes (changes requested):
Please also remove `VTGroup` in this PR and just use the core models.

ryanfang5:
Also updated user test snapshots to include "name" in GroupMinimal.

virtool/groups/data.py
```diff
@@ -77,6 +81,11 @@ async def update(
         if not permissions:
             raise ResourceNotFoundError
 
+        user_list = [
```
igboyes, line 84: I think this would work better after the database update.

The first reason that comes to mind is that this DB call will be wasted if the update fails for some reason. The only reason you needs to get `users_list` is if you are going to build a response.

virtool/groups/data.py
```diff
@@ -36,12 +37,17 @@ async def get(self, group_id: str) -> VTGroup:
         """
         document = await self._db.groups.find_one(group_id)
 
+        user_list = [
+            UserNested(**base_processor(user))
+            async for user in self._db.users.find({"groups": {"$in": [group_id]}})
```
igboyes, line 42: This should work just as well.
```suggestion
            async for user in self._db.users.find({"groups": group_id})
```
You don't actually have to use the `$in` operator. Sorry for the misdirection.

---

What you do here is identical to the code in `update`.

Extract to a function in `virtool.groups.db` and call it `fetch_group_users()`. It will just need to take a `db`  and `group_id` and return the list of nested users.

igboyes (changes requested):

igboyes (approved):

PR #2640, opened 2022-08-08
test: rerun transaction error failures

ryanfang5:


igboyes (approved):

igboyes (changes requested):
This should be installed as a dev dependency.

https://python-poetry.org/docs/cli/#add

igboyes (approved):

PR #2645, opened 2022-08-10
chore: review account tests and update models

ryanfang5:


ryanfang5:
Notes:
1. Should the email field in the Account model be optional? The field is currently optional in the request body, so the email validation was adjusted slightly.
2. Tests "missing_all" and "missing_email" are not technically errors but I placed them under there, should the names be changed?

igboyes:
> Notes:
> 
> 1. Should the email field in the Account model be optional? The field is currently optional in the request body, so the email validation was adjusted slightly.
> 2. Tests "missing_all" and "missing_email" are not technically errors but I placed them under there, should the names be changed?

1. Yup. https://linear.app/virtool/issue/VIR-1483/make-account-email-option-in-core-model
2. Yes. That should probably be corrected.
    I think the best course might be to parameterize `body` and `status`. Something like:
    ```py
   @pytest.mark.parametrize(
      "body,status",
      [
          ({"email": "virtool@example.com", 200})
          ({"email": "invalid_email@", 400})
          ({"password": "short", "old_password": "valid_old_password"}, 400})
          ...
      ],
    )
    ```
    You can replace `resp_is` with snapshots:
    ```py
    assert resp.status == status
    assert await resp.json() == snapshot(name="response")
    ```
    That will cover both success and error responses.

    You can add ids for each parametrized invocation using the `ids` parameter of `parametrize` to retain the nice test naming we get from the current implementation using `error`.

    https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.Metafunc.parametrize

    For example, `ids` for above would be `ids=["good_email", "invalid_email", "short_password", ...]`

    

Sound good?

virtool/account/api.py
```diff
@@ -195,8 +201,15 @@ async def post(
 
         user_id = self.request["client"].user_id
 
+        data_dict = data.dict(exclude_unset=True)
+
+        permissions_dict = {}
+
+        if "permissions" in data_dict:
+            permissions_dict = data.permissions.dict(exclude_unset=True)
```
igboyes, line 209: Can't this just be:
```suggestion
        permissions_dict = data.permissions.dict(exclude_unset=True)
```
The `permissions_dict` variable could also be dropped if the above works.

igboyes (changes requested):

virtool/account/api.py
ryanfang5: If permissions is not provided when creating a key I believe this would result in an error.

igboyes: I have just been working on another issue with this problem.

Calling `EditAccountSchema.dict(exclude_unset=True)` will actually recurse down into any nested models. This means `data.permissions` is turned into a `dict` and permissions not explicity provided in the request are excluded.

igboyes: https://github.com/virtool/virtool/blob/52d9b08645a7cc024c49144e17475fdb0b29368b/virtool/groups/data.py#L72

ryanfang5: Would this not result in an error if permissions itself is not provided in the request though? For example, in TestCreateAPIKey the body could be as below:

        body = {"name": "Foobar"}

        resp = await client.post("/account/keys", body)

Then data.permissions.dict gives an error. Maybe a try: except block would be better?

Also, if data["permissions"] is being used in your example, should we use data_dict["password"] after all instead of data.password to be consistent?

igboyes: It's not necessary to call `data.permissions.dict()` calling `data.dict(exclude_unset=True)` converts `permissions` to a `dict` if it exists. Everything gets converted to a `dict`.

ryanfang5:
data.dict(exclude_none=True) includes default values but excludes none values, which is preferred for POST.

Should I use data.dict for @routes.view("/account/reset")? Since no fields are optional in ResetPasswordSchema, I think the functionality is the same either way.

igboyes:
> data.dict(exclude_none=True) includes default values but excludes none values, which is preferred for POST.
> 
> Should I use data.dict for @routes.view("/account/reset")? Since no fields are optional in ResetPasswordSchema, I think the functionality is the same either way.

1. In some cases you want `None` to be the default value for a field in a new record created by `POST`. Is `exclude_none` going to mess that up?

2. Sounds good to me.

igboyes (approved):
I think we're good. Thanks for perfecting this!

PR #2649, opened 2022-08-11
feat: improve jobs find endpoint

ryanfang5:
BREAKING CHANGE: use new search result and restore pagination


ryanfang5:
I think the old functionality is restored while using aggregate to find the total_count and found_count. However, I'm not sure how to fit this into a pagination helper function.

virtool/jobs/data.py
```diff
@@ -101,33 +76,97 @@ async def _find_beta(self, query: MultiDictProxy) -> Document:
...
+                                        compose_regex_query(term, ["user.id"])
+                                        if term
+                                        else {}
+                                    ),
+                                    **(
+                                        {"archived": archived}
+                                        if archived is not None
+                                        else {}
+                                    ),
+                                }
```
igboyes, line 144: Hmm. Is there no way to avoid doing this twice?

If not, you can at least define the query prior to the aggregation pipeline and pass the query variable as the value for `$match` in each facet.

virtool/jobs/data.py
```diff
@@ -136,29 +175,43 @@ async def _find_beta(self, query: MultiDictProxy) -> Document:
...
+                    }
+                },
+            ],
+        ):
+            data = paginate_dict["data"]
+            documents = [base_processor(document) for document in data]
+            try:
+                found_count = paginate_dict["found_count"]
+            except KeyError:
+                found_count = 0
```
igboyes, line 194: Is a `KeyError` possible here? I'd expect that the key is always defined.

Otherwise, cleaner to just use `paginate_dict.get("found_count", 0)`.

Same for `total_count`.

igboyes (changes requested):
Looks good. A few questions and changes before we merge.

igboyes:
> I think the old functionality is restored while using aggregate to find the total_count and found_count. However, I'm not sure how to fit this into a pagination helper function.

No necessary. It is the only endpoint using aggregation at this point.

virtool/jobs/data.py
ryanfang5: I checked for a while but the only thing I could find one would be two separate aggregation calls. The first just for the total_count, and the second can use the match to both facets. I'm not sure if this would be better though.

virtool/jobs/data.py
ryanfang5: The keys weren't defined whenever no data matched the query.

igboyes (approved):

virtool/jobs/data.py
igboyes: OK. Thanks for checking.

PR #2655, opened 2022-08-15
Prevent incorrect usage of user_handle during api key authentication

ReeceHoffmann:
One line fix that prevents issues caused by the `user_id` field of `req["client"]` being populated by the users `handle` rather than `id`


igboyes (changes requested):
Update naming of `user_id` function parameter. Should be `handle`.

igboyes (approved):

PR #2659, opened 2022-08-20
chore: stop storing settings in memory

ryanfang5:


virtool/settings/data.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+        List all settings.
+
+        :return: settings object
+
+        """
+
+        settings = Settings(**await get(self._db))
+
+        return settings
```
igboyes, line 23:
```suggestion
        return Settings(**await get(self._db))
```

virtool/settings/data.py
```diff
@@ -0,0 +1,41 @@
...
+
+        :param data: updates to the current settings
+        :return: the settings
+        """
+
+        data = data.dict(exclude_unset=True)
+
+        settings = await update(self._db, data)
+
+        settings.pop("software_channel", None)
```
igboyes, line 37: Not necessary. This not be included in `Settings` because it is not defined in the model.

virtool/settings/data.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+        Update the settings.
+
+        :param data: updates to the current settings
+        :return: the settings
+        """
+
+        data = data.dict(exclude_unset=True)
+
+        settings = await update(self._db, data)
```
igboyes, line 35:
```suggestion
        settings = await update(self._db, data.dict(exclude_unset=True))
```

virtool/settings/data.py
```diff
@@ -0,0 +1,41 @@
...
+
+        settings.pop("software_channel", None)
+
+        settings = Settings(**settings)
+
+        return settings
```
igboyes, line 41:
```suggestion
        return Settings(**settings)
```

tests/account/test_api.py
```diff
@@ -62,7 +63,9 @@ async def test_get(snapshot, spawn_client, static_time):
 async def test_edit(body, status, snapshot, spawn_client, resp_is, static_time):
     client = await spawn_client(authorize=True)
 
-    client.app["settings"].minimum_password_length = 8
+    await client.db.settings.find_one_and_update(
+        {"_id": "settings"}, {"$set": {"minimum_password_length": 8}}
+    )
```
igboyes, line 68: Should use the data layer method here:
```py
await get_data_from_app(client.app).settings.update(UpdateSettingsSchema(minimum_password_length=8))
```
Make this change everywhere else in this PR as necessary.

tests/fixtures/client.py
```diff
@@ -144,7 +145,9 @@ async def func(
             app, auth=auth, cookies=cookies, auto_decompress=False
         )
 
-        return VirtoolTestClient(test_client)
+        settings = await get_data_from_app(app).settings.get_all()
+
+        return VirtoolTestClient(test_client, settings)
```
igboyes, line 150:
```suggestion
        return VirtoolTestClient(test_client, await get_data_from_app(app).settings.get_all())
```

tests/fixtures/client.py
```diff
@@ -198,7 +201,7 @@ async def _spawn_job_client(
 
         client = await aiohttp_client(app, auth=auth, auto_decompress=False)
         client.db = dbi
-        client.settings = app["settings"]
+        client.settings = await get_data_from_app(app).settings.get_all()
```
igboyes, line 204: Remove this. We should be using `get_all()`.

tests/samples/test_api.py
```diff
@@ -746,11 +761,11 @@ async def test_job_remove(
 
     resp = await client.delete("/samples/test")
 
+    settings = client.settings
```
igboyes, line 764:
```suggestion
```

tests/samples/test_api.py
```diff
@@ -746,11 +761,11 @@ async def test_job_remove(
...
+
     if exists and not ready:
         await resp_is.no_content(resp)
-        assert not await check_name(
-            client.app["db"], client.app["settings"], "test", "test"
-        )
+        assert not await check_name(client.app["db"], settings, "test", "test")
```
igboyes, line 768:
```suggestion
        assert not await check_name(client.app["db"], get_data_layer_from_app(client.app).settings, "test", "test")
```

virtool/settings/data.py
```diff
@@ -0,0 +1,41 @@
...
+
+class SettingsData:
+    def __init__(self, db):
+        self._db: DB = db
+
+    async def get_all(self) -> Settings:
+        """
+        List all settings.
+
+        :return: settings object
```
igboyes, line 17:
```suggestion
        :return: the application settings
```

virtool/settings/data.py
```diff
@@ -0,0 +1,41 @@
...
+        settings = Settings(**await get(self._db))
+
+        return settings
+
+    async def update(self, data: UpdateSettingsSchema) -> Settings:
+        """
+        Update the settings.
+
+        :param data: updates to the current settings
+        :return: the settings
```
igboyes, line 30:
```suggestion
        :return: the application settings
```

virtool/startup.py
```diff
@@ -301,9 +304,10 @@ async def startup_routes(app: Application):
 
 
 async def startup_sentry(app: typing.Union[dict, Application]):
+    settings = Settings(**await get(app["db"]))
     if (
         not app["config"].no_sentry
-        and app["settings"].enable_sentry is not False
+        and settings.enable_sentry is not False
         and not app["config"].dev
```
igboyes, line 311: Can we not use `get_data_from_app(app).settings` here?

virtool/startup.py
```diff
@@ -315,14 +319,14 @@ async def startup_sentry(app: typing.Union[dict, Application]):
...
-    Draws settings from the settings database collection and populates `app["settings"`.
+    Draws settings from the settings database collection.
 
     Performs migration of old settings style to `v3.3.0` if necessary.
 
     :param app: the app object
 
     """
-    app["settings"] = await ensure(app["db"])
+    await ensure(app["db"])
```
igboyes, line 329: Should be a data layer method.

virtool/users/checks.py
```diff
@@ -1,13 +1,16 @@
...
     Return error text if the password in the request does not meet
     application length requirements.
 
     """
-    minimum_password_length = req.app["settings"].minimum_password_length
+    settings = await get_data_from_req(req).settings.get_all()
+    minimum_password_length = settings.minimum_password_length
 
     if len(password) < minimum_password_length:
         return f"Password does not meet minimum length requirement ({minimum_password_length})"
```
igboyes, line 16:
```suggestion
    if len(password) < settings.minimum_password_length:
        return f"Password does not meet minimum length requirement ({settings.minimum_password_length})"
```

virtool/settings/data.py
```diff
@@ -0,0 +1,41 @@
...
+
+    async def get_all(self) -> Settings:
+        """
+        List all settings.
+
+        :return: settings object
+
+        """
+
+        settings = Settings(**await get(self._db))
```
igboyes, line 21: Merge the `get()` function into this method. It shouldn't be necessary anymore. The data layer should always be used to access settings.

igboyes (changes requested):

virtool/fake/factory.py
```diff
@@ -232,6 +233,10 @@ async def load_test_case_from_yml(
 
     settings = Settings(**await get(app["db"]))
 
+    # settings2 = await get_data_from_app(app).settings.get_all()
+
+    # Not sure why settings2 is None when it should be the same as settings?
+
```
ryanfang5, line 239: I was trying to use get_all instead of get here, but I'm not sure why settings2 is None and not the same as settings as I would expect.

ryanfang5:
Should update and ensure both be merged into the data layer or is it only get?

igboyes:
> Should update and ensure both be merged into the data layer...

Yes. The bottom line with the data layer project is to consolidate data access there.

virtool/fake/factory.py
igboyes: Is this still a problem? I don't quite understand.

tests/fixtures/client.py
```diff
@@ -8,16 +8,17 @@
...
 
 class VirtoolTestClient:
-    def __init__(self, test_client):
+    def __init__(self, test_client, settings):
         self._test_client = test_client
 
         self.server = self._test_client.server
         self.app = self.server.app
-        self.settings = self.app["settings"]
+        self.settings = settings
```
igboyes, line 21: Is this necessary anymore?

tests/samples/test_api.py
```diff
@@ -309,10 +310,15 @@ async def test(
...
-        client.app["settings"].sample_group_write = True
+        settings_data = {
+            "sample_group": group_setting,
+            "sample_all_write": True,
+            "sample_group_write": True,
+        }
+
+        await get_data_from_app(client.app).settings.update(
+            UpdateSettingsSchema(**settings_data)
+        )
```
igboyes, line 321: No need for extra `dict` here.

tests/samples/__snapshots__/test_api.ambr
```diff
@@ -4,16 +4,16 @@
 # name: TestCreate.test[uvloop-force_choice].1
```
igboyes, line 4: Why are there changes in this snapshot file? I don't see why there should be when you are just changing how settings are retrieved.

virtool/fake/factory.py
ryanfang5: Fixed, the issue was that {} was being returned when no settings were found. Now get_all is returning Settings() instead.

virtool/fake/factory.py
```diff
@@ -59,14 +60,14 @@ class WorkflowTestCase:
 class TestCaseDataFactory:
     """Initialize the database with fake data for a test case."""
 
-    def __init__(self, app: App, user_id: str, job_id: str = None):
+    def __init__(self, settings: Settings, app: App, user_id: str, job_id: str = None):
```
igboyes, line 63: I don't think you need to pass in `settings` separately here.

Can't you just use `get_data_from_app` on the `app` object that's also a parameter here?

virtool/hmm/db.py
```diff
@@ -60,7 +61,7 @@ async def delete_unreferenced_hmms(db, config: Config) -> pymongo.results.Delete
 
     delete_result = await db.hmm.delete_many({"_id": {"$nin": referenced_ids}})
 
-    logger.debug(f"Deleted {delete_result.deleted_count} unreferenced HMMs")
+    logger.debug("Deleted %d unreferenced HMMs", delete_result.deleted_count)
```
igboyes, line 64: Can you explain this? Are we gaining something with this change or just satisfying Deepsource?

virtool/samples/data.py
```diff
@@ -30,18 +31,18 @@
...
 class SamplesData(DataLayerPiece):
-    def __init__(self, config: Config, db, pg: AsyncEngine):
-        self._config = config
-        self._db = db
-        self._pg = pg
+    def __init__(self, app):
+        self._app = app
+        self._config: Config = app["config"]
+        self._db = app["db"]
+        self._pg: AsyncEngine = app["pg"]
```
igboyes, line 45: The settings should be available at `self.data.settings`.

Explanation:
* https://github.com/virtool/virtool/blob/main/virtool/data/piece.py
* https://github.com/virtool/virtool/blob/main/virtool/data/layer.py

virtool/settings/data.py
```diff
@@ -0,0 +1,57 @@
...
+        :return: the application settings
+
+        """
+
+        settings = await self._db.settings.find_one(
+            {"_id": "settings"}, projection=PROJECTION
+        )
+
+        if settings:
+            return Settings(**settings)
```
igboyes, line 26: Hmm. I would think we should raise an exception if the settings are not defined. It should never happen.

virtool/settings/data.py
```diff
@@ -0,0 +1,57 @@
...
+            {"_id": "settings"}, {"$set": data.dict(exclude_unset=True)}
+        )
+
+        return Settings(**updated)
+
+    async def ensure(self) -> Settings:
+        """
+        Ensure the settings document is updated and filled with default values.
+
+        :return: a dictionary with settings data
```
igboyes, line 48: Not a dictionary.

virtool/settings/data.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+
+        existing = await self._db.settings.find_one({"_id": "settings"}, {"_id": False}) or {}
```
igboyes, line 51:
```suggestion
        existing = await self._db.settings.find_one({"_id": "settings"}, PROJECTION) or {}
```

igboyes (changes requested):
A few more fixes and questions.

Looking good!

tests/fixtures/client.py
ryanfang5: You are right, for some reason I thought I needed to maintain the self.settings feature by passing settings in separately, but it's completely unneeded as anything that needs the settings would just call the get_all command.

virtool/fake/factory.py
ryanfang5: Same realisation as in fixtures/client.py

tests/samples/__snapshots__/test_api.ambr
ryanfang5: I think these snapshots are actually correct. When the samples data layer was added, the get_settings method in db was used to get the settings since my data layer changes weren't added yet. However, sometimes client.app["settings"] was used to update the settings, which I believe would not be read in get_settings since that checks from the db.

virtool/hmm/db.py
ryanfang5: This was just to satisfy Deepsource.

virtool/settings/data.py
ryanfang5: I believe this only occurs in test_yaml_loading.py because the app fixture does not have a data layer and ensure is not being called. I could add them directly to the fixture or I could do this as a workaround: 

async def test_load_yml(app, data_layer, fake, example_test_case):

    app["data"] = data_layer

    await get_data_from_app(app).settings.ensure()

then it would work as well.

igboyes (changes requested):
Close! Address comment and revert all string interpolation to use f-strings.

virtool/settings/db.py
```diff
@@ -1,73 +1,5 @@
-import logging
-from typing import Any, Dict, Optional
-
-from motor.motor_asyncio import AsyncIOMotorClientSession
 from virtool_core.models.settings import Settings
 
-logger = logging.getLogger(__name__)
+__all__ = ["Settings"]
```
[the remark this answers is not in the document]
ryanfang5: I wanted to keep the db file for any future functions and this was just to satisfy DeepSource.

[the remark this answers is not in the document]
igboyes: The file can be added in a future commit if it is need again. Please remove it.

igboyes (dismissed):

PR #2662, opened 2022-08-24
Send custom code on websocket authentication failure

ReeceHoffmann:
Changes:
  - Adds new authentication policy that allows any client through but still collects authentication information
  - If authentication fails the server will immediately close the connection with a custom closure code.

### DIFF
--- a/virtool/dispatcher/connection.py
+++ b/virtool/dispatcher/connection.py
@@ -28,9 +28,9 @@ class Connection:
             if "Cannot write to closing transport" not in str(err):
                 raise
 
-            await self.close()
+            await self.close(1002)
 
-    async def close(self, code: Union[int, str] = None):
+    async def close(self, code: int):
         """
         Closes the underlying websocket connection.
         :param code: closure code to send to the client
--- a/virtool/dispatcher/connection.py
+++ b/virtool/dispatcher/connection.py
@@ -30,8 +30,9 @@ class Connection:
 
             await self.close()
 
-    async def close(self):
+    async def close(self, code: Union[int, str] = None):
         """
         Closes the underlying websocket connection.
+        :param code: closure code to send to the client
         """
-        await self._ws.close()
+        await self._ws.close(code=code)
--- a/virtool/http/policy.py
+++ b/virtool/http/policy.py
@@ -68,6 +68,12 @@ class PublicRoutePolicy(DefaultRoutePolicy):
     allow_unauthenticated = True
 
 
+class WebSocketRoutePolicy(DefaultRoutePolicy):
+    """Any client can access the route."""
+
+    allow_unauthenticated = True
+
+
 def policy(route_policy: Union[DefaultRoutePolicy, Type[DefaultRoutePolicy]]):
     def decorator(func):
         try:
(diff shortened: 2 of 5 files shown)


igboyes:
Can you give a little info about the custom code in your PR comment. I'm guessing it is a websocket feature.

ReeceHoffmann:
> Can you give a little info about the custom code in your PR comment. I'm guessing it is a websocket feature.

I assume your referencing the closure code? When websocket sends the closing message (if it sends one) there is an option to include an `int` closure code analogous to the status codes of http response. There are 15 defined closure codes but none of those seemed to fit overly well, though I suppose an argument could be made for 1008 or 1011, but both are fairly general. In the websocket spec, codes from 4000-4999 are set aside for private use to be determined by application sending the codes. I went with the first (4000) but arguably 4401 would be more memorable. Anyway link to the standard in case you want to delve deeper: https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1

igboyes:
This is a cleaner solution than I hoped for!

virtool/http/policy.py
```diff
@@ -68,6 +68,12 @@ class PublicRoutePolicy(DefaultRoutePolicy):
     allow_unauthenticated = True
 
 
+class WebSocketRoutePolicy(DefaultRoutePolicy):
+    """Any client can access the route."""
+
+    allow_unauthenticated = True
```
igboyes, line 74: Is this not just the same as `PublicRoutePolicy`?

virtool/dispatcher/connection.py
```diff
@@ -30,8 +30,9 @@ async def send(self, message: Union[dict, list]):
 
             await self.close()
 
-    async def close(self):
+    async def close(self, code: Union[int, str] = None):
```
igboyes, line 33: 1. Should we just required `int` codes? I don't see the point of allowing both types here.
2. Should this be a keyword argument? Are sometimes not passing `code` to `Connection.close()`?

virtool/http/policy.py
ReeceHoffmann: Yes mostly, but with one annoying distinction. The auth functions checks for the `PublicRoutePolicy` and if it finds it, the function does not bother authenticating the client and just passes default values. The result of that is that I can't tell whether the user is logged in in websocket's `root` as the client object will always have default values. 

That said I could probably clean this up a little by inheriting from `publicRoutePolicy` and will do that with the next push if it works as expected

virtool/dispatcher/connection.py
ReeceHoffmann: 1. Yeah seems fair, limited to `int` in next push
2. I gave a default value since we aren't always passing a code, although perhaps we should be. The other instance where the server will proactively close the connection can probably be covered by the code for the server going down (1001).  I will add the code and remove the default value.

virtool/http/policy.py
ReeceHoffmann: Turns out if I make it a subclass then `isInstance` still considers `websocketRoutePolicy` an instance of `publicRoutePolicy` left mostly as is but with a more suitable docstring.

ReeceHoffmann:
Force pushed to change commit message for ESLint

igboyes (approved):

PR #2663, opened 2022-08-24
chore: return job objects from data layer

ryanfang5:


virtool/jobs/data.py
```diff
@@ -170,22 +172,26 @@ async def find(self, query: MultiDictProxy) -> Document:
...
+        return JobSearchResult(
+            **{
+                "counts": await self._get_counts(),
+                "documents": [JobMinimal(**document) for document in documents],
+                "total_count": total_count,
+                "found_count": found_count,
+                "page_count": page_count,
+                "per_page": per_page,
+                "page": page,
+            }
```
igboyes, line 193: The extra `dict` is unnecessary here.
```py
JobSearchResult(
    counts=await self.get_counts(),
    ...
)
```

virtool/jobs/data.py
```diff
@@ -226,7 +232,7 @@ async def create(
         await self._db.jobs.insert_one(document)
         await self._client.enqueue(workflow, document["_id"])
 
-        return document
+        return Job(**await fetch_complete_job(self._db, document))
```
igboyes, line 235: Return `Job` directly from `fetch_complete_job()`.

virtool/fake/factory.py
```diff
@@ -228,6 +228,9 @@ async def load_test_case_from_yml(
         "workflow",
     )(yml)
 
+    if workflow is None:
+        workflow = "build_index"
+
```
igboyes, line 233: Why?

igboyes (changes requested):

virtool/fake/factory.py
ryanfang5: test_yaml_loading.py was failing because it was creating a job with workflow as None, so I gave it a default value here.

igboyes: The root of the problem is that the test YAML input file has an invalid value for `workflow`. Update this instead.

Be careful about changes like this that are made only to pass tests. Many little patches like this eventually add up to a lot of murky behaviour in a codebase.

ryanfang5: I see, thank you.

virtool/jobs/api.py
```diff
@@ -262,4 +262,4 @@ async def push_status(req):
     except ResourceConflictError:
         raise HTTPConflict(text="Job is finished")
 
-    return json_response(document["status"][-1], status=201)
+    return json_response(document, status=201)
```
igboyes, line 265: You had this right before. You are creating a new status resource at this endpoint and should return it as the response, not the parent job.

virtool/jobs/db.py
```diff
@@ -61,3 +63,14 @@ async def processor(db, document: dict) -> dict:
         ),
         [AttachUserTransform(db)],
     )
+
+
+async def fetch_complete_job(db, document, key=None) -> Job:
+    document = await processor(db, document)
+    document["user"] = UserNested(**document["user"])
+    document["status"] = [JobStatus(**status) for status in document["status"]]
```
igboyes, line 71: Is this necessary? I would think that Pydantic would deal with making the objects for you.

igboyes (changes requested):

virtool/jobs/api.py
ryanfang5: I created and returned the status in the data layer, which is why I just return document here. Is this incorrect?

igboyes: Sorry, that part was hidden in the review tool. You're right.

igboyes: You're right!

virtool/jobs/db.py
ryanfang5: You are right, I think I was missing parameters earlier which caused it not to work.

igboyes (approved):

PR #2671, opened 2022-08-29
fix: correctly import hex color validator

ryanfang5:


tests/samples/__snapshots__/test_api.ambr
```diff
@@ -31,9 +31,9 @@
     'locale': '',
     'name': 'Foobar',
     'notes': '',
-    'nuvs': False,
+    'nuvs': 'False',
     'paired': False,
-    'pathoscope': False,
+    'pathoscope': 'False',
```
igboyes, line 36: These can't be coerced to strings.

Can you make sure that bool and str don't get coerced to each other? I think I must be using `Union` incorrectly in Pydantic.

https://github.com/virtool/virtool-core/blob/179ec43cd055c37fa687f86bb23497b0b2e3cf03/virtool_core/models/samples.py#L25

igboyes (changes requested):

PR #2673, opened 2022-08-31
chore: add uploads to data layer

ryanfang5:


virtool/uploads/api.py
```diff
@@ -37,22 +38,17 @@ async def get(self) -> r200[List[GetUploadsResponse]]:
...
-            {
-                "documents": await apply_transforms(
-                    uploads, [AttachUserTransform(self.request.app["db"])]
-                )
-            }
+        uploads = await get_data_from_req(self.request).uploads.find(
+            user, upload_type, ready
         )
 
+        return json_response({"documents": uploads})
```
igboyes, line 50: The response doesn't match the method return value typing.

Create an `UploadSearchResult` model in core and import here. Return an `UploadSearchResult` object directly from the data layer method.

virtool/uploads/api.py
```diff
@@ -130,23 +125,18 @@ async def get(self) -> Union[r200[FileResponse], r404]:
...
         """
-        pg = self.request.app["pg"]
 
         upload_id = int(self.request.match_info["id"])
 
-        upload = await virtool.uploads.db.get(pg, upload_id)
+        upload = await get_data_from_req(self.request).uploads.get(upload_id)
 
         if not upload:
             raise NotFound()
```
igboyes, line 134: Instead of this, follow the `ResourceNotFoundError` pattern used in other data layer pieces.

You should only have one call to the data layer that looks like this:
```py
upload_path = await get_data_from_req(self.request).uploads.get_upload_path(
    upload_id
)
```

virtool/uploads/api.py
```diff
@@ -168,10 +158,9 @@ async def delete(self) -> Union[r204, r401, r403, r404]:
...
             404: Not found
         """
-        pg = self.request.app["pg"]
         upload_id = int(self.request.match_info["id"])
 
-        upload = await virtool.uploads.db.delete(self.request, pg, upload_id)
+        upload = await get_data_from_req(self.request).uploads.delete(upload_id)
 
         if not upload:
             raise NotFound()
```
igboyes, line 166: Use `ResourceNotFoundError` pattern instead. See already implemented pieces of the data layer.

Do this for all other applicable situations in this module.

virtool/uploads/data.py
```diff
@@ -0,0 +1,199 @@
...
+            upload.uploaded_at = virtool.utils.timestamp()
+            upload.ready = True
+
+            upload = upload.to_dict()
+
+            await session.commit()
+
+        return upload
+
+    async def get_upload_path(self, name_on_disk: str) -> str:
```
igboyes, line 189: Return type is incorrect.

virtool/uploads/data.py
```diff
@@ -0,0 +1,199 @@
...
+        try:
+            await run_in_thread(
+                rm, self._config.data_path / "files" / upload.name_on_disk
+            )
+        except FileNotFoundError:
+            pass
+
+        return upload
+
+    async def delete_row(self, upload_id: int) -> Optional[Upload]:
```
igboyes, line 134: I don't know why this was a separate function, just merge it into `delete`.

virtool/uploads/data.py
```diff
@@ -0,0 +1,199 @@
...
+            ).scalar()
+
+            if not upload:
+                return None
+
+        return upload
+
+    async def delete(self, upload_id: int) -> Upload:
+        """
+        "Delete" a row in the `uploads` table and remove it from the local disk.
```
igboyes, line 114: The data layer is meant to abstract away the underlying services like MongoDB, SQL, Redis, storage media, and more. This means the layer looks them same even if implementation is changing behind the scenes.

In this light, the docstring doesn't make sense. Come up with a docstring that says what the method does without talking about the underlying implementation.

virtool/uploads/data.py
```diff
@@ -0,0 +1,199 @@
...
+
+            upload.size = size
+            upload.uploaded_at = virtool.utils.timestamp()
+            upload.ready = True
+
+            upload = upload.to_dict()
+
+            await session.commit()
+
+        return upload
```
igboyes, line 187: Return the core `Upload` object, not a `dict`.

virtool/uploads/data.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        async with AsyncSession(self._pg) as session:
+            upload = (
+                await session.execute(
+                    select(Upload).filter_by(id=upload_id, removed=False)
+                )
+            ).scalar()
+
+            if not upload:
+                return None
```
igboyes, line 108: Raise `ResourceNotFound` instead of returning none.

virtool/uploads/data.py
```diff
@@ -0,0 +1,199 @@
...
+
+        return upload
```
igboyes, line 110: Return a core `Upload` object. This is a SQLAlchemy object.

virtool/uploads/db.py
```diff
@@ -1,196 +1,16 @@
...
-        upload.removed_at = virtool.utils.timestamp()
-
-        upload = upload.to_dict()
-
-        await session.commit()
-
-    return upload
-
-
 async def release(pg: AsyncEngine, upload_ids: Union[int, List[int]]):
```
igboyes, line 14: Why wasn't this moved into the data layer?

virtool/uploads/data.py
```diff
@@ -0,0 +1,199 @@
...
+            upload.removed = True
+            upload.removed_at = virtool.utils.timestamp()
+
+            upload = upload.to_dict()
+
+            await session.commit()
+
+        return Upload(**upload)
+
+    async def finalize(
```
igboyes, line 160: This does not abstract the backing services. You have a `model` parameter that is a SQLAlchemy model class.

This method should stay in `db.py`. This 'uploads' piece of the data layer is for user uploads. However, there are other uploads that come from workflows. Both have a finalization step for the SQL row when the upload succeeds.

1. Restore the `virtool.uploads.db.finalize` function.
2. The finalize method of the data layer should just call this function and pass it `Upload` (SQL) for the model argument.

igboyes (changes requested):

virtool/uploads/api.py
ryanfang5: I made get_upload_path as a utility function since I saw the code was being used twice in the jobs_api as well as the view, whereas get(upload_id) is used to get the actual upload. Should I move get_upload_path to utils?

virtool/uploads/utils.py
```diff
@@ -76,3 +79,16 @@ async def naive_writer(
...
+
+async def get_upload_path(config: Config, name_on_disk: str) -> str:
+    """
+    Get the local upload path and return it.
+    """
+    upload_path = config.data_path / "files" / name_on_disk
+
+    # check if the file has been manually removed by the user
+    if not upload_path.exists():
+        raise ResourceNotFoundError("Uploaded file not found at expected location")
```
igboyes, line 92:
```suggestion
    if not await run_in_thread(upload_path.exists):
        raise ResourceNotFoundError("Uploaded file not found at expected location")
```

virtool/uploads/data.py
```diff
@@ -0,0 +1,203 @@
...
+
+            await session.flush()
+
+            upload.name_on_disk = f"{upload.id}-{upload.name}"
+
+            upload = upload.to_dict()
+
+            await session.commit()
+
+        return Upload(**upload)
```
igboyes, line 90: You are using the same class here where you need both the SQLAlchemy and core models.

In the subtraction data layer I am importing the `SubtractionFile` SQL model like this:
```py
from virtool.subtractions.models import SubtractionFile as SQLSubtractionFile
```
This allows us to differentiate between the two models. You can do the same for uploads.

virtool/uploads/utils.py
```diff
@@ -76,3 +79,16 @@ async def naive_writer(
             size += len(chunk)
 
     return size
+
+
+async def get_upload_path(config: Config, name_on_disk: str) -> str:
```
igboyes, line 84: This returns a `pathlib.Path` object.
```suggestion
async def get_upload_path(config: Config, name_on_disk: str) -> Path:
```

igboyes (changes requested):
Be careful about SQL vs Pydantic models.

SQL models are defined in Virtool (not core) and are only used for building rows in Postgres. Core models are returned from the data layer and passed to `json_response`.

Please review your changes to make sure you are using the models correctly. I am not sure if I have caught all the problems.

pyproject.toml
```diff
@@ -53,7 +53,8 @@ msal = "^1.16.0"
 python-jose = {extras = ["cryptography"], version = "^3.3.0"}
 aiojobs = "^1.0.0"
 aiohttp-pydantic = "^1.12.1"
-virtool-core = "^2.26.2"
+virtool-core = "^2.27.1"
+setuptools = "^65.3.0"
```
igboyes, line 57: Where did this come from?

virtool/uploads/data.py
```diff
@@ -0,0 +1,204 @@
...
+
+        uploads = await apply_transforms(uploads, [AttachUserTransform(self._db)])
+
+        return UploadSearchResult(documents=uploads)
+
+    async def create(
+        self,
+        name: str,
+        upload_type: str,
+        reserved: bool = False,
```
igboyes, line 63: Just pass in the model object as a positional argument called `data` like we do elsewhere.

It is a recipe for bugs to have default values on these function parameters in addition to the defaults and validation on the models.

The `user` parameter will have to stay as is because it is not part of the request model.

igboyes (changes requested):
Looking good. Just a few more things to tweak.

pyproject.toml
ryanfang5: I was having an error only on my local machine where pkg_resources from setuptools was not found, so I was unable to run any tests. This must have been added accidentally as I was trying to solve the issue.

igboyes (changes requested):
Please address built-in name clash issue from Deepsource. Changing name to `upload_type` is a fine solution.

https://deepsource.io/gh/virtool/virtool/run/5e39fc7b-6b3c-4f26-8f8c-58172d34ef09/python/PYL-W0622

igboyes (approved):

PR #2676, opened 2022-09-01
chore: add data layer component for accounts

buddy326:


virtool/account/data.py
```diff
@@ -0,0 +1,364 @@
...
+        :param data: the fields to create a new API key
+        :param user_id: the user ID
+        :return: the API key
+        """
+        if isinstance(data.permissions, dict):
+            permissions_dict = data.permissions
+        else:
+            permissions_dict = data.permissions.dict(exclude_unset=True)
+
+        name = data.name
```
igboyes, line 166: Delete this. Just use `data.name` where you would use `name`.

virtool/account/data.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        Delete all API keys for the account associated with the requesting session.
+
+        :param user_id: the user ID
+        """
+        await self._db.keys.delete_many({"user.id": user_id})
+
+    async def get_key(
+        self, user_id: str, key_id: str, api_key_projection: dict
+    ) -> dict:
```
igboyes, line 213: Need to return a model instance, not a dict.

virtool/account/data.py
```diff
@@ -0,0 +1,364 @@
...
+        Create a new API key.
+
+        :param data: the fields to create a new API key
+        :param user_id: the user ID
+        :return: the API key
+        """
+        if isinstance(data.permissions, dict):
+            permissions_dict = data.permissions
+        else:
+            permissions_dict = data.permissions.dict(exclude_unset=True)
```
igboyes, line 164: 1. Are you sure this is necessary? I would have thought all permission dictionaries used to populate the model would be converted to objects.
2. This is repeated elsewhere. Extract it to a reusable function.

**Number 1 is the preferred change here!**

virtool/account/data.py
```diff
@@ -0,0 +1,364 @@
...
+        :param api_key_projection: the projection for the api key
+        :return: the API key
+        """
+        if data.permissions is None:
+            permissions_update = {}
+        else:
+            permissions_update = data.permissions.dict(exclude_unset=True)
+
+        if not await self._db.keys.count_documents({"id": key_id}):
+            raise ResourceNotFoundError()
```
igboyes, line 253: See `id_exists` helper function.

virtool/account/data.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        # When `remember` is set, the session will last for 1 month instead of the
+        # 1-hour default
+
+        # Re-render the login page with an error message if the username doesn't
+        # correlate to a user_id value in
+        # the database and/or password are invalid.
+        handle = data.username
+        password = data.password
+        remember = data.remember
```
igboyes, line 304: Not necessary to redeclare attributes as separate variables. This was done when using dictionaries to improve readibility.

virtool/account/oas.py
```diff
@@ -123,6 +126,13 @@ class Config:
 
 
 class CreateAPIKeyResponse(BaseModel):
+    created_at: datetime
+    groups: List[GroupMinimal]
+    id: str
+    key: str
+    name: str
+    permissions: Permissions
+
```
igboyes, line 135: Needs to be in core as `APIKey`. Only example should be here. This applies to the rest of your API key response models.

virtool/account/oas.py
```diff
@@ -262,6 +278,9 @@ class Config:
 
 
 class GetAPIKeysResponse(BaseModel):
+    id: str
+    name: str
```
igboyes, line 282: You sure that's all that is in this response?

Should be called `ListAPIKeysResponse`. If a reader misses the 's' at first glace, they may think this is a response model for getting a single API key.

virtool/account/api.py
```diff
@@ -92,39 +93,22 @@ async def patch(
             400: Invalid input
             401: Requires Authorization
         """
-        db = self.request.app["db"]
-        user_id = self.request["client"].user_id
-
-        update = {}
-
-        data = data.dict(exclude_unset=True)
+        password = data.password
```
igboyes, line 96: Not necessary to declare as new variable. Just access attribute where needed.

virtool/account/api.py
```diff
@@ -92,39 +93,22 @@ async def patch(
...
 
-        if "password" in data:
-            # Request model ensures that if one password is passed in,
-            # the other is as well.
-            if error := await check_password_length(self.request, data["password"]):
+        if password is not None:
```
igboyes, line 98: Model should not be using default values for a PATCH endpoint.

virtool/account/api.py
```diff
@@ -92,39 +93,22 @@ async def patch(
...
-        if "email" in data:
-            update["email"] = data["email"]
-
-        if update:
-            document = await db.users.find_one_and_update(
-                {"_id": user_id},
-                {"$set": update},
-                projection=virtool.account.db.PROJECTION,
+        try:
+            updated_account = await get_data_from_req(self.request).account.edit(
```
igboyes, line 105:
```suggestion
            account = await get_data_from_req(self.request).account.edit(
```

virtool/account/api.py
```diff
@@ -139,11 +123,11 @@ async def get(self) -> Union[r200[AccountSettingsResponse], r401]:
...
             401: Requires authorization
         """
-        account_settings = await get_one_field(
-            self.request.app["db"].users, "settings", self.request["client"].user_id
+        account_settings = await get_data_from_req(self.request).account.get_settings(
+            "settings", self.request["client"].user_id
         )
 
-        return json_response(account_settings)
+        return json_response(AccountSettingsResponse.parse_obj(account_settings).dict())
```
igboyes, line 130: Not necessary to call `dict()`. The `json_response` function will convert for you.
```suggestion
        return json_response(AccountSettingsResponse.parse_obj(account_settings))
```

virtool/account/api.py
```diff
@@ -280,38 +251,17 @@ async def patch(
...
-
-        key_id = self.request.match_info.get("key_id")
-
-        if not await db.keys.count_documents({"id": key_id}):
+        try:
+            key = await get_data_from_req(self.request).account.edit_key(
+                self.request["client"].user_id,
+                self.request.match_info.get("key_id"),
+                data,
+                API_KEY_PROJECTION,
```
igboyes, line 259: Why is this projection being passed in? It is a constant.

igboyes (changes requested):
Return model objects from the data layer, not dictionaries.

virtool/account/api.py
buddy326: so that means remove the Optional keyword right? And a request body containing no fields for a PATCH request should end in a status code of 400?

igboyes: No.

Remember we are using `exclude_unset` to drop fields that are not set in the request body. The fields are optional, but they do not have default values.

A `400` is automatically returned by `aiohttp-pydantic` when the request JSON is `'{}'`. That is not valid JSON—it must have at least one key.

virtool/account/data.py
buddy326: id_exists wasnt working properly (would return false for an id that DID exist) so I just used count_documents

tests/account/__snapshots__/test_api.ambr
```diff
@@ -1,10 +1,10 @@
 # name: TestCreateAPIKey.test[uvloop-False-False]
   <class 'dict'> {
+    'administrator': False,
     'created_at': '2015-10-06T20:00:00Z',
     'groups': <class 'list'> [
     ],
     'id': 'foobar_0',
-    'key': 'raw_key',
```
igboyes, line 7: The `key` is returned in the create request, but no other. In the UI it is displayed just this once.

I think the best solution is to create a response model that inherits from core `APIKey` and  adds the `key` field. This will only be returned on the create endpoint.

tests/account/test_db.py
```diff
@@ -73,22 +76,27 @@ async def test_create_api_key(administrator, has_permission, mocker, dbi, static
...
                 Permission.create_sample.value: True,
-                "create_subtraction": has_permission,
+                "modify_subtraction": has_permission,
             },
         }
     )
 
-    document = await virtool.account.db.create_api_key(
-        dbi, "Foo", {Permission.create_sample.value: True, "create_subtraction": True}, "bob"
+    account = AccountData(dbi)
```
igboyes, line 84:
```suggestion
    account_data = AccountData(dbi)
```

virtool/account/data.py
```diff
@@ -0,0 +1,327 @@
...
+        :param user_id: the user ID
+        :return: the user account
+        """
+        document = await self._db.users.find_one(user_id, PROJECTION)
+
+        document["primary_group"] = {"id": document["primary_group"]}
+
+        return Account(**document)
+
+    async def edit(self, user_id: str, data: dict) -> Account:
```
igboyes, line 57: Why is `data` a dictionary here? It should be a model.

igboyes (changes requested):

PR #2688, opened 2022-09-22
chore: add data layer piece for references

ryanfang5:


ryanfang5:
I have marked a few endpoints that I believe are currently missing tests, which may have unknown validation errors.

virtool/references/data.py
```diff
@@ -0,0 +1,658 @@
...
+                raise ResourceConflictError("User already exists")
+
+            if "does not exist" in str(err):
+                raise ResourceConflictError("User does not exist")
+
+            raise
+
+        return ReferenceUser(**await extend_user(self._mongo, subdocument))
+
+    async def edit_user(
```
igboyes, line 618:
```suggestion
    async def update_user(
```

virtool/references/data.py
```diff
@@ -0,0 +1,658 @@
...
+        subdocument = await virtool.references.db.edit_group_or_user(
+            self._mongo, ref_id, user_id, "users", data
+        )
+
+        if subdocument is None:
+            raise ResourceNotFoundError()
+
+        return ReferenceUser(**await extend_user(self._mongo, subdocument))
+
+    async def remove_user(self, ref_id: str, user_id: str, req):
```
igboyes, line 643:
```suggestion
    async def delete_user(self, ref_id: str, user_id: str, req):
```

virtool/references/data.py
```diff
@@ -0,0 +1,658 @@
...
+            user_id,
+            rights,
+            job_id=job_id,
+        )
+
+        document = await virtool.indexes.db.processor(self._mongo, document)
+
+        return IndexMinimal(**document)
+
+    async def find_groups(self, ref_id: str) -> ReferenceGroup:
```
igboyes, line 495:
```suggestion
    async def list_groups(self, ref_id: str) -> ReferenceGroup:
```

virtool/references/data.py
```diff
@@ -0,0 +1,658 @@
...
+
+        targets = data.get("targets")
+
+        if targets:
+            names = [t["name"] for t in targets]
+
+            if len(names) != len(set(names)):
+                raise ResourceConflictError(
+                    "The targets field may not contain duplicate names"
+                )
```
igboyes, line 292: It seems to me this would be done better in Pydantic.

igboyes (changes requested):
Please address the comments.

Once these have been resolved, I want to add API and data layer tests for all functions that modify the database. I will identify these in the next review.

virtool/references/utils.py
```diff
@@ -266,12 +267,13 @@ def get_otu_schema(require_id: bool) -> dict:
     }
 
 
-def get_owner_user(user_id: str) -> dict:
+def get_owner_user(user_id: str, created_at: Optional[datetime]) -> dict:
```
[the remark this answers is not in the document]
ryanfang5: `create_document` in references/db calls `get_owner_user`, but `created_at` is a keyword argument set to None. Therefore I just typed it with Optional and assigned it to `virtool.utils.timestamp` in case it was None.

virtool/references/data.py
```diff
@@ -0,0 +1,658 @@
...
+                data.organism,
+                data.description,
+                data.data_type,
+                user_id=user_id,
+            )
+
+        document = await attach_computed(self._mongo, document)
+
+        await self._mongo.references.insert_one(document)
+
```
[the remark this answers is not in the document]
ryanfang5: This was done to include `handle` and `administrator` to the user and users fields before adding to the database as this may fix the bug where these values do not show in the find endpoint according to https://preview.virtool.ca/api/refs

virtool/references/data.py
```diff
@@ -0,0 +1,658 @@
...
+        data = await paginate(
+            self._mongo.references,
+            db_query,
+            query,
+            sort="name",
+            base_query=base_query,
+            projection=virtool.references.db.PROJECTION,
+        )
+
+        await apply_transforms(data["documents"], [AttachUserTransform(self._mongo)])
```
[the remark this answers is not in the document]
ryanfang5: You are right, I simply copied the code without taking a closer look at it. My mistake.

ryanfang5:
Unmerged edit from references/db to edit_reference in references/data for testing.

Reverted change so that the reference is inserted before attach_computed is called.

virtool/references/utils.py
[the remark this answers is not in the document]
igboyes: I've reviewed this and it looks like `create_document()` is always passed a value for `created_at`. It should not be optional or have a default value of `None`.

Can you verify this and make the change accordingly.

ryanfang5:
Added clone and remote reference tests for POST /refs. Moved tests for creating an OTU into refs and reordered existing tests.

Also, `newer` was not set in the release when creating a remote reference, so I have hard-coded it to False each time.

virtool/references/data.py
```diff
@@ -0,0 +1,685 @@
...
+        if not await virtool.mongo.utils.id_exists(self._mongo.references, ref_id):
+            raise ResourceNotFoundError()
+
+        data = await virtool.otus.db.find(
+            self._mongo, names, term, query, verified, ref_id
+        )
+
+        return OTUSearchResult(**data)
+
+    async def create_otus(self, ref_id: str, data: CreateOTURequest, req, user_id: str) -> OTU:
```
igboyes, line 396:
```suggestion
    async def create_otu(self, ref_id: str, data: CreateOTURequest, req, user_id: str) -> OTU:
```

virtool/references/data.py
```diff
@@ -0,0 +1,685 @@
...
+
+        otu = await self.data.otus.create(
+            ref_id,
+            data,
+            user_id=user_id
+        )
+
+        return otu
+
+    async def get_history(
```
igboyes, line 423:
```suggestion
    async def find_history(
```

virtool/references/data.py
```diff
@@ -0,0 +1,685 @@
...
+        if not await virtool.references.db.check_right(req, ref_id, "modify"):
+            raise InsufficientRights()
+
+        await virtool.references.db.delete_group_or_user(
+            self._mongo, ref_id, user_id, "users"
+        )
+
+        raise HTTPNoContent
+
+    async def edit_reference(self, ref_id: str, data: dict) -> dict:
```
igboyes, line 667: Why is this separate from the `update()` method?

igboyes (approved):

PR #2690, opened 2022-09-26
feat: paginate uploads

ryanfang5:
BREAKING CHANGE: Change API response shape for uploads

### DIFF
--- a/virtool/uploads/api.py
+++ b/virtool/uploads/api.py
@@ -7,7 +7,7 @@ from aiohttp.web_fileresponse import FileResponse
 from aiohttp.web_response import Response
 from aiohttp_pydantic import PydanticView
 from aiohttp_pydantic.oas.typing import r200, r201, r204, r401, r403, r404
-from pydantic import Field
+from pydantic import Field, conint
 
 from virtool.api.response import InvalidQuery, json_response, NotFound
 from virtool.data.errors import ResourceNotFoundError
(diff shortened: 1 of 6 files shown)


virtool/uploads/data.py
```diff
@@ -45,14 +48,36 @@ async def find(self, user, upload_type, ready) -> UploadSearchResult:
...
+            if page > 1:
+                skip_count = (page - 1) * per_page
+
+            results = await session.execute(
+                select(SQLUpload).filter(*filters).offset(skip_count).limit(per_page)
+            )
+
+            total_count = await session.execute(select(SQLUpload))
+
+            found_count = await session.execute(select(SQLUpload).filter(*filters))
```
igboyes, line 62: The way you are getting the counts is not optimal. Please figure out how to just get the counts via query instead of loading all of the results and using `len()` in Python.

Can we do this with fewer queries? I feel it is likely we can do something similar to `aggregate` in MongoDB where we can get multiple values back from a single query. If we are stuck with multiple queries, using `asyncio.gather` to run them concurrently instead of  serially.

virtool/uploads/api.py
```diff
@@ -30,6 +30,8 @@ class UploadsView(PydanticView):
     async def get(
         self,
         user: Optional[str] = None,
+        page: conint(gt=0) = 1,
+        per_page: conint(ge=1, le=100) = 25,
```
igboyes, line 34: A bit more consistent.
```suggestion
        page: conint(ge=1) = 1,
        per_page: conint(ge=1, le=100) = 25,
```

igboyes (changes requested):

igboyes (changes requested):
I don't believe this will return the full list of uploads for the non-paginated response. It still uses the pagination parameters even if paginate is not `true`. Correct?

ryanfang5:
> I don't believe this will return the full list of uploads for the non-paginated response. It still uses the pagination parameters even if paginate is not `true`. Correct?

Right, offset and limit should not be used if paginate is False. I will go ahead and fix this.

PR #2693, opened 2022-09-26
Move user session storage into redis

ReeceHoffmann:
Changes:
  - Changes session storage from mongo into redis
  - Slight refactor of account datalayer to improve readability of account reset related functions, further refactors needed.
  - Update tests as needed to accommodate the switch to redis for authentication

### DIFF
--- a/virtool/http/auth.py
+++ b/virtool/http/auth.py
@@ -117,7 +117,7 @@ async def authenticate_with_b2c(req: Request, handler: Callable) -> Response:
 
     try:
         token_claims = await validate_token(req.app, token)
-    except (JWTClaimsError, JWTError, ExpiredSignatureError):
+    except JWTError:
         raise HTTPUnauthorized(text="Invalid authorization")
 
     user = await get_data_from_req(req).users.find_or_create_b2c_user(
--- a/virtool/http/auth.py
+++ b/virtool/http/auth.py
@@ -190,7 +190,6 @@ async def middleware(req, handler) -> Response:
     if session is None:
         session_id, session, session_token = await create_session(db, redis, ip)
 
-
     if session_token:
         req["client"] = UserClient(
             db,
--- a/virtool/http/auth.py
+++ b/virtool/http/auth.py
@@ -4,8 +4,7 @@ from typing import Callable, Tuple
 from aiohttp import BasicAuth, web
 from aiohttp.web import Request, Response
 from aiohttp.web_exceptions import HTTPUnauthorized
-from jose import ExpiredSignatureError
-from jose.exceptions import JWTClaimsError, JWTError
+from jose.exceptions import JWTError
 
 from virtool.data.utils import get_data_from_req
 from virtool.errors import AuthError
--- a/virtool/http/auth.py
+++ b/virtool/http/auth.py
@@ -4,7 +4,7 @@ from typing import Callable, Tuple
 from aiohttp import BasicAuth, web
 from aiohttp.web import Request, Response
 from aiohttp.web_exceptions import HTTPUnauthorized
-from jose.exceptions import JWTError
+from jose.exceptions import JWTError, JWTClaimsError, ExpiredSignatureError
 
 from virtool.data.utils import get_data_from_req
 from virtool.errors import AuthError
@@ -116,7 +116,7 @@ async def authenticate_with_b2c(req: Request, handler: Callable) -> Response:
 
     try:
         token_claims = await validate_token(req.app, token)
-    except JWTError:
+    except (JWTClaimsError, JWTError, ExpiredSignatureError):
         raise HTTPUnauthorized(text="Invalid authorization")
 
     user = await get_data_from_req(req).users.find_or_create_b2c_user(
--- a/virtool/http/auth.py
+++ b/virtool/http/auth.py
@@ -155,6 +155,7 @@ async def middleware(req, handler) -> Response:
     :return: the response
     """
     db = req.app["db"]
+    redis = req.app["redis"]
 
     if isinstance(get_handler_policy(handler, req.method), PublicRoutePolicy):
         req["client"] = UserClient(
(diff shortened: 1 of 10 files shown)


virtool/http/auth.py
```diff
@@ -117,7 +116,7 @@ async def authenticate_with_b2c(req: Request, handler: Callable) -> Response:
 
     try:
         token_claims = await validate_token(req.app, token)
-    except (JWTClaimsError, JWTError, ExpiredSignatureError):
+    except JWTError:
```
igboyes, line 119: Why? Did you actually test this with B2C?

virtool/users/sessions.py
```diff
@@ -29,9 +33,7 @@ async def create_session(
         expires_at = utc.shift(minutes=10)
 
     new_session = {
-        "_id": session_id,
-        "created_at": virtool.utils.timestamp(),
-        "expiresAt": expires_at.datetime,
+        "created_at": virtool.utils.timestamp().timestamp(),
```
igboyes, line 36: This works?

virtool/users/sessions.py
```diff
@@ -86,34 +88,38 @@ async def get_session(
...
 
-    if document is None:
+    unparsed_session = None
+    if session_id:
+        unparsed_session = await redis.get(session_id)
+
+    if not unparsed_session:
         return None, None
 
+    session = json.loads(unparsed_session)
```
igboyes, line 105: Do you know what will happen if no session matching the id is found?

If `redis.get()` just returns `None` when a value is not found, then this code can be simplified a bit.

If an exception is raised, handling is missing.

virtool/users/sessions.py
```diff
@@ -131,42 +137,42 @@ async def create_reset_code(
...
-        {
-            "$set": {
-                "reset_code": reset_code,
-                "reset_remember": remember,
-                "reset_user_id": user_id,
-            }
-        },
-    )
+    session = json.loads(await redis.get(session_id))
+    session_expiry = await redis.ttl(session_id)
```
igboyes, line 141: Use `asyncio.gather` here to make requests concurrently.

Extra context from `session_` is not required IMO.

virtool/users/sessions.py
```diff
@@ -131,42 +137,42 @@ async def create_reset_code(
...
+
+    session.update({"reset_code": reset_code, "reset_remember": remember, "reset_user_id": user_id})
+    await redis.set(session_id, json.dumps(session, cls=CustomEncoder), expire=session_expiry)
```
igboyes, line 144:
```suggestion
    await redis.set(session_id, json.dumps({
      **session,
      "reset_code": reset_code,
      "reset_remember": remember,
      "reset_user_id": user_id
    }, cls=CustomEncoder), expire=session_expiry)
```

virtool/users/sessions.py
```diff
@@ -131,42 +137,42 @@ async def create_reset_code(
...
-    await db.sessions.update_one(
-        {"_id": session_id},
-        {"$unset": {"reset_code": "", "reset_remember": "", "reset_user_id": ""}},
-    )
+    session = json.loads(await redis.get(session_id))
+    session_expiry = await redis.ttl(session_id)
+
+    for key in ["reset_code", "reset_remember", "reset_user_id"]:
+        if key in session:
+            del session[key]
```
igboyes, line 163:
```suggestion
    session = {key: session[key] for key in session if key not in {"reset_code", "reset_remember", "reset_user_id"}}
```
You could go further avoid declaring `session` again and stick this right in the `dumps`  call.

igboyes (changes requested):
In addition to the changes in the review, please configure the **BlackConnect** plugin in Pycharm and make sure all your modified files are auto-formatted.

Install Black globally with `sudo python3 -m pip install black[d]`.

virtool/users/sessions.py
igboyes: Nevermind. Simplfication is not possible.

virtool/http/auth.py
ReeceHoffmann: Deepsource, no. Reverted. Deepsource is probably right, but I am not convinced that it is a good idea to merge without being able to test it.

virtool/users/sessions.py
ReeceHoffmann: Yeah, kinda ugly, but the first timestamp call is to the `virtool.utils.timestamp()` which returns a datetime object, while the second is a call to `datetime.timestamp()`. The second call converts the datetime object to POSIX float timestamp that is compatible with the limitations of redis's allowed datatypes and json's decoding. I could sidestep this a bit by calling `arrow.utcnow().naive` right in this function, but it doesn't make things better imo.

virtool/users/sessions.py
ReeceHoffmann: It returns `None` if the key is not found

virtool/users/sessions.py
ReeceHoffmann: > Use `asyncio.gather` here to make requests concurrently.
Done, good call.

> Extra context from `session_` is not required IMO.
Not sure what you mean here? The limitations of `redis.set()` mean that the whole json string is set every time, so I need the full session data. I could probably prevent this by switching to storing the data in a redis hash, but that will require writing a custom decoder encoder for the non-compatible datatypes. Certainly doable and I can swtich course, but last we talked the plan was to stick with json.

virtool/users/sessions.py
ReeceHoffmann: Nice, changed

virtool/users/sessions.py
ReeceHoffmann: Changed to match styles, but I feel compelled to point out that this forces iteration through the whole dict. Since the session dict is so small the performance downgrade of using a dict comprehension should be minimal, but trading O(1) for O(n) hurts me a little.

virtool/http/auth.py
igboyes: Oh. I guess the two removed errors are subclasses of `JWTError`. Let's just leave it.

virtool/users/sessions.py
igboyes: At this scale it will make no difference, however feel free to use your implementation if it hurts you. I would suggest using try-except though instead of `if`.

The payoffs (negligible) of my suggestion are:
* Resistant to `KeyError`.
* Does not mutate the dictionary.

ReeceHoffmann: Nah I am content to go with you solution. I agree that the performance difference would be negligible and I would rather keep the code consistent given how small the session dict is. If the dict were to really grow (which I don't foresee) then the performance difference could matter but not as it is now.

PR #2694, opened 2022-09-28
chore: add data layer for indexes

buddy326:


buddy326:
The only failing tests are for Uploads as support for pagination has not been added in that endpoint yet.

tests/indexes/test_api.py
```diff
@@ -21,62 +21,138 @@
...
+                    "created_at": static_time.datetime,
+                    "manifest": {"foo": 2},
+                    "ready": False,
+                    "has_files": True,
+                    "job": {"id": "foo"},
+                    "reference": {"id": "foo"},
+                    "user": {"id": user.id},
+                    "sequence_otu_map": {"foo": "foo_otu"},
+                },
+            ]
```
igboyes, line 56: This is not a very effective test as it doesn't check that both ready and non-ready indexes are returned for a plain `GET /indexes`.

tests/indexes/test_api.py
```diff
@@ -21,62 +21,138 @@
...
+                    "created_at": static_time.datetime,
+                    "manifest": {"foo": 2},
+                    "ready": True,
+                    "has_files": True,
+                    "job": {"id": "foo"},
+                    "reference": {"id": "foo"},
+                    "user": {"id": user.id},
+                },
+            ]
+        )
```
igboyes, line 116: Ineffective. Both fake indexes have ready set `true` so it doesn't actually test that non-ready indexes are filtered out of the result.

I would combine this with the above test and use parametrization to vary the value for `ready`.

virtool/indexes/data.py
```diff
@@ -0,0 +1,303 @@
...
+        if index is None:
+            raise ResourceNotFoundError()
+
+        ref_id = index["reference"]["id"]
+
+        json_path = (
+            join_index_path(self._config.data_path, ref_id, index_id) / "otus.json.gz"
+        )
+
+        if not json_path.exists():
```
igboyes, line 168: If the filesystem is under load this might block the application unless we use `run_in_thread`.
```suggestion
        if not await run_in_thread(json_path.exists):
```

virtool/indexes/data.py
```diff
@@ -0,0 +1,303 @@
...
+        self, index_id: str, req_query: MultiDictProxy[str]
+    ) -> HistorySearchResult:
+        """
+        Find the virus changes that are included in a given index build.
+
+        :param index_id: the index ID
+        :param req_query: the request query object
+        :return: the changes
+        """
+        if not await self._db.indexes.count_documents({"_id": index_id}):
```
igboyes, line 261:
```suggestion
        if not await id_exists(self._db.indexes, index_id):
```

virtool/indexes/data.py
```diff
@@ -0,0 +1,303 @@
...
+            raise ResourceNotFoundError()
+
+        db_query = {"index.id": index_id}
+
+        term = req_query.get("term")
```
igboyes, line 266: Move this to the request handler. It has the responsibility of interpreting and validation request parameters.

virtool/indexes/data.py
```diff
@@ -0,0 +1,303 @@
...
+        Delete an index given it's id.
+
+        :param index_id: the ID of the index to delete
+        """
+
+        delete_result = await self._db.indexes.delete_one({"_id": index_id})
+
+        if delete_result.deleted_count != 1:
+            # Document could not be deleted.
+            raise ResourceNotFoundError()
```
igboyes, line 293: Comment is redundant. Use simpler conditional checks whenever possible.
```suggestion
        if delete_result.deleted_count == 0:
            raise ResourceNotFoundError()
```

igboyes (changes requested):

PR #2704, opened 2022-10-03
chore: use query and path parameter validation

ryanfang5:


virtool/samples/api.py
```diff
@@ -356,7 +348,7 @@ async def job_remove(req):
 
 @routes.view("/samples/{sample_id}/analyses")
 class AnalysesView(PydanticView):
-    async def get(self) -> Union[r200[List[GetSampleAnalysesResponse]], r403, r404]:
+    async def get(self, sample_id: str, /, term: Optional[str]) -> Union[r200[List[GetSampleAnalysesResponse]], r403, r404]:
```
igboyes, line 351: Use `Field` to add a description so this can be shown in the docs.

Write a description similar to that for labels. Check the pagination query for the actual fields we are text matching.

virtool/users/api.py
```diff
@@ -32,7 +32,7 @@
 @routes.view("/users")
 class UsersView(PydanticView):
     @policy(AdministratorRoutePolicy)
-    async def get(self) -> Union[r200[User], r403]:
+    async def get(self, find: Optional[str]) -> Union[r200[User], r403]:
```
igboyes, line 35: Use `Field` to add a description so this can be shown in the docs. Write a description similar to that for labels.

igboyes (changes requested):

tests/jobs/test_api.py
```diff
@@ -33,6 +33,27 @@ async def test_find_beta(archived, state, fake, snapshot, spawn_client):
...
+    user = await fake2.users.create()
+
+    test_job["user"] = {"id": user.id}
+
+    await client.db.jobs.insert_one(test_job)
+
+    url = "/jobs?"
+
+    if job_filter:
+        url += f"&filter={job_filter}"
```
igboyes, line 49: The `&` is only required when there is more than one path parameter.
```suggestion
    url = "/jobs"

    if job_filter:
        url += f"?filter={job_filter}"
```

igboyes (changes requested):

igboyes (approved):

PR #2710, opened 2022-10-04
fix: rename id path parameter in uploads

ryanfang5:


igboyes (changes requested):
I don't think `alias` is actually needed here since it is a path parameter, not a query parameter. Make sense?

ryanfang5:
> I don't think `alias` is actually needed here since it is a path parameter, not a query parameter. Make sense?

I think it is still required as the path parameter must be the same name as in the view. The tests are failing without `alias` for me.

igboyes:
@ryanfang5 I think the following change is acceptable to make things clearer:
```py
@routes.view("/uploads/{upload_id}")
```
It is the same thing we do everywhere else.

ryanfang5:
Experienced transient transaction error in`test_import_reference_task`, added pytest rerun which should fix it.

ryanfang5:
@igboyes Getting a API rate limit exceeded error, not sure what should be done about this.

PR #2712, opened 2022-10-05
feat: use orjson to serialize responses

ryanfang5:


ryanfang5:
Used version 2.33.6 of virtool-core as the latest version causes a pydantic validation error which I believe should be resolved in another commit.

Uses orjson to serialize json responses, however extra methods that return strings needed to be made as SQLAlchemy expects strings.

It was possible to get rid of `CustomEncoder`- `isoformat` could also be replaced but I am not sure if it is necessary.

virtool/api/custom_json.py
```diff
@@ -44,28 +25,59 @@ def isoformat(obj: datetime.datetime) -> str:
     return obj.replace(tzinfo=datetime.timezone.utc).isoformat().replace("+00:00", "Z")
 
 
-def dumps(obj: object) -> str:
+def default(obj):
```
igboyes, line 28: Write a docstring that describes where this fits in the grand scheme.

virtool/api/custom_json.py
```diff
@@ -44,28 +25,59 @@ def isoformat(obj: datetime.datetime) -> str:
...
+    )
 
 
-def pretty_dumps(obj: object) -> str:
+def pretty_dumps(obj: object) -> bytes:
     """
     A wrapper for :func:`json.dumps` that applies pretty formatting to the output.
 
     Sorts keys and adds indentation. Used as ``dumps`` argument for
     :func:`.json_response`.
```
igboyes, line 56: This docstring needs an update.

virtool/api/custom_json.py
```diff
@@ -44,28 +25,59 @@ def isoformat(obj: datetime.datetime) -> str:
...
+    """
+    return pretty_dumps(obj).decode()
+
+
+def orjson_serializer(obj) -> str:
+    """
+    Used by SQLAlchemy as they expect strings.
     """
-    return json.dumps(obj, cls=CustomEncoder, indent=4, sort_keys=True)
+    return dumps(obj).decode()
```
igboyes, line 83: Can you explain why there are two of these?

virtool/indexes/api.py
```diff
@@ -155,7 +154,7 @@ async def download_otus_json(req):
             db, req.app["config"], index["manifest"]
         )
 
-        json_string = json.dumps(patched_otus, cls=CustomEncoder)
+        json_string = dumps(patched_otus)
 
         await run_in_thread(compress_json_with_gzip, json_string, json_path)
```
igboyes, line 159: May as well.
```suggestion
        await run_in_thread(compress_json_with_gzip, dumps(patched_otus), json_path)
```

virtool/indexes/tasks.py
```diff
@@ -89,7 +89,7 @@ async def _add_one_json(self, index: Document):
 
             await run_in_thread(
                 compress_json_with_gzip,
-                json.dumps(
+                orjson.dumps(
```
igboyes, line 92: Just use your custom `dumps`. What is `"targets"` gets a `datetime` in it at some point.

virtool/pg/utils.py
```diff
@@ -36,7 +37,8 @@ async def connect(postgres_connection_string: str) -> AsyncEngine:
     try:
         pg = create_async_engine(
             postgres_connection_string,
-            json_serializer=pretty_dumps,
+            json_serializer=pretty_orjson_serializer,
```
igboyes, line 40: Why pretty?

virtool/api/custom_json.py
```diff
@@ -9,30 +9,11 @@
...
-        if isinstance(obj, datetime.datetime):
-            return isoformat(obj)
-
-        if issubclass(type(obj), BaseModel):
-            return obj.dict(by_alias=True)
-
-        return json.JSONEncoder.default(self, obj)
-
-
 def isoformat(obj: datetime.datetime) -> str:
```
igboyes, line 17: Are we still using this?

igboyes (changes requested):

virtool/api/custom_json.py
ryanfang5: In pg/utils, the json_serializer was `pretty_dumps`. However, in pg/testing the serializer was `dumps`, so I made two to preserve that functionality.

virtool/pg/utils.py
ryanfang5: I will change this to regular dumps and get rid of `pretty_orjson_serializer`.

virtool/api/custom_json.py
igboyes: That seems backwards to me.

I think we should get rid of the pretty serializer. The use I see for pretty is that the JSONB looks good in a snapshot or during debugging. There are better ways to do that than storing pretty JSON in Postgres.

Agree?

virtool/api/custom_json.py
igboyes: Ah. Thanks for that.

PR #2725, opened 2022-10-14
feat: support setting group name during creation

ryanfang5:


virtool/groups/api.py
```diff
@@ -48,17 +48,17 @@ async def post(
...
-            group = await get_data_from_req(self.request).groups.create(group_id)
+            group = await get_data_from_req(self.request).groups.create(name)
         except ResourceConflictError:
             raise HTTPBadRequest(text="Group already exists")
 
         return json_response(
             GroupResponse.parse_obj(group),
             status=201,
-            headers={"Location": f"/groups/{group.id}"},
+            headers={"Location": f"/groups/{group.name}"},
```
igboyes, line 61: Location still uses the `id`.
```suggestion
            headers={"Location": f"/groups/{group.id}"},
```

igboyes (changes requested):
I take it back. The names should be unique.

I had a brainfart during the meeting and thought the groups were in SQL. You can add a unique index on the groups collection in MongoDB. Search for a large block of code calling `create_index` and add a line to create the unique index there.

Please test this.

PR #2726, opened 2022-10-17
feat: support filtering jobs by user

ryanfang5:


virtool/jobs/api.py
```diff
@@ -25,7 +25,14 @@
...
 @routes.view("/jobs")
 class JobsView(PydanticView):
-    async def get(self) -> Union[r200[List[GetJobResponse]], r400]:
+    async def get(
+        self,
+        archived: Optional[bool] = None,
+        page: conint(ge=1) = 1,
+        per_page: conint(ge=1, le=100) = 25,
+        state: List[str] = Field(default_factory=list),
+        users: List[str] = Field(default_factory=list),
```
igboyes, line 34:
```suggestion
        user: List[str] = Field(default_factory=list),
```

virtool/jobs/api.py
```diff
@@ -25,7 +25,14 @@
...
+    ) -> Union[r200[List[GetJobResponse]], r400]:
```
igboyes, line 35: Isn't this endpoint paginated? If so, this response model is incorrect.

tests/jobs/test_api.py
```diff
@@ -18,6 +19,10 @@ async def test_find_beta(archived, state, fake, snapshot, spawn_client):
     if state is not None:
         url += f"&state={state}"
 
+    if users is not None:
+        for user in users.split(","):
```
igboyes, line 23: Is `split` still necessary here?

igboyes (changes requested):

tests/jobs/test_api.py
ryanfang5: Not sure why I kept this, I have changed the parametrization to pass in a list of users.

igboyes (approved):

PR #2736, opened 2022-11-16
feat: connect to openfga on start

ryanfang5:


virtool/auth/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+    try:
+        api_client = openfga_sdk.ApiClient(configuration)
+        api_instance = open_fga_api.OpenFgaApi(api_client)
+
+        await create_store(api_instance, configuration)
+        await write_auth_model(api_instance)
+
+    except ClientConnectorError:
+        logger.fatal("Could not connect")
```
igboyes, line 31: This is missing an important call. See other uses of `logger.fatal` in Virtool for guidance.

virtool/auth/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        configuration.store_id = response.stores[0].id
+
+    logger.info("Configuring store id")
+
+
+async def write_auth_model(api_instance):
+    response = await api_instance.read_authorization_models()
+
+    if not response.authorization_models:
+
```
igboyes, line 57:
```suggestion
```

virtool/auth/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        response = await api_instance.create_store(body)
+        logger.info("Creating store")
+        configuration.store_id = response.id
+    else:
+        configuration.store_id = response.stores[0].id
+
+    logger.info("Configuring store id")
+
+
+async def write_auth_model(api_instance):
```
igboyes, line 53: Missing typing and docstring.

virtool/auth/utils.py
```diff
@@ -0,0 +1,93 @@
...
+                        upload_file=Userset(
+                            this=dict(),
+                        ),
+                    ),
+                ),
+            ],
+        )
+
+        await api_instance.write_authorization_model(type_definitions)
+        logger.info("Writing authorization model")
```
igboyes, line 93: This log message is inaccurate as the model has already been written at this point. I think it can be dropped completely.

virtool/startup.py
```diff
@@ -261,6 +263,8 @@ async def startup_databases(app: Application):
 
     app["pg"] = pg
 
+    app["auth"] = auth
```
igboyes, line 266:
```suggestion
    app.update({
        "pg": pg,
        "auth": auth
    })
```

virtool/auth/utils.py
```diff
@@ -0,0 +1,93 @@
...
+    Userset,
+)
+from openfga_sdk.api import open_fga_api
+
+logger = getLogger("OpenFGA")
+
+
+async def connect_openfga():
+    configuration = openfga_sdk.Configuration(
+        api_scheme="http", api_host="localhost:8080"
```
igboyes, line 18: OpenFGA is not always going to be running on `localhost`. This will even be a problem in Docker Compose where each service has its own IP and DNS name within the Docker network.

These values need to be configurable via CLI and have reasonable defaults.

virtool/auth/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        body = CreateStoreRequest(
+            name="Virtool",
+        )
+        response = await api_instance.create_store(body)
+        logger.info("Creating store")
+        configuration.store_id = response.id
+    else:
+        configuration.store_id = response.stores[0].id
+
+    logger.info("Configuring store id")
```
igboyes, line 50:
```suggestion
```

igboyes (changes requested):
Is there any way to retrieve the OpenFGA server version and log it like has been done with all the other services?

tests/fixtures/client.py
```diff
@@ -83,6 +83,8 @@ def _create_app(dev: bool = False, base_url: str = ""):
             postgres_connection_string=pg_connection_string,
             redis_connection_string=redis_connection_string,
             fake=False,
+            fga_api_scheme="http",
+            fga_api_host="localhost:8080",
```
igboyes, line 87:
```suggestion
            openfga_scheme="http",
            openfga_host="localhost:8080",
```

virtool/auth/utils.py
```diff
@@ -7,33 +8,53 @@
...
         api_client = openfga_sdk.ApiClient(configuration)
+
+        await check_openfga_version(api_client)
+
         api_instance = open_fga_api.OpenFgaApi(api_client)
 
-        await create_store(api_instance, configuration)
+        store_id = await get_or_create_store(api_instance)
+
+        configuration.store_id = store_id
```
igboyes, line 41:
```suggestion
        configuration.store_id  = await get_or_create_store(api_instance)
```

virtool/auth/utils.py
```diff
@@ -7,33 +8,53 @@
...
 
     except ClientConnectorError:
         logger.fatal("Could not connect")
+        sys.exit(1)
 
     return api_client
 
 
-async def create_store(api_instance, configuration):
+async def get_or_create_store(api_instance):
```
igboyes, line 52: You can clean this function up quite a bit. Start by flipping the conditional so it is `if response.stores`. If you have `if not something: ... else` it should always be reversed to reduce mental overhead.

Then, think about how can further slim down the conditional.

virtool/auth/utils.py
```diff
@@ -90,4 +113,15 @@ async def write_auth_model(api_instance):
...
 
         await api_instance.write_authorization_model(type_definitions)
-        logger.info("Writing authorization model")
+
+
+async def check_openfga_version(client: ApiClient):
+    """
+    Check the OpenFGA version.
+
+    :param client: the application client object
```
igboyes, line 122: Not specific enough. Please correct this elsewhere as well.
```suggestion
    :param client: the application OpenFGA client
```

virtool/config/cls.py
```diff
@@ -4,6 +4,8 @@
 
 @dataclass
 class Config:
+    fga_api_scheme: str
+    fga_api_host: str
     db_connection_string: str
```
igboyes, line 9: Please alphabetize these fields like the existing class.

virtool/auth/utils.py
```diff
@@ -90,4 +113,15 @@ async def write_auth_model(api_instance):
...
+    """
+
+    version = client.user_agent
+
+    logger.info("Found OpenFGA %s", version)
```
igboyes, line 127:
```suggestion
    logger.info("Found OpenFGA %s", client.user_agent)
```

virtool/config/cli.py
```diff
@@ -74,6 +74,18 @@ def entry():
...
     default="https://9a2f8d1a3f7a431e873207a70ef3d44d:ca6db07b82934005beceae93560a6794@sentry.io/220532",
 )
+@click.option(
+    "--fga-api-scheme",
+    help="The OpenFGA API scheme",
+    type=str,
+    default="http",
+)
+@click.option(
+    "--fga-api-host",
```
igboyes, line 84: Fix these to match previous comment.

virtool/config/cli.py
```diff
@@ -74,6 +74,18 @@ def entry():
     type=str,
     default="https://9a2f8d1a3f7a431e873207a70ef3d44d:ca6db07b82934005beceae93560a6794@sentry.io/220532",
 )
+@click.option(
+    "--fga-api-scheme",
+    help="The OpenFGA API scheme",
+    type=str,
+    default="http",
```
igboyes, line 81: Default should actually be `https`. This is the default in the docs for OpenFGA, I believe.

Your default value should always be secure. Let people opt into less secure methods.

igboyes (changes requested):

virtool/auth/utils.py
```diff
@@ -0,0 +1,121 @@
...
+        await write_auth_model(api_instance)
+
+    except ClientConnectorError:
+        logger.fatal("Could not connect")
+        sys.exit(1)
+
+    return api_client
+
+
+async def get_or_create_store(api_instance):
```
igboyes, line 50: Types missing.

virtool/auth/utils.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    Get the OpenFGA Store or create one if it does not exist.
+
+    :return: the store id
+    """
+
+    response = await api_instance.list_stores()
+
+    if response.stores:
```
igboyes, line 59:
```suggestion

    if response.stores:
        logger.info("Found existing store")
```

virtool/auth/utils.py
```diff
@@ -0,0 +1,121 @@
...
+
+    body = CreateStoreRequest(
+        name="Virtool",
+    )
+    logger.info("Creating store")
+    response = await api_instance.create_store(body)
+    return response.id
+
+
+async def write_auth_model(api_instance):
```
igboyes, line 70: Types missing.

igboyes (changes requested):
Last one.

PR #2739, opened 2022-11-18
fix: update dockerfile

ryanfang5:


igboyes (changes requested):
Commit message is not appropriate for release notes. Reword to say what problem was resolved.

igboyes (approved):

PR #2741, opened 2022-11-21
chore: add check functions for analysis data layer

buddy326:


virtool/analyses/data.py
```diff
@@ -374,7 +376,7 @@ async def finalize(self, analysis_id: str, results: dict) -> Analysis:
             raise ResourceNotFoundError()
 
         if "ready" in analysis_document and analysis_document["ready"]:
-            raise ResourceConflictError
+            raise ResourceConflictError()
```
igboyes, line 379: This has no effect. Leave off round brackets unless passing a parameter.

tests/analyses/test_checks.py
```diff
@@ -0,0 +1,97 @@
+from datetime import datetime
```
igboyes, line 1: I find the tests in this module unecessarily hard to read.

I think talking a class approach would be clearer. For each check function, you can have a class with two test methods: `test_ok`, `test_error`. The methods will end up being very obvious compared to the parametrized tests below.

Parametrization is used a lot in the API tests because they would be repetivitive if we didn't. The tests in this module are better without parametrization.

tests/analyses/test_checks.py
```diff
@@ -0,0 +1,97 @@
...
+    jobs_api_flag = True
+    ready = True
+
+    if error:
+        with pytest.raises(ResourceConflictError):
+            await check_if_analysis_ready(jobs_api_flag, ready)
+        return
+
+    jobs_api_flag = False
+    assert await check_if_analysis_ready(jobs_api_flag, ready) is None
```
igboyes, line 45: You're going to refactor this, but why not just:
```suggestion
    if error:
        with pytest.raises(ResourceConflictError):
            await check_if_analysis_ready(True, True)
        return

    jobs_api_flag = False
    assert await check_if_analysis_ready(False, True) is None
```
You do this quite a bit in this PR.

virtool/analyses/checks.py
```diff
@@ -0,0 +1,47 @@
...
+        if if_modified_since == updated_at:
+            raise ResourceNotModifiedError()
+
+
+async def check_if_analysis_ready(jobs_api_flag: bool, ready: bool):
+    if jobs_api_flag and ready:
+        raise ResourceConflictError()
+    else:
+        if not ready:
+            raise ResourceConflictError()
```
igboyes, line 30: The `else` is pointless here. When your initial `if` evaluates `True`, you will immediately leave the function and never get to `else`.

Also note the existence of `elif` in Python.

```suggestion
    if jobs_api_flag and ready:
        raise ResourceConflictError()
        
    if not ready:
        raise ResourceConflictError()
```

virtool/analyses/data.py
```diff
@@ -107,14 +114,11 @@ async def get(
...
-                updated_at = document["updated_at"]
-            except KeyError:
-                updated_at = document["created_at"]
+        aws = []
+
+        aws.append(check_if_analysis_modified(if_modified_since, document))
 
-            if if_modified_since == updated_at:
-                raise ResourceNotModifiedError()
+        await wait_for_checks(*aws)
```
igboyes, line 121: This can be collapsed into one readable line without the declaration of a new list. Make sure you understand how `*` works.

virtool/analyses/data.py
```diff
@@ -192,12 +196,11 @@ async def delete(self, analysis_id: str, jobs_api_flag: bool):
...
-            if document["ready"]:
-                raise ResourceConflictError()
-        else:
-            if not document["ready"]:
-                raise ResourceConflictError()
+        aws = []
+
+        aws.append(check_if_analysis_ready(jobs_api_flag, document["ready"]))
+
+        await wait_for_checks(*aws)
```
igboyes, line 203: Can be one-liner.

virtool/analyses/data.py
```diff
@@ -307,16 +310,15 @@ async def blast(self, analysis_id: str, sequence_index: int) -> Optional[str]:
...
+        aws.append(check_analysis_workflow(document["workflow"]))
 
-        sequence = find_nuvs_sequence_by_index(document, sequence_index)
+        aws.append(check_if_analysis_running(document["ready"]))
 
-        if sequence is None:
-            raise ResourceNotFoundError()
+        aws.append(check_analysis_nuvs_sequence(document, sequence_index))
+
+        await wait_for_checks(*aws)
```
igboyes, line 321: No need for list or multiple expressions here. You can do this in a single call to `wait_for_checks`. Read up on `*` and variadic arguments in Python.

igboyes (changes requested):

igboyes (approved):

PR #2753, opened 2022-12-03
feat: add abstract authorization client for openfga

ryanfang5:


igboyes:
> 1. Should I make another fixture for connecting to openfga for the spawn_auth_client fixture? Right now I am reusing code used in startup.

I think this is fine.

> 2. What is the expected return for if adding or removing a permission in openfga fails? For example, trying to add a tuple that is already in the store. Currently I just have pass.

I think it should stay as is. These are idempotent operations. Agree?

tests/auth/test_client.py
```diff
@@ -0,0 +1,267 @@
...
+    async def func(
+        permissions=None,
+    ):
+        user_document = create_user(
+            user_id="test",
+            permissions=permissions,
+        )
+        await mongo.users.insert_one(user_document)
+
+        open_fga_client = await connect_openfga("localhost:8080", "http")
```
igboyes, line 22: Use `asyncio.gather`.

tests/auth/test_client.py
```diff
@@ -0,0 +1,267 @@
...
+    response = await api_instance.list_stores()
+
+    for store in response.stores:
+        configuration.store_id = store.id
+        await api_instance.delete_store()
+
+
+@pytest.mark.parametrize("db", ["mongo", "open_fga"])
+@pytest.mark.parametrize("has_permission", [True, False])
+async def test_check(delete_store, db, has_permission, spawn_auth_client):
```
igboyes, line 50: Should be a two separate tests and maybe organized into a class: `TestCheck` with `test_mongo` and `test_openfga`.

tests/auth/test_client.py
```diff
@@ -0,0 +1,267 @@
...
+            "virtool",
+        )
+
+        response = await abs_client.check("ryanf", permission, "app", "virtool")
+
+    assert response is has_permission
+
+
+@pytest.mark.parametrize("db", ["mongo", "open_fga"])
+async def test_list(delete_store, db, spawn_auth_client, snapshot):
```
igboyes, line 77: Separate into two tests. Do this elsewhere as well.

tests/auth/test_client.py
```diff
@@ -0,0 +1,267 @@
...
+    await fake2.users.create()
+    await fake2.users.create(groups=[group])
+    await fake2.users.create(groups=[group])
+    await fake2.users.create(groups=[group])
+
+    if db == "mongo":
+        await mongo.groups.find_one_and_update(
+            {"_id": group.id},
+            {"$set": {"permissions.cancel_job": True, "permissions.create_ref": True}},
+        )
```
igboyes, line 207: The document is returned by `find_one_and_update` so it is more expensive.
```suggestion
        await mongo.groups.update_one(
            {"_id": group.id},
            {"$set": {"permissions.cancel_job": True, "permissions.create_ref": True}},
        )
```

tests/auth/test_client.py
```diff
@@ -0,0 +1,267 @@
...
+
+        await write_tuple(
+            abs_client.open_fga,
+            "group",
+            "sidney#member",
+            "create_ref",
+            "app",
+            "virtool",
+        )
```
igboyes, line 237: Use `asyncio.gather`.

virtool/app.py
```diff
@@ -21,6 +21,7 @@
     shutdown_executors,
     shutdown_redis,
     shutdown_scheduler,
+    shutdown_auth_client,
```
igboyes, line 24: Need to be specific as this can be confused with authentication.
```suggestion
    shutdown_authorization_client,
```

virtool/app.py
```diff
@@ -89,14 +91,15 @@ def create_app(config: Config):
...
+            startup_http_client,
             startup_databases,
             startup_dispatcher,
-            startup_http_client,
             startup_paths,
             startup_routes,
             startup_executors,
             startup_task_runner,
             startup_data,
+            startup_auth,
```
igboyes, line 102: Be consistent with `_client`.
```suggestion
            startup_authentication_client,
```

virtool/auth/client.py
```diff
@@ -0,0 +1,126 @@
...
+        permission: Permission,
+        object_type: str,
+        object_id: Union[str, int],
+    ) -> bool:
+        ...
+
+    @abstractmethod
+    async def list_permissions(
+        self, user_id: str, object_type: str, object_id: Union[str, int]
+    ) -> dict:
```
igboyes, line 39: This should return a `list`.

virtool/auth/client.py
```diff
@@ -0,0 +1,126 @@
...
+
+    async def check(
+        self,
+        user_id: str,
+        permission: Permission,
+        object_type: str,
+        object_id: Union[str, int],
+    ) -> bool:
+        """
+        Check a permission in Mongo and OpenFGA.
```
igboyes, line 65:
```suggestion
        Check whether a user has a permission on a resource.
```

virtool/auth/client.py
```diff
@@ -0,0 +1,126 @@
...
+                self.open_fga, user_id, object_type, object_id
+            )
+
+    async def add(self, relationship: BaseRelationship):
+        """
+        Add permissions for a group or user in mongo and OpenFGA.
+        """
+        try:
+            await relationship.add(self.data)
+        except NotFound:
```
igboyes, line 98: Use `ResourceNotFound`.

`NotFound` is a special exception that can bubble up and return an HTTP error response if it is not caught.

virtool/auth/client.py
```diff
@@ -0,0 +1,126 @@
...
+            pass
+
+        for relation in relationship.relation:
+            tuple_key = TupleKey(
+                user=f"{relationship.user_type}:{relationship.user_id}",
+                relation=relation,
+                object=f"{relationship.object_type}:{relationship.object_name}",
+            )
+
+            await add_in_open_fga(self.open_fga, tuple_key)
```
igboyes, line 108: Optimize with `asyncio.gather`. Is there any way this can be done in a single request to OpenFGA?

virtool/auth/client.py
```diff
@@ -0,0 +1,126 @@
...
+
+    async def remove(self, relationship: BaseRelationship):
+        """
+        Remove permissions for a group or user in mongo and OpenFGA.
```
igboyes, line 112:
```suggestion
        Remove an authorization relationship.
```

virtool/auth/mongo.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+async def check_in_mongo(mongo: DB, user_id: str, permission: Permission) -> bool:
+    """
+    Check permissions in Mongo.
+    """
+
+    return (
+        await mongo.users.count_documents(
+            {"_id": user_id, f"permissions.{permission.name}": True}
```
igboyes, line 14: This returns a count as soon as one document is found instead of traversing all documents.
```suggestion
            {"_id": user_id, f"permissions.{permission.name}": True}, limit=1
```

virtool/auth/mongo.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+
+async def list_permissions_in_mongo(mongo: DB, user_id: str) -> dict:
+    """
+    List user permissions in Mongo.
+    """
+    try:
+        user = await mongo.users.find_one(user_id, ["permissions"])
+        return user["permissions"]
```
igboyes, line 26: Use existing Virtool `get_one_field` utility.

virtool/auth/openfga.py
```diff
@@ -0,0 +1,111 @@
...
+        open_fga: ApiClient,
+        user_id: str,
+        permission: Permission,
+        object_type: str,
+        object_id: Union[str, int],
+) -> bool:
+    """
+    Check a permission in OpenFGA.
+    """
+    api_instance = open_fga_api.OpenFgaApi(open_fga)
```
igboyes, line 28: Can't we pass around the `api_instance` instead of `ApiClient`?

virtool/auth/openfga.py
```diff
@@ -0,0 +1,111 @@
...
+    groups = [relation_tuple.key.object for relation_tuple in response.tuples]
+
+    for group in groups:
+        body = ReadRequest(
+            tuple_key=TupleKey(
+                user=f"{group}#member", object=f"{object_type}:{object_id}"
+            ),
+        )
+
+        response = await api_instance.read(body)
```
igboyes, line 69: Use `asyncio.gather` unless possible to do in single request to OpenFGA.

virtool/auth/relationships.py
```diff
@@ -0,0 +1,72 @@
...
+
+    async def add(self, data: DataLayer):
+        permission_dict = {permission.name: True for permission in self.relation}
+
+        data_dict = UpdateGroupRequest(permissions=permission_dict)
+
+        try:
+            await data.groups.update(self.user_id, data_dict)
+        except ResourceNotFoundError:
+            self.user_id = f"{self.user_id}#member"
```
igboyes, line 54: What does this do?

igboyes (changes requested):
Looking good.

virtool/auth/client.py
ryanfang5: Would this just be a list of the allowed permissions without the boolean value? Right now I am returning a dict of permissions such as: 
{'cancel_job': True, 'create_ref': False, 'create_sample': False, 'modify_hmm': False, 'modify_subtraction': False, 'remove_file': False, 'remove_job': False, 'upload_file': False}.

virtool/shutdown.py
```diff
@@ -22,6 +22,20 @@ async def shutdown_client(app: Application):
...
+    Attempt to close the OpenFGA client session.
+
+    :param app: The application object
+    """
+    logger.info("Stopping OpenFGA client")
+
+    try:
+        await app["auth"].open_fga.close()
+    except KeyError:
+        pass
```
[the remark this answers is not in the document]
ryanfang5: Sorry, I just realized my comments did not post from yesterday. I just used the same logic as the other shutdown methods for the app.

virtool/auth/relationships.py
ryanfang5: When adding permissions for a group in OpenFGA, the format must be group:sidney#member for example to work on each member of the group. Therefore I added #member after the mongo attempt so that it would work for OpenFGA.

virtool/auth/client.py
igboyes: We'll figure this out on Monday.

I want to move away from the dictionary model. It doesn't serve much purpose and will be a burden once permissions start spreading to resources other than `app`.

The problem is that the API returns permissions objects like the one shown in your comment. We'll have to strategize on this.

igboyes (changes requested):

PR #2757, opened 2022-12-06
chore: sort uploads by descending created_at

buddy326:


igboyes (changes requested):
Sorting needs to be done using the SQL query.

The current implementation does not fetch the newest _x_ uploads. Instead it is going to fetch _x_ rows then sort only that subset by `created_at`.

igboyes (approved):

PR #2761, opened 2022-12-14
feat: add endpoints for managing users' app permissions

ryanfang5:


igboyes (changes requested):
Please add a test that shows what happens when an invalid permission name is provided in the URL.

igboyes (changes requested):
I discussed this with @ReeceHoffmann and we'd like to return a list of objects from the API like:
```
[
  {"id": "create"sample"},
  {"id": "create_subtraction"}
]
```
It will be easier to add additional fields like `description` or `name` later if needed. Sound good?

Don't change the authz client to do this. Do it in the request handler.

ryanfang5:
> I discussed this with @ReeceHoffmann and we'd like to return a list of objects from the API like:
> 
> ```
> [
>   {"id": "create"sample"},
>   {"id": "create_subtraction"}
> ]
> ```
> 
> It will be easier to add additional fields like `description` or `name` later if needed. Sound good?
> 
> Don't change the authz client to do this. Do it in the request handler.

Do you mean a list of objects that the user has access to, or did you just want me to add {"id": permission} in the current list permissions response? If it is the former, I'm not sure why this wouldn't be implemented in the authz client and called in another endpoint.

igboyes (approved):

PR #2762, opened 2022-12-19
chore: don't run start up tasks during tests

ryanfang5:
Prevents snapshots from failing due to attached tasks

### DIFF
--- a/tests/references/test_api.py
+++ b/tests/references/test_api.py
@@ -1,5 +1,6 @@
 import asyncio
 from asyncio import gather
+from enum import Enum
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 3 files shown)


tests/references/test_api.py
```diff
@@ -209,6 +209,9 @@ async def test_import_reference(
 
         task_id = reference["task"]["id"]
 
+        import_reference_task = ImportReferenceTask(client.app, task_id)
+        await import_reference_task.run()
```
igboyes, line 213: Make another separate test for the task. Keeping the API and task tests together no longer makes sense here.

Please put task test in `references/test_tasks.py` module.

igboyes (changes requested):

tests/references/test_api.py
ryanfang5: Removed this test as it is already done more extensively in `test_import_reference_task`.

igboyes (approved):

PR #2768, opened 2023-01-03
feat: exit if data path is not found

ryanfang5:


igboyes (changes requested):
https://click.palletsprojects.com/en/8.1.x/arguments/#file-path-arguments

```py
@click.argument('filename', type=click.Path(exists=True))
```

igboyes (approved):

PR #2773, opened 2023-01-10
fix: prevent validation errors for apikey

ryanfang5:


virtool/account/data.py
```diff
@@ -133,7 +137,7 @@ async def get_keys(self, user_id: str) -> List[APIKey]:
         """
         cursor = self._db.keys.find({"user.id": user_id}, API_KEY_PROJECTION)
 
-        return [APIKey(**d) async for d in cursor]
+        return [await fetch_complete_key(self._db, d["id"]) async for d in cursor]
```
igboyes, line 140: This isn't very efficient.

I think you can do this in a single request. If that is not possible, at the very least `gather` should be used to execute your `fetch_complete_key` calls concurrently.

virtool/account/db.py
```diff
@@ -63,6 +65,37 @@ async def get_alternate_id(db, name: str) -> str:
         suffix += 1
 
 
+async def fetch_complete_key(mongo, key_id: str) -> Optional[APIKey]:
```
igboyes, line 68: Needs brief docstring.

```suggestion
async def fetch_complete_api_key(mongo, key_id: str) -> Optional[APIKey]:
```

igboyes (changes requested):

igboyes (approved):

PR #2777, opened 2023-01-19
fix: remove default dsn value

ChristineWC:


ReeceHoffmann (changes requested):
Looks mostly good, couple of tweaks though:

1. The sentry DSN is also provided in the `Config` dataclass as a default value. Replace it with an empty string.
2. Double check this, but I think the default value for any click option is an empty string. If so no need to explicitly specify

ReeceHoffmann (approved):

PR #2782, opened 2023-01-27
fix: virtool/analyses/api::AnalysisView::get raises 304 data not modified

SnickerDoodlePop:


igboyes (changes requested):
Returning `304` should happen at lines 105-106 only:
```py
except ResourceNotModifiedError:
    raise HTTPNotModified()
```
Take a look at `virtool.data.AnalysisData.get`. The bug suggests that `ResourceNotModifiedError` is not being raised there as expected. The data modules contain definitions of Virtool's data layer, which is accessed from the request handlers.

Your current implementation will save data from being sent over the wire, but it doesn't avoid building a complete response in the data layer.

We can discuss this afternoon if necessary.

igboyes (approved):

PR #2791, opened 2023-02-07
feat: enable decoupled running of tasks

ReeceHoffmann:
Changes:
 - add cli command "spawn_task" for spawning tasks
 - add cli command "tasks" for executing standalone task runner
 - use TaskClient for pushing and pulling tasks from redis
 - replace use of redis channel with redis list
 - add test for task spawner
 - add "/" base route for task runner

Notes:
 - Running api with `tasks` command starts up a task runner that will work through one task at a time from a list stored in redis
    - Task runners can be run independently or in conjunction with task runners running within the main api server
    - Tested for support up to 3 simultaneous task runners, no expected theoretical limit to the number of runners possible.
 - Running api with "spawn_tasks --task_name <TaskName>" will attempt to create a task of "TaskName" if such a task exists or return an error and exit
   - **NB:** When running the API with `--no-tasks` argument it is necessary to spawn routine jobs via this command, as the independent task runners make no assumptions about the number of other workers. Periodic tasks spawned by an API instance will be picked up by independent task runners.
 - Decoupled task runner is intended to be run in kubernetes. If running the api in docker or locally stick to the coupled task runner

### DIFF
--- a/virtool/tasks/main.py
+++ b/virtool/tasks/main.py
@@ -1,0 +1,103 @@
+import aiojobs
+import aiojobs.aiohttp
+from aiohttp import web
+from aiohttp.web import Application
+from sqlalchemy.util import asyncio
+
+import virtool.http.accept
+import virtool.http.errors
+from virtool.config import Config
+from virtool.dispatcher.events import DispatcherSQLEvents
+from virtool.process_utils import create_app_runner, wait_for_restart, wait_for_shutdown
+from virtool.shutdown import (
+    shutdown_client,
+    shutdown_executors,
+    shutdown_scheduler,
+    shutdown_redis,
+    shutdown_authorization_client,
+)
+from virtool.startup import (
+    startup_events,
+    startup_http_client,
+    startup_databases,
+    startup_paths,
+    startup_executors,
+    startup_data,
+    startup_task_runner,
+    startup_sentry,
+    startup_version,
+    startup_fake,
+    startup_fake_config,
+)
+from virtool.tasks.api import TasksRunnerView
+
+
+async def startup_dispatcher_SQL_listener(app: Application):
+    """
+
+
+    :param app: the app object
+
+    """
+    DispatcherSQLEvents(app["dispatcher_interface"].enqueue_change)
+
+
+async def create_app(config: Config):
+    """
+    Creates the Virtool application.
+
+    """
+    app = Application(
+        middlewares=[
+            virtool.http.accept.middleware,
+            virtool.http.errors.middleware,
+        ]
+    )
+
+    app["config"] = config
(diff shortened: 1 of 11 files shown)


virtool/tasks/main.py
```diff
@@ -0,0 +1,103 @@
...
+    startup_task_runner,
+    startup_sentry,
+    startup_version,
+    startup_fake,
+    startup_fake_config,
+)
+from virtool.tasks.api import TasksRunnerView
+
+
+async def startup_dispatcher_SQL_listener(app: Application):
```
igboyes, line 35:
```suggestion
async def startup_dispatcher_sql_listener(app: Application):
```

virtool/tasks/main.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
```
igboyes, line 37: Missing docstring.

virtool/tasks/spawner.py
```diff
@@ -0,0 +1,92 @@
...
+    return matching_task[0]
+
+
+async def spawn(config: Config, task_name: str):
+    app = await create_app(config)
+
+    task = get_task_from_name(task_name)
+    logger.info("Spawning task %s", task.name)
+
+    await get_data_from_app(app).tasks.create(task)
```
igboyes, line 90: The above is a lot of work for this. Did you have any ideas on workarounds?

igboyes (changes requested):
This looks great! Only a few questions and minor problems.

virtool/tasks/spawner.py
ReeceHoffmann: I agree, and yes it is definitely possible but might be a pain from an upkeep perspective. 

The task spawner only really needs to A) register a task into the db and B) push the task id onto the redis list. Technically I could write up a bespoke solution to do that, but from an upkeep perspective I would way rather make use the `tasks` data_layer so that there is less things that need to be updated in the case that we change what it means to create a task.  This currently involves creating a bunch of connections and extra stuff that isn't really needed though

A good compromise could be to instantiate tasks data_layer on its own without the rest of the data_layers. This would mean that it only needs to connect to redis and pg which is pretty big on its own. Downside here is that if the tasks data_layer ever starts needing `open_fga`, `mongo` or other data_layers then this starts to make in some cases a lot less sense. I don't foresee tasks needing any of those resources though, so it is probably worth the risk.

Short version: Yes, I'll tweak it to only use the tasks data_layer and thereby greatly reduce the number of dependencies

igboyes: OK. I remember discussing this a bit already. You can leave as is if you want.

PR #2794, opened 2023-02-13
fix: add test cases to check for correct workflow filtering

ChristineWC:


tests/samples/test_api.py
```diff
@@ -209,6 +216,10 @@ async def test_find(
         label_query = "&label=".join(str(label) for label in labels)
         query.append(f"label={label_query}")
 
+    if workflows is not None:
+        workflows_query = "&workflows=".join(str(workflow) for workflow in workflows)
```
ReeceHoffmann, line 220: Workflow should already be a string, no need to cast it via `str`

ReeceHoffmann (changes requested):
Nice job getting this test working! That said, I think the number of parameters is getting pretty unwieldy and it might just be best to start defining separate tests all wrapped in a single class rather than adding extra parameters. To this end I would suggest the following: 

1. Factor out the initial test setup into a separate function so that it can be reused between all the cases
2. Create a class called `TestFindSamples` which has methods for running each of the tests split up by what filter is being tested. (ei: one for no filters, one for labels, one for workflows, ... ect)

This should simplify each test type and make it easier to add new tests downstream. There examples of similar thing being done for a couple api routes in the `references/test_api.py` file which could be useful reference for doing the conversion. Feel free to shoot me a message as well if you have any questions.

tests/samples/test_api.py
```diff
@@ -105,31 +105,7 @@ async def get_sample_data(mongo, fake2, pg, static_time):
...
-    per_page,
-    page,
-    labels,
-    snapshot,
-    fake2,
-    spawn_client,
-    static_time,
-    pg: AsyncEngine,
-):
+async def setup_initial_test_setup(fake2, spawn_client, static_time):
```
ReeceHoffmann, line 108: Name is confusing, would change to something like: "setup_find_user_test" as it reduces redundancy as makes it clear that the function is for setting up one particular set of tests

That said, since we are in a class the invocation of the initial test setup can be moved into a setup_method which automates running the setup before every test is run

https://doc.pytest.org/en/latest/how-to/xunit_setup.html

tests/samples/test_api.py
```diff
@@ -188,33 +166,114 @@ async def test_find(
...
+        static_time,
+        snapshot,
+    ):
+        client = await setup_initial_test_setup(fake2, spawn_client, static_time)
+        path = "/samples"
+        query = []
+        if find is not None:
+            query.append(f"find={find}")
+        if query:
+            path += f"?{'&'.join(query)}"
```
ReeceHoffmann, line 194: I suspect this can be greatly simplified as you know that there will only ever be zero or one `find` term passed.

tests/samples/test_api.py
```diff
@@ -188,33 +166,114 @@ async def test_find(
...
+        path = "/samples"
+        query = []
 
-    if labels is not None:
-        label_query = "&label=".join(str(label) for label in labels)
-        query.append(f"label={label_query}")
+        if labels is not None:
+            label_query = "&label=".join(str(label) for label in labels)
+            query.append(f"label={label_query}")
+        if query:
```
ReeceHoffmann, line 239: As with above you can simplify since you are only querying labels

tests/samples/test_api.py
```diff
@@ -188,33 +166,114 @@ async def test_find(
...
+    ):
+        client = await setup_initial_test_setup(fake2, spawn_client, static_time)
+        path = "/samples"
+        query = []
 
-    if labels is not None:
-        label_query = "&label=".join(str(label) for label in labels)
-        query.append(f"label={label_query}")
+        if labels is not None:
+            label_query = "&label=".join(str(label) for label in labels)
```
ReeceHoffmann, line 237: Looks like your appending "&" both here and when you append the query to the path

tests/samples/test_api.py
```diff
@@ -188,33 +166,114 @@ async def test_find(
...
+    ):
+        client = await setup_initial_test_setup(fake2, spawn_client, static_time)
+
+        path = "/samples"
+        query = []
+        if workflows is not None:
+            workflows_query = "&workflows=".join(workflow for workflow in workflows)
+            query.append(f"workflows={workflows_query}")
+
+        if query:
```
ReeceHoffmann, line 271: simplify as above

ReeceHoffmann (changes requested):
Looking better having everything split up. That said I think there is still some simplification that can be done now that we know the number of parameters is limited.

igboyes:
@ChristineWC Is this still a draft?

ReeceHoffmann (approved):
Looks good to me!

PR #2795, opened 2023-02-14
feat: add user lookup to aggregation pipeline

SnickerDoodlePop:


virtool/users/db.py
```diff
@@ -336,3 +336,27 @@ async def fetch_complete_user(mongo, user_id: str) -> Optional[User]:
         )
 
     return None
+
+
+def lookup_user_by_id(local_field: str, set_as: str = "user") -> list[Dict]:
```
ReeceHoffmann, line 341: Should be named `lookup_nested_user_by_id` Since the lookup results in the addition of a `nestedUser` rather than a full `user` object

Also while technically correct, I would prefer a more specific return type than `list[dict]` there may be something defined as part of pymotor. On a related note this function should have a the summarises what it is for and the inputs and outputs of the function. The docstring should make it clear that this function returns the aggregation step to be put into an aggregation pipeline and not the data itself.

virtool/users/db.py
```diff
@@ -336,3 +336,27 @@ async def fetch_complete_user(mongo, user_id: str) -> Optional[User]:
...
+                "from": "users",
+                "let": {"user_id": f"${local_field}"},
+                "pipeline": [
+                    {"$match": {"$expr": {"$eq": ["$_id", "$$user_id"]}}},
+                    {
+                        "$project": {
+                            "id": "$_id",
+                            "_id": 0,
+                            "handle": 1,
+                            "administrator": 1,
```
ReeceHoffmann, line 354: Stolen from Ian's review of another PR: the `ints` should be replaced with bools, makes it a bit more clear what is going on

virtool/users/db.py
```diff
@@ -336,3 +336,27 @@ async def fetch_complete_user(mongo, user_id: str) -> Optional[User]:
...
+
+
+def lookup_user_by_id(local_field: str, set_as: str = "user") -> list[Dict]:
+    return [
+        {
+            "$lookup": {
+                "from": "users",
+                "let": {"user_id": f"${local_field}"},
+                "pipeline": [
+                    {"$match": {"$expr": {"$eq": ["$_id", "$$user_id"]}}},
```
ReeceHoffmann, line 348: Double check, but this can probably be simplified to something along the lines of:
```suggestion
                    {"$match": { "_id", "$$user_id"}},
```

ReeceHoffmann (commented):
Looks mostly good, but on a second glance through there are a some documentation issues and simplification that can be done

virtool/users/db.py
```diff
@@ -336,3 +336,40 @@ async def fetch_complete_user(mongo, user_id: str) -> Optional[User]:
...
+    """
+    Create a mongoDB aggregation pipeline step to look up a nested user by id.
+
+        Params:
+            local_field (str): user id field to search look up (default "user.id");
+
+            set_as (str): desired name of the returned record (default "user").
+
+        Returns: $lookup step (list[Dict]): mongoDB aggregation step to use in
+        an aggregation pipeline; unpack for inline use in aggregation pipeline.
```
ReeceHoffmann, line 353:
```suggestion
    Create a mongoDB aggregation pipeline step to look up a nested user by id.

    :param local_field:  user id field to search look up
    :param set_as: desired name of the returned record
    :return: mongoDB aggregation steps for use in an aggregation pipeline
```

ReeceHoffmann (approved):

PR #2796, opened 2023-02-15
feat: attach job data to analyses

ryanfang5:
Also replaced Subtraction, User and Reference attach transforms with $lookup.

### DIFF
--- a/virtool/analyses/data.py
+++ b/virtool/analyses/data.py
@@ -17,12 +17,11 @@ import virtool.analyses.format
 import virtool.samples.db
 import virtool.uploads.db
 
-from virtool.analyses.db import PROJECTION, TARGET_FILES
+from virtool.analyses.db import TARGET_FILES
 from virtool.analyses.files import create_analysis_file, create_nuvs_analysis_files
 from virtool.analyses.models import AnalysisFile
 from virtool.analyses.utils import (
     attach_analysis_files,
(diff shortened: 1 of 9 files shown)


virtool/analyses/data.py
```diff
@@ -47,79 +46,160 @@
...
+            data = paginate_dict["data"]
+            found_count = paginate_dict.get("found_count", 0)
+            total_count = paginate_dict.get("total_count", 0)
+
+        for analysis in data:
+            analysis["job"] = (
+                await processor(self._db, analysis["job"])
+                if "job" in analysis
+                else None
+            )
```
igboyes, line 177: This is going to kill performance after you have spent all this time optimizing it.

Can this not also be done in the aggregation?

virtool/analyses/data.py
```diff
@@ -47,79 +46,160 @@
...
-        db_query = {}
-
-        data = await paginate(
-            self._db.analyses,
-            db_query,
-            query,
-            projection=PROJECTION,
-            sort=[("created_at", -1)],
-        )
+        dict_projection = {item: True for item in PROJECTION}
```
igboyes, line 89: Bad name and only used once I think. Just declare it where you are using it.

igboyes (changes requested):

igboyes (approved):

PR #2797, opened 2023-02-16
fix: remove mongodb index creation

ChristineWC:


igboyes (changes requested):
This is great!

The only change needed is changing the required revision to `"l20h8fsbbb28"`. This is the migration that applies all of the indexes.

https://github.com/virtool/virtool-migration/blob/main/revisions/rev_l20h8fsbbb28_create_all_database_indexes.py

igboyes (approved):

PR #2801, opened 2023-02-20
fix: elegantly handle redis connection errors in task runner

ReeceHoffmann:


virtool/tasks/client.py
```diff
@@ -1,10 +1,15 @@
...
 from typing import List
 
-from aioredis import Redis
+from aioredis import (
+    Redis,
+    ConnectionClosedError,
+)
 
 REDIS_TASKS_LIST_KEY = "tasks"
+logging.getLogger("tasksClient")
```
igboyes, line 12: I'd like to keep these more general and limited in number. There were too many logger names before and it made it hard to scan the logs visually.
```suggestion
logging.getLogger("tasks")
```

virtool/tasks/client.py
```diff
@@ -22,15 +27,26 @@ def __init__(self, redis: Redis):
...
-
-            if result is not None:
-                return int(result[1])
+        result = await self._blpop(self.redis)
+
+        if result is not None:
+            return int(result[1])
+
+    @staticmethod
+    async def _blpop(redis: Redis):
```
igboyes, line 39: Why make this a static method? I'm not sure why it can't just reference to `self.redis`.

Please add brief docstring as to why this is necessary and what it's doing. It's not immediately clear to me.

igboyes (changes requested):

virtool/tasks/client.py
ReeceHoffmann: Ahh yeah, that's a holdover from when I was experimenting with a decorator based solution for wrapping redis commands we might want to retry. (something like `@retryRedisCommand`) I could never make the api work in a way I thought was intuitive and reliable so I dropped it.

Has been updated to get redis from self.

igboyes (approved):

PR #2805, opened 2023-02-22
feat: update sqlalchemy to 2.0.4

SnickerDoodlePop:


igboyes (changes requested):
This does not start for me locally. It seems the problem is related to imports like:
```
from sqlalchemy.util import asyncio
```
This should probably be:
```
import asyncio
```

It is not caused by your changes.

PR #2810, opened 2023-02-23
feat: add modify hmm as an app permission to authorization model

ryanfang5:


igboyes:
We're going to use the `AdministratorRoutePolicy` instead, right?

ryanfang5:
> We're going to use the `AdministratorRoutePolicy` instead, right?

Yes, closing this since we are just checking for the Base role.

PR #2811, opened 2023-02-24
Use aggregation pipeline for SamplesData.get

SnickerDoodlePop:


virtool/samples/data.py
```diff
@@ -158,29 +158,32 @@ async def find(
...
-                AttachSubtractionTransform(self._db),
-                AttachUserTransform(self._db),
-            ],
-        )
+            document = await apply_transforms(
+                base_processor(document),
+                [
+                    ArtifactsAndReadsTransform(self._pg),
+                    AttachLabelsTransform(self._pg),
+                    AttachSubtractionTransform(self._db),
```
ReeceHoffmann, line 180: `AttachSubtractionTransform` should also be replaceable with an aggregation pipeline step as it is fetching data from mongo

virtool/samples/data.py
```diff
@@ -158,29 +158,32 @@ async def find(
...
-        if not document:
-            raise ResourceNotFoundError
+        async for document in self._db.samples.aggregate(
+            [
+                {"$match": {"_id": sample_id}},
+                *lookup_nested_user_by_id(local_field="user.id")
+            ]
+        ):
+            if not document:
+                raise ResourceNotFoundError
```
ReeceHoffmann, line 168: As written I don't think this can ever occur. If there is no document then it will never enter the loop

virtool/samples/api.py
```diff
@@ -156,6 +156,9 @@ async def get(
 
         try:
             sample = await get_data_from_req(self.request).samples.get(sample_id)
+            if sample is None:
+                raise ResourceNotFoundError()
```
ReeceHoffmann, line 160: Error should be raised by the DataLayer and caught here

virtool/samples/data.py
```diff
@@ -158,29 +158,32 @@ async def find(
         )
 
     async def get(self, sample_id: str) -> Sample:
-        document = await self._db.samples.find_one({"_id": sample_id})
-
-        if not document:
-            raise ResourceNotFoundError
+        async for document in self._db.samples.aggregate(
```
ReeceHoffmann, line 161: Since you know that you should only ever get one document I would be tempted to simply call `to_list` on the mongo cursor returned by `self._db.samples.aggregate` rather than using an async for loop. Then if the list is empty throw the  `ResoruceNotFound` error.

ReeceHoffmann (changes requested):
Part way there. There are a couple transforms that should be convertible into aggregation pipeline steps. Happy to have a quick chat if needed

virtool/samples/data.py
```diff
@@ -158,23 +159,28 @@ async def find(
...
 
-        document["caches"] = [
-            base_processor(cache)
-            async for cache in self._db.caches.find({"sample.id": sample_id})
-        ]
+        if len(documents) == 0:
+            raise ResourceNotFoundError
+        else:
+            document = documents[0]
+            del documents
```
ReeceHoffmann, line 177: No need for the `else`, if `len(documents) == 0` then an exception will be raised preventing execution of the rest of the function. This also lets you avoid pre-declaring `document` as `None`

virtool/subtractions/db.py
```diff
@@ -164,3 +164,22 @@ async def unlink_default_subtractions(
         {"$pull": {"subtractions": subtraction_id}},
         session=session,
     )
+
+
+def lookup_nested_subtractions(
+    local_field: str = "subtractions", set_as: str = "subtractions"
+) -> list[dict]:
+    """ """
```
ReeceHoffmann, line 172: Populate docstring

virtool/caches/db.py
```diff
@@ -93,3 +93,17 @@ async def remove(app: App, cache_id: str):
         await to_thread(rm, path, True)
     except FileNotFoundError:
         pass
+
+
+def lookup_caches(local_field: str = "_id", set_as: str = "caches") -> list[dict]:
+    """ """
```
ReeceHoffmann, line 99: Populate docstring

ReeceHoffmann (changes requested):
Looking good! just a couple tweaks and this should be good to go

virtool/samples/data.py
```diff
@@ -158,23 +159,27 @@ async def find(
...
+            [
+                {"$match": {"_id": sample_id}},
+                *lookup_nested_user_by_id(local_field="user.id"),
+                *lookup_nested_subtractions(local_field="subtractions"),
+                *lookup_caches(local_field="_id"),
+            ]
+        ).to_list(length=1)
 
-        if not document:
+        document = None
```
ReeceHoffmann, line 171: redundant `document` declaration

virtool/samples/data.py
```diff
@@ -158,23 +159,27 @@ async def find(
...
+
+        if (documents is None) or (len(documents) == 0):
```
ReeceHoffmann, line 173: Were you running into problems where `Documents` was `None`? I Think `to_list` always returns a list even if it is an empty list

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):
Looks good!

PR #2813, opened 2023-02-28
feat: add endpoints for managing administrators

ryanfang5:


ryanfang5:
I have left the updated Enum and models in administrators/data; I will update in core if you approve.

I passed in the user_id of the requester in `PATCH /administrators/:id` to prevent administrators from modifying their own role. Should I also prevent administrators from removing their own role? 

Also, I will also integrate the finer-defined Enums in another PR.

igboyes (approved):

PR #2815, opened 2023-02-28
feat: timeout long-running jobs

igboyes:
* Add the `timeout` status to jobs that have not received a ping in the past 10 minutes.
* Use `JobState` enum as much as possible.

### DIFF
--- a/virtool/jobs/data.py
+++ b/virtool/jobs/data.py
@@ -207,7 +207,7 @@ class JobsData:
             "args": job_args,
             "key": None,
             "rights": rights.as_dict(),
-            "state": "waiting",
+            "state": JobState.WAITING.value,
             "status": [compose_status(JobState.WAITING, None)],
             "user": {"id": user_id},
             "ping": None,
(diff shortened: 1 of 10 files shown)


virtool/jobs/data.py
```diff
@@ -472,3 +481,50 @@ async def force_delete(self):
...
+        Times out all jobs that have a ping field and haven't received a ping in 5
+        minutes.
+
+        """
+        now = arrow.utcnow()
+
+        async with self._db.create_session() as session:
+            async for document in self._db.jobs.find(
+                {
+                    "state": {"$in": ["preparing", "running"]},
```
ReeceHoffmann, line 501: Should waiting be on this list? If a job has been waiting for greater than 30 days, I would assume that redis state got wiped before the job could ever be ingested. Might cause problems if the `ping` field is defined on row creation though

... Also despite the syntactic verbosity, should probably use the Enum here as well

igboyes: > Should waiting be on this list? I would assume that redis state got wiped before the job could ever be ingested.

No. If Redis state somehow got wiped in production (should never happen), I would prefer to see obvious buggy behaviour in the app than hiding a serious issue behind timeout states.

If jobs somehow get stuck waiting, it would be better to just push their ids back into the list. Luckily this is a non-issue outside of flaky dev environments. I have never seen it in cloud.

> should probably use the Enum here as well

Yes

PR #2816, opened 2023-02-28
feat: add space id for existing mongodb resources

igboyes:


ReeceHoffmann (dismissed):
This looks good to me, I double checked against our list and it seems all of the normal resources are here. Only thing "missing" is `settings`, but I am guessing that omission is intentional based on previous discussion. 

Only other thing I noticed is that there is a bit of inconsistency in regards to where the default `space_id` is defined. For example for references the `create_document` function in `db.py` defines the default `space_id` internally while for creating a sample the default space is defined in`api.py`. Not critical and can be changed later as part of adding the new api routes

PR #2818, opened 2023-02-28
feat: update administrator role when updating administrator flag

ryanfang5:


virtool/users/data.py
```diff
@@ -150,6 +156,12 @@ async def update(self, user_id: str, data: UpdateUserRequest):
 
         try:
             update["administrator"] = data["administrator"]
+
+            if data["administrator"]:
+                await self._authorization_client.add(AdministratorRoleAssignment(user_id, AdministratorRole.FULL))
+            else:
+                await self._authorization_client.remove(AdministratorRoleAssignment(user_id, AdministratorRole.FULL))
```
igboyes, line 163:
```suggestion
            role_assignment = AdministratorRoleAssignment(user_id, AdministratorRole.FULL)
            
            if data["administrator"]:            
                await self._authorization_client.add(role_assignment)
            else:
                await self._authorization_client.remove(role_assignment)
```

virtool/users/data.py
```diff
@@ -150,6 +156,12 @@ async def update(self, user_id: str, data: UpdateUserRequest):
 
         try:
             update["administrator"] = data["administrator"]
```
igboyes, line 158: I think you should use `if "administrator" in update`.

The try-except encloses too much code and there is a risk a `KeyError` could bubble up from further down the stack and be ignored.

igboyes (changes requested):

PR #2819, opened 2023-02-28
feat: add tasks to deduplicate samples names

ReeceHoffmann:
Adds a new task to deduplicate sample names. Deduplicated samples are given incremented names based on `created_at` field (1). Designed to respect spaces, so samples with the same name in separate spaces will not be deduplicated.

(1) If a sample is encountered that already has the target name, the target name is skipped and the next one used. ex if you are deduplicating a list of samples with the names: `(sample,  sample, sample (2))` the final list of names in the same order will become `(sample,  sample (3), sample (2))`.


virtool/samples/db.py
```diff
@@ -642,3 +641,27 @@ async def get_sample(app, sample_id: str):
     document["paired"] = len(document["reads"]) == 2
 
     return document
+
+
+async def incrementing_sample_name_generator(mongo, base_name, space_id, session):
```
igboyes, line 646: Typing.

virtool/samples/db.py
```diff
@@ -642,3 +641,27 @@ async def get_sample(app, sample_id: str):
...
+    :return: the next available sample name
+    """
+    sample_number = 2
+    while True:
+        if not await mongo.samples.find_one(
+            {
+                "name": f"{base_name} ({sample_number})",
+                "space_id": space_id,
+            },
+            session=session,
```
igboyes, line 663: Faster and cheaper.
```suggestion
    sample_number = 2
    
    while True:
        if not await mongo.samples.count_documents(
            {
                "name": f"{base_name} ({sample_number})",
                "space_id": space_id,
            },
            limit=1,
            session=session,
```

virtool/samples/data.py
```diff
@@ -423,3 +424,55 @@ async def move_sample_files(self, progress_handler: AbstractProgressHandler):
...
+            ):
+                name_generator = incrementing_sample_name_generator(
+                    self._db,
+                    duplicate_name_documents["name"],
+                    duplicate_name_documents.get("space_id", None),
+                    session,
+                )
+
+                for sample_id in duplicate_name_documents["sample_ids"][1:]:
+                    new_name = await name_generator.asend(None)
```
igboyes, line 473: I think it would be better to make a simple class to do this instead of an async generator.

```
name_generator = NameGenerator(db, name, space_id, session)

name = name_generator.get()
```

The nice thing about generators is being able to use `for` to iterate through them. The iterable nature of the generator isn't being used here.

Thoughts?

igboyes (changes requested):
A few nitpicks.

virtool/samples/data.py
ReeceHoffmann: Hmm yeah your probably right. This method cuts down boilerplate a bit, but it doesn't really need to be a generator. Plus the `generator.asend` syntax is pretty awkward for getting a new name. Will convert to a class

igboyes (approved):

PR #2822, opened 2023-03-02
feat: add updated field to tasks table

ryanfang5:


ReeceHoffmann:
I believe that this change needs to be made using a DB migration via alembic. Also I need to rethink the issue as a whole, going to close for now.

PR #2824, opened 2023-03-03
fix: create task equivalent for refresh remotes

ChristineWC:


virtool/references/data.py
```diff
@@ -982,3 +985,16 @@ async def clean_all(self):
                         {"$pop": {"updates": -1}, "$set": {"updating": False}},
                         session=session,
                     )
+
+    async def update_references(self):
+        while True:
```
ReeceHoffmann, line 990: Scheduling + repetition will be handled by the task spawner. Task should run only once per call, otherwise it will block other tasks for running

virtool/startup.py
```diff
@@ -402,6 +401,11 @@ async def startup_tasks(app: Application):
     if app["config"].no_check_db:
         return logger.info("Skipping subtraction FASTA files checks")
 
+    if not app["config"].no_fetching:
+        await tasks_data.create(RefreshReferencesTask)
```
ReeceHoffmann, line 405: Should be a repeating task like `migrateFilesTask`

ReeceHoffmann (changes requested):
Looking pretty good, needs a few tweaks before it is ready to merge

ReeceHoffmann (changes requested):

virtool/references/data.py
```diff
@@ -982,3 +985,14 @@ async def clean_all(self):
                         {"$pop": {"updates": -1}, "$set": {"updating": False}},
                         session=session,
                     )
+
+    async def update_references(self):
```
ReeceHoffmann, line 989:
```suggestion
    async def fetch_and_update_reference_releases(self):
```
The name as written suggests that the function will update the references themselves, rather than simply updating the  list of known remote releases

virtool/references/tasks.py
```diff
@@ -207,3 +207,15 @@ def __init__(
 
     async def clean(self):
         await self.data.references.clean_all()
+
+
+class RefreshReferencesTask(BaseTask):
```
ReeceHoffmann, line 212:
```suggestion
class RefreshReferenceReleasesTask(BaseTask):
```

ReeceHoffmann (changes requested):
Looking super close! Just a few naming tweaks and I am happy to merge.

ReeceHoffmann (approved):

PR #2829, opened 2023-03-14
feat: add job field to new samples

SnickerDoodlePop:


igboyes (changes requested):
The job id needs to be passed to `data.jobs.create` further down in order for the sample and job to be linked.

PR #2831, opened 2023-03-15
feat: add administrators task to promote users

ryanfang5:


tests/administrators/test_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    static_time,
+    tmpdir,
+    fake2,
+):
+
+    await fake2.users.create()
+
+    user = await fake2.users.create()
+
+    async with mongo.create_session() as session:
```
igboyes, line 23: Don't need a session for this.

They are slower and data consistency doesn't matter so much for tests.

virtool/administrators/data.py
```diff
@@ -177,3 +177,12 @@ async def delete(self, user_id: str):
             return
 
         raise ResourceNotFoundError()
+
+    async def promote_administrators(self):
+        async with self._db.create_session() as session:
```
igboyes, line 182: Session doesn't add anything here. You are not modifying MongoDB, which is where as transaction is important.

virtool/administrators/data.py
```diff
@@ -177,3 +177,12 @@ async def delete(self, user_id: str):
...
+            async for user in self._db.users.find(
+                {"administrator": True}, session=session
+            ):
```
igboyes, line 185: You can use `users.distinct("_id", {"administrator": True})` here to be more efficient.

tests/administrators/test_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+
+    task = PromoteAdministratorsTask(1, data_layer, {}, get_temp_dir())
+
+    await task.run()
+
+    row = await get_row_by_id(pg, Task, 1)
+    assert row.complete is True
+
+    assert (
+        await data_layer.administrators._authorization_client.list_administrators()
```
igboyes, line 53: I think it's better to use `data_layer.administrators.find` here. That way we're testing the same interface that tasks is touching.

igboyes (changes requested):

virtool/startup.py
```diff
@@ -405,4 +407,5 @@ async def startup_tasks(app: Application):
     await tasks_data.create(CleanReferencesTask)
     await tasks_data.create(TimeoutJobsTask)
 
+    await tasks_data.create_periodically(PromoteAdministratorsTask, 600)
```
ReeceHoffmann, line 410: No need to add this here, this method of spawning tasks will be deprecated and removed in a later version.

igboyes (dismissed):

igboyes (approved):

PR #2833, opened 2023-03-15
feat: add job id to new samples

SnickerDoodlePop:


tests/samples/test_api.py
```diff
@@ -376,7 +376,11 @@ async def test(
         assert resp.headers["Location"] == snapshot
         assert await resp.json() == snapshot
 
-        assert await client.db.samples.find_one() == snapshot
+        test_sample = await client.db.samples.find_one()
+
+        test_sample.pop("job")
+
+        assert test_sample == snapshot
```
ReeceHoffmann, line 383: I would update the snapshot instead of mutating the returned data. Doing it like this means we don't get testing coverage for the job field.

virtool/samples/data.py
```diff
@@ -284,6 +289,7 @@ async def create(
             user_id,
             JobRights(),
             0,
+            job_id,
```
ReeceHoffmann, line 292: Gonna need @igboyes to chime in on this one, but I really am not sold on how this is currently structured as it is not atomic. It shouldn't happen much, but as written you could end up having a sample which reference a non-existent `job` if creating the job document fails. 

Unfortunately this is an artifact of how we designed the `dataLayer` since `dataLayer` methods are never passed a session. It could be fixed by moving most of the job creation logic into a `db` method, but that will essentially make the jobs `dataLayer.create` method redundant, as I suspect that method is only ever really called from other `dataLayers`. Also it means any `dataLayer` that spawns a job would have to get a copy of the `jobsClient`.

ReeceHoffmann (changes requested):

virtool/samples/data.py
igboyes: Agreed. Please guide @SnickerDoodlePop on fixing this.

Focus only on samples for this PR. He can fix the other occurences in another PR + issue.

ReeceHoffmann:
Converting this PR to a draft for now PR will be ready for review/merge once:

1. The PR required to transform a `job_id` into a `jobMinimal` for samples has been merged. (Due to pydantic errors that will result from pydantic being handed an incomplete `jobMinimal`)
3. The sample and related job are created in the same transaction.

PR #2836, opened 2023-03-21
feat: add endpoints for managing spaces

ryanfang5:


virtool/spaces/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    user_id: Union[str, int],
+    space_id: int,
+    enum_list: List[SpaceRoleType],
+):
+
+    for enum_type in enum_list:
+        for role in enum_type:
+            await authorization_client.remove(
+                UserRoleAssignment(user_id, space_id, enum_type(role))
+            )
```
igboyes, line 81: I think these needs a more elegant solution.

The `ENUM_LIST` constant naming is not good. Two types mushed together...

Can you not transiently define `ENUM_LIST` as a tuple when you are creating `AVAILABLE_ROLES`. Then, it looks like you can iterate through available roles in `remove_user_roles` instead of `ENUM_LIST`.

Alternatively, `ENUM_LIST` needs a rename.

Let me know if you want to discuss.

igboyes (changes requested):
Thanks for this. First round. Pretty close to done IMO.

PR #2838, opened 2023-03-23
Markus/vir 1749 create a lookup function for minimal

SnickerDoodlePop:
I manually tested this function to ensure the returned aggregation produces all the specified fields. I attempted to augment the existing automated testing, but the length of the rabbit hole I went down is far beyond what I am capable of. I may have to write a new unit test.


ReeceHoffmann (approved):
Looks good to me!

PR #2840, opened 2023-03-24
feat: create task to populate workflows field in sample data 

ChristineWC:


virtool/samples/data.py
```diff
@@ -474,3 +475,23 @@ async def deduplicate_sample_names(self):
...
+                    self._db.samples, "library_type", sample_id
+                )
+                analyses = await self._db.analyses.find(
+                    {"sample.id": sample_id}, ["ready", "workflow"]
+                ).to_list(None)
+
+                workflow_states = derive_workflow_state(analyses, library_type)
+                await self._db.samples.find_one_and_update(
+                    {"_id": sample_id}, {"$set": {"workflows": workflow_states}}
+                )
```
igboyes, line 497: 1. Iterate over the cursor. Using `to_list` loads all of the matching documents into memory, which could 
2. You are only using `_id` at line 483. Fetching the entire sample documents for one string is wasteful.
3. No need for `if`. The iteration will not happen on a zero-length list.
4. Use `get_one_field` is redundant. You already have this field in hand at line 480.
5. At line 495. Just use `update_one` if you are not using the returned updated document. This avoid unnecessary data transimission.

```suggestion
        async for sample in self._db.samples.find({"workflows": {"$exists": False}}, ["_id", "library_type"]):
            sample_id = sample["_id"]
            
            analyses = await self._db.analyses.find(
                {"sample.id": sample_id}, ["ready", "workflow"]
            ).to_list(None)

            workflow_states = derive_workflow_state(analyses, sample["library_type"])
            
            await self._db.samples.find_one_and_update(
                {"_id": sample_id}, {"$set": {"workflows": workflow_states}}
            )
```

virtool/samples/tasks.py
```diff
@@ -51,3 +51,18 @@ def __init__(self, task_id, data, context, temp_dir):
 
     async def deduplicate_sample_names(self):
         await self.data.samples.deduplicate_sample_names()
+
+
+class PopulateWorkflowsFieldTask(BaseTask):
+    """
+    Updates sample workflows field
```
igboyes, line 58: Inaccurate.

igboyes (changes requested):
In addition to changes requested in the comments, this needs testing.

You can get away with inserting reduced documents with only the fields required to test your behaviour. Make sure of the following:
* Only samples without the `workflows` field are modified.
* The `workflows` field is correctly populated for all samples even when no analyses exist for the sample.

Look to other task tests to see how these should be written. Ask for help if necessary.

virtool/samples/tasks.py
```diff
@@ -51,3 +51,18 @@ def __init__(self, task_id, data, context, temp_dir):
...
     async def deduplicate_sample_names(self):
         await self.data.samples.deduplicate_sample_names()
+
+
+class UpdateWorkflowsFieldsTask(BaseTask):
+    """
+    Updates workflows, nuvs, and pathoscoope fields for samples
+    """
+
+    name = "populate_workflows_field"
```
igboyes, line 61: The class name and `name` class attribute should match up.

Let's call it `UpdateSampleWorkflowsTask` and `update_sample_workflows`.

tests/samples/test_tasks.py
```diff
@@ -215,3 +216,66 @@ async def test_deduplicate_sample_names(
     await task.run()
 
     assert await mongo.samples.find().to_list(None) == snapshot
+
+
+@pytest.mark.parametrize("ready", [[True, False], [False, False]])
```
igboyes, line 221: Since `ready[1]` is `False` both times, can't this just be hardcoded in the function body.

Then `ready` can just be `[True, False]`.

tests/samples/test_tasks.py
```diff
@@ -215,3 +216,66 @@ async def test_deduplicate_sample_names(
...
+            "nuvs": False,
+            "pathoscope": True,
+            "workflows": {
+                "aodp": "incompatible",
+                "nuvs": "none",
+                "pathoscope": "none",
+            },
+        },
+    ]
+    ready_1, ready_2 = ready
```
igboyes, line 243: This can go since `ready_2` is always `False`.

tests/samples/test_tasks.py
```diff
@@ -215,3 +216,66 @@ async def test_deduplicate_sample_names(
...
+        {
+            "_id": "test1",
+            "sample": {"id": "test_id"},
+            "ready": ready_2,
+            "workflow": "nuvs",
+        },
+    ]
+
+    await mongo.samples.insert_many(samples, session=None)
+    await mongo.analyses.insert_many(analyses, session=None)
```
igboyes, line 260: I'd prefer you just put the `list` definitions right in the function call instead of defining `samples` and `analyses` separately.

Use whitespace between the `insert_many` calls to keep things legible.

igboyes (changes requested):

igboyes (approved):

PR #2842, opened 2023-03-27
fix: add spaces compatible url for samples

ChristineWC:


tests/samples/test_api.py
```diff
@@ -385,8 +385,15 @@ async def test(
 
         assert upload.reserved is True
 
+    @pytest.mark.parametrize("path", ["/samples", "/space/test_id/samples"])
```
igboyes, line 388: Space IDs are integers. The path prefix is 'spaces', not 'space'.
```suggestion
    @pytest.mark.parametrize("path", ["/samples", "/spaces/0/samples"])
```

igboyes (changes requested):

PR #2843, opened 2023-03-28
feat: attach job data to subtraction responses

SnickerDoodlePop:


tests/subtractions/__snapshots__/test_api.ambr
```diff
@@ -34,7 +34,7 @@
           'name': 'palm.fa.gz',
         },
       ],
-      'subtraction_id': 'abc123',
+      'subtraction_id': None,
```
ReeceHoffmann, line 37: Also seems wrong, nothing about the intended changes should result in the ID changing

virtool/subtractions/data.py
```diff
@@ -83,33 +85,84 @@ async def find(self, find: str, short: bool, ready: bool, query: MultiDictProxy)
...
+                        "page": [
+                            {"$count": "page"},
+                        ],
+                        "documents": [
+                            {
+                                "$match": query,
+                            },
+                            {"$sort": {"name": 1}},
+                            *lookup_nested_user_by_id(local_field="user.id"),
+                            *lookup_minimal_job_by_id(),
```
ReeceHoffmann, line 113: There are two general important problems here:

1. `$count` merely counts all documents currently in the list. I suggest reading up on what count does and refactoring this code, as most of these are not getting you what your looking for. https://www.mongodb.com/docs/manual/reference/operator/aggregation/count/
2. In the documents facet the it will just return the whole list of subtractions rather than applying any sort of pagination. Looks at the samples find function for an example of how this can be done.

virtool/subtractions/data.py
```diff
@@ -83,33 +85,84 @@ async def find(self, find: str, short: bool, ready: bool, query: MultiDictProxy)
...
-        documents, ready_count = await asyncio.gather(
-            apply_transforms(
-                data["documents"],
-                [AttachUserTransform(self._mongo, ignore_errors=True)],
-            ),
-            self._mongo.subtraction.count_documents({"ready": True}),
-        )
+        data = data[0]
+
+        documents = data["documents"]
```
ReeceHoffmann, line 152: Check behaviour when no subtraction matching the search query are found. Also if you are only returning the result there is no need to assign it to variables before returning

virtool/subtractions/data.py
```diff
@@ -83,33 +85,84 @@ async def find(self, find: str, short: bool, ready: bool, query: MultiDictProxy)
...
+
+        ready_count = await self._mongo.subtraction.count_documents({"ready": True})
 
         return SubtractionSearchResult(
             **{**data, "documents": documents, "ready_count": ready_count}
```
ReeceHoffmann, line 157: Isn't documents already a part of the data dict?

ReeceHoffmann (changes requested):
There are some smaller code style tweaks but there are a three big items that need to be addressed:

1. Read up on mongos `$count` aggregation step and reimplement the aggregation pipeline
2. Be careful when updating snapshots! The updated snapshots in this PR are wrong which is a big problems from a test reliability perspective
3. The tests should be updated to test that the user filter queries are respected and that jobs are being properly attached

virtool/subtractions/data.py
```diff
@@ -83,33 +83,65 @@ async def find(self, find: str, short: bool, ready: bool, query: MultiDictProxy)
...
+                        "found_count": [
+                            {"$count": "found_count"},
+                        ],
+                        "documents": [
+                            {
+                                "$match": query,
+                            },
+                            {"$sort": {"name": 1}},
+                            *lookup_nested_user_by_id(local_field="user.id"),
+                            *lookup_minimal_job_by_id(local_field="job.id"),
```
ReeceHoffmann, line 102: You need to paginated the data, best done in this step

virtool/subtractions/data.py
```diff
@@ -149,16 +183,14 @@ async def create(
...
             }
         )
 
-        subtraction = await self.get(document["_id"])
-
         await self.data.jobs.create(
             "create_subtraction",
             {
-                "subtraction_id": subtraction.id,
+                "subtraction_id": subtraction_id or default_subtraction_id,
```
ReeceHoffmann, line 193: Should use the returned `document["_id]`. As written a job could be created for a subtraction that may or may not even exist or be related to the one that is being created.

virtool/subtractions/data.py
```diff
@@ -83,33 +83,65 @@ async def find(self, find: str, short: bool, ready: bool, query: MultiDictProxy)
...
+                    "$facet": {
+                        "total_count": [
+                            {"$count": "total_count"},
+                        ],
+                        "found_count": [
+                            {"$count": "found_count"},
+                        ],
+                        "documents": [
+                            {
+                                "$match": query,
```
ReeceHoffmann, line 98: I believe that as written the aggregation pipeline will include documents which documents which have their `deleted` flag set to true. None of the facets should count samples which are `deleted`

virtool/subtractions/db.py
```diff
@@ -181,3 +183,31 @@ def lookup_nested_subtractions(
...
+    try:
+        per_page = int(query["per_page"])
+    except (KeyError, ValueError):
+        per_page = 25
+    finally:
+        data["per_page"] = per_page
+
+    data["page_count"] = int(math.ceil(data["found_count"] / per_page))
+
+    return data
```
ReeceHoffmann, line 213: Not a bad idea in principle to separate out complicated logic, but as written this function is a bit confused. There are core logic problems. Most problematically it appears to be both getting parameters needed to do pagination (`per_page`, `page`), while also calculating values like `page_count` which would be calculated after running the aggregation pipeline.

`per_page` and `page` are essential to have before the pipeline so that the correct documents can be returned. (per page indicates how many documents there are per page, while page is which page the user is on) The two parameters combined can be used to tell the pipeline which documents should be returned. (imagine the case where there are 100 subtractions, which ones should be returned if the user requests 25 documents per page and the third page)

ReeceHoffmann (changes requested):
Mix of larger and smaller problems. `find` method still needs some major changes as it does not currently accomplish what it needs to. `Get` method seems to be most of the way there.

Also looking through the dataLayer we also need to update subtractions `update` function to return a `jobMinimal` Since that endpoint also returns a `Subtraction`.  Any subtractions with a `job_id` will cause that endpoint to 500 as currently written

ReeceHoffmann (changes requested):
I think we are getting close here. Hoping we can get the last few problem resolved pretty quick and get this thing merged.

1. For now I would revert your changes to the `create` function of the subtraction dataLayer. The current changes unfortunately don't solve the atomicity problem and introduces the possibility of the jobs being assigned non-existent `subtraction_ids`
2. The testing for the `find` function is a bit lacking as it doesn't check if the various parameters are used/stable when requested. (`page`, `per_page`,  `find`, `ready` and `short`)

tests/subtractions/__snapshots__/test_data.ambr
```diff
@@ -16,6 +16,9 @@
       't': 0.319,
     },
     'has_file': True,
+    'job': <class 'dict'> {
+      'id': 'fb085f7f',
+    },
```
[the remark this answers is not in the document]
ReeceHoffmann: Markus pointed out that this snapshot is checking the raw returned data from a mongo query. The same test also checks the values returned from the API request so the lookup is covered.

igboyes (approved):

PR #2844, opened 2023-03-31
chore: return minimal job data in sample responses

ryanfang5:


ryanfang5:
Returned AnalysisSearchResult for GET in ("/samples/{sample_id}/analyses") - not sure if AnalysisMinimal should be returned for POST. For now I have only attached the JobMinimal data to the document.

tests/samples/test_api.py
```diff
@@ -834,14 +834,26 @@ async def test_find_analyses(
...
+
+    await client.db.references.insert_many(
+        [
+            {"_id": "foo", "data_type": "genome", "name": "Foo"},
+            {"_id": "baz", "data_type": "genome", "name": "Baz"},
+        ],
+        session=None,
+    )
+
     await client.db.analyses.insert_many(
```
igboyes, line 849: Need a test for when the `job.id` field is set in the analysis, but no corresponding job is found in the database. This should result in the Pydantic field being set `None`.

virtool/samples/data.py
```diff
@@ -159,6 +160,210 @@ async def find(
...
+                            "job.id": True,
+                            "job.archived": True,
+                            "ready": True,
+                            "reference": True,
+                            "sample": True,
+                            "subtractions": True,
+                            "updated_at": True,
+                            "user._id": True,
+                            "user.handle": True,
+                            "user.administrator": True,
```
igboyes, line 338: Is there no way to simplify this. Can you not just do `"job": True` to cover all job fields?

igboyes (changes requested):

virtool/samples/data.py
ryanfang5: I originally wanted to remove the "key" field from being returned in jobs, but I realised it was easier to do with an $unset. I also did this in analyses/data.

igboyes (approved):

PR #2847, opened 2023-04-04
chore: use lookup functions in aggregations

ryanfang5:


ReeceHoffmann:
I haven't looked through this PR, but if you have time can you check for flakey tests introduced by this? The test from your latest commit looks like it it could be an example of flake. The mongo lookup makes no promises about return order to we should sort inside the lookup or elsewhere in the test since the snapshots check the order and contents of a list.

igboyes (approved):

PR #2856, opened 2023-04-12
refactor admin view

ReeceHoffmann:
Main changes:
- refactor admin API
- add new endpoint for listing and updating users
- add fully featured aggregation pagination function
- ensure more complete test coverage of admin API
- fix flaky tests introduced by the usage of lookups
- remove vestigal session updating code


igboyes (changes requested):

tests/administrators/test_data.py
```diff
@@ -0,0 +1,169 @@
...
+
+    updated_user = await mongo.users.find_one()
+    assert updated_user["administrator"] == (role == AdministratorRole.FULL)
+
+    expected_list = [(user_1.id, role)] if role is not None else []
+
+    assert (
+        await data_layer.administrators._authorization_client.list_administrators()
+        == expected_list
+    )
```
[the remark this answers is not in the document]
ReeceHoffmann: I can switch it snapshots, but I find the DX of diffing more than one per test suboptimal. The expected result of this test was sufficiently simple that to me it made more sense to predict the expected result rather than use snapshots. Main upside I can see of snapshots is that it would flag a change in what the role is, which could be good if that change is borked, otherwise it is annoying.

Happy to change it though if you want snapshots use throughout

virtool/administrators/api.py
```diff
@@ -1,130 +1,156 @@
...
-            201: Successful operation
-            400: User not found
+            200: Successful operation
         """
 
-        try:
-            administrator = await get_data_from_req(self.request).administrators.create(
-                data
+        return json_response(
+            await get_data_from_req(self.request).administrators.find_users(
```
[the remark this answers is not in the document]
ReeceHoffmann: Yeah I think your right that it could be parsed via aiohttp. The short version is that when updating `paginate_aggregate` to do the aggregation I was trying to keep the api as consistent with other `paginate` functions as possible. `paginate` expects the `url_query` while `paginate2` expects `page` and `per_page` as `ints`. `paginate` is in use `paginate2` is not.

If you prefer the `paginate2` api I can shift it over to expecting page and per page be handed to it as `ints`. (which imo makes more sense with our current separation of concerns)

virtool/administrators/data.py
```diff
@@ -1,186 +1,258 @@
...
+        if administrator is not None:
+            operator = "$in" if administrator else "$nin"
+            administrator_query = {"_id": {operator: list(admin_dict.keys())}}
+
+        term_query = compose_regex_query(term, ["handle"]) if term else {}
+
+        client_query = {
+            **administrator_query,
+            **term_query,
+        }
```
igboyes, line 87: Converting to `dict`, then calling `keys()` is indirect and not particularly efficient. It is also not necessary to access OpenFGA if `administrator is None` .

Not sure if the unary will work for `dict` keys, but my guess is that it will. Up to you if you want to do it.

```suggestion
        query = {}
        
        if administrator is not None:
            administrator_ids = [a[0] for a in await self._authorization_client.list_administrators()]            
            query["_id"]: {"$in" if administrator else "$nin": administrator_ids}
            
        if term:
            query.update(compose_regex_query(term, ["handle"]))
```

virtool/authorization/client.py
```diff
@@ -83,16 +83,21 @@ async def get_space_roles(self, space_id: int) -> List[str]:
             [relation_tuple.key.relation for relation_tuple in response.tuples]
         )
 
-    async def get_administrator(self, user_id: str) -> Tuple[str, AdministratorRole]:
+    async def get_administrator(
+        self, user_id: str
+    ) -> Tuple[str, Optional[AdministratorRole]]:
```
igboyes, line 88: Docstring.

virtool/api/utils.py
```diff
@@ -180,50 +183,77 @@ async def paginate_aggregate(
...
+                        {"$count": "found_count"},
+                    ],
+                    "data": [
+                        {
+                            "$match": client_query,
+                        },
+                        {"$sort": sort},
+                        {"$skip": skip_count},
+                        {"$limit": per_page},
+                        *transforms,
```
igboyes, line 233: These aren't really transforms are they. They are functions that return lookups for MongoDB aggregations. I think a new name needs to be found.

virtool/groups/db.py
```diff
@@ -99,53 +99,50 @@ async def fetch_group_users(db, group_id: str) -> List[UserNested]:
     ]
 
 
-def lookup_minimal_groups(
-    local_field: str = "groups", set_as: str = "groups"
-) -> list[dict]:
+def lookup_group_minimal_by_id(
+    local_field: str = "group", set_as: str = "group"
+) -> List[dict]:
```
igboyes, line 104: Is there somewhere where `local_field` and `set_as` arent the default values here?

virtool/groups/db.py
```diff
@@ -99,53 +99,50 @@ async def fetch_group_users(db, group_id: str) -> List[UserNested]:
...
+        {"$set": {set_as: {"$first": f"${set_as}"}}},
     ]
 
 
-def lookup_minimal_group_by_id(
-    local_field: str = "group", set_as: str = "group"
-) -> list[dict]:
+def lookup_groups_minimal_by_id(
+    local_field: str = "groups",
+    set_as: str = "groups",
```
igboyes, line 128: Is there somewhere where `local_field` and `set_as` arent the default values here?

tests/administrators/test_data.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+@pytest.mark.parametrize("administrator", [True, False])
+async def test_get_user(
+    spawn_client, authorization_client, fake2, snapshot, data_layer, administrator
+):
+    group = await fake2.groups.create()
+    user = await fake2.users.create(
+        handle="test_user", groups=[group], primary_group=group
```
igboyes, line 57: Why is it necessary to hard code `handle` here?

igboyes (changes requested):
Either or:

1. Data layer attribute name should be `data.admin` and method is `update_user()`
2. Data layer attribute name is administrators and method is `update()`.

I suggest **2** to limit the size of data layer "domains".

virtool/administrators/data.py
ReeceHoffmann: Gonna split this into separate items for clarity:

 1. Accessing openFGA is required regardless of filtering, since the endpoint returns a user object with `administrator_role` attached. Unfortunately that info is not stored in mongo. If there is a large number of admins there may be an argument for aggregating the users to be sent, then requesting specific tuples for the users. Given the small number of admins and the relatively large number of users my suspicion is that it will be faster as is though.
 
 2. Your right the dict is indirect. That said, I converted it into a dict since the dict is O(c) for doing lookups when a attaching admin roles to the returned users. Though given how short the admin list is it might be possible that it would be faster to eat the O(n) of searching through the whole list when attaching the ids. I need to retain the `role:id` association from the tuple, so the list soln would have to be a list of tuples or similar.

virtool/api/utils.py
ReeceHoffmann: Yeah I will try to think of something else. I was mostly trying to avoid `lookups` as technically these steps could do anything to the data

virtool/groups/db.py
ReeceHoffmann: No, not that I know of, could technically be changed. Written as is to keep consistency between lookup functions

tests/administrators/test_data.py
ReeceHoffmann: Good catch, it isn't

virtool/administrators/data.py
igboyes: 1. Right. I missed this in the GitHub diff view.
2. I see how the `dict` makes sense now. Keep it as is. Please come up with a better name than `admin_dict`. Names with types in them are a code smell (even though I have git blame on some of them...).

virtool/administrators/api.py
```diff
@@ -159,11 +161,11 @@ async def check_can_edit_user(
...
 class AdminRoleView(PydanticView):
     @policy(AdministratorRoutePolicy(AdministratorRole.FULL))
-    async def post(
+    async def put(
         self, user_id: str, /, data: UpdateAdministratorRoleRequest
-    ) -> Union[r201[UserResponse], r404]:
+    ) -> Union[r200[UserResponse], r404]:
         """
-        Change a users administrator role
+        Change a users administrator role.
```
igboyes: Set administrator role.

Updates the user's administrator role.

virtool/administrators/api.py
```diff
@@ -83,17 +85,17 @@ class AdminUserView(PydanticView):
     @policy(AdministratorRoutePolicy(AdministratorRole.USERS))
     async def get(self, user_id: str, /) -> Union[r200[UserResponse], r404]:
         """
-        Get the complete representation of a user
+        Get a user.
+
+        Returns a complete representation of a user
```
igboyes, line 92:
```suggestion
        Fetches the details of a user
```

igboyes (changes requested):

PR #2857, opened 2023-04-12
refactor: rename db connection function imports

SnickerDoodlePop:


virtool/pg/utils.py
```diff
@@ -20,7 +20,7 @@ def to_list(cls):
         return [e.value for e in cls]
 
 
-async def connect(postgres_connection_string: str) -> AsyncEngine:
+async def connect_pg_utils(postgres_connection_string: str) -> AsyncEngine:
```
igboyes, line 23:
```suggestion
async def connect_pg(postgres_connection_string: str) -> AsyncEngine:
```

igboyes (changes requested):

igboyes (changes requested):
Please double-check my requested change to what you committed.

igboyes (approved):

PR #2862, opened 2023-04-14
Markus/vir 1744 set job field on new samples

SnickerDoodlePop:


ReeceHoffmann (changes requested):
Looks like it solves the atomicity problem! Need to tweak how `create_job` is getting access to the data sources it needs though. Looking close

virtool/jobs/data.py
```diff
@@ -201,27 +202,20 @@ async def create(
...
-
-        if job_id:
-            document["_id"] = job_id
-
-        document = await self._db.jobs.insert_one(document)
-        await self._client.enqueue(workflow, document["_id"])
 
-        return await fetch_complete_job(self._db, document)
+        job_mongo = copy.copy(self._db)
+        job_client = copy.copy(self._client)
```
igboyes, line 207: Why?

igboyes (commented):
Looking good other than one oddity.

virtool/jobs/data.py
SnickerDoodlePop: oops, that shouldn't be there

PR #2863, opened 2023-04-17
fix!: create application data directories as needed

ChristineWC:


igboyes (changes requested):
1. Creating the directories shouldn't be done in the `join` functions. Those are specifically for returning a `Path` object and some cases where they are called, you do not want a directory being created as a side effect. Make the directories in the scope that is calling the `join` function. The way you did it for uploads is correct.
2. The function `startup_paths` can be completely removed. The `logs/jobs` path is no longer used.
3. Making a directory is a blocking operation because the whole application waits while the Python runtime interacts with the filesystem. This can be avoided by spinning `mkdir` off into a thread. You can do this by calling `mkdir` in an `async` function using  [`asyncio.to_thread`](https://docs.python.org/3.10/library/asyncio-task.html#asyncio.to_thread).

virtool/hmm/api.py
```diff
@@ -191,6 +192,9 @@ async def get_hmm_annotations(req):
 
     Fetches a compressed json file containing the database documents for all HMMs.
     """
+    hmm_path = req.app["config"].data_path / "hmm"
```
igboyes, line 195: Please use `get_config_from_app` here.

It maintains typing automatically since the function has a return type annotation. This is not possible with dictionary style access like `["config"]`.
```

File: virtool/hmm/api.py [hunk 2]
```diff
@@ -192,7 +193,8 @@ async def get_hmm_annotations(req):
 
     Fetches a compressed json file containing the database documents for all HMMs.
     """
-    hmm_path = req.app["config"].data_path / "hmm"
+    config = get_config_from_req(req)
+    hmm_path = config.data_path / "hmm"
```
igboyes, line 197:
```suggestion
    hmm_path = get_config_from_req(req).data_path / "hmm"
```

virtool/subtractions/api.py
```diff
@@ -171,7 +172,9 @@ async def upload(req):
     subtraction_id = req.match_info["subtraction_id"]
     filename = req.match_info["filename"]
 
-    subtraction_path = req.app["config"].data_path / "subtractions" / subtraction_id
+    config = get_config_from_req(req)
+
+    subtraction_path = config.data_path / "subtractions" / subtraction_id
```
igboyes, line 177:
```suggestion
    subtraction_path = get_config_from_req(req).data_path / "subtractions" / subtraction_id
```

virtool/samples/api.py
```diff
@@ -707,16 +704,15 @@ async def upload_cache_reads(req):
     name = req.match_info["filename"]
     sample_id = req.match_info["sample_id"]
     key = req.match_info["key"]
+    config = get_config_from_req(req)
 
     if name not in ["reads_1.fq.gz", "reads_2.fq.gz"]:
         raise HTTPBadRequest(text="File name is not an accepted reads file")
 
-    caches_path = req.app["config"].data_path / "caches"
+    caches_path = join_cache_path(config, key)
```
igboyes, line 710: Naming is not great here.

For `caches_path`; confusing as `join_cache_path` returns the path to a single, specific cache (singular). This makes more sense as `cache_path`.

virtool/samples/api.py
```diff
@@ -707,16 +704,15 @@ async def upload_cache_reads(req):
...
     await asyncio.to_thread(caches_path.mkdir, parents=True, exist_ok=True)
 
-    cache_path = (
-        Path(virtool.caches.utils.join_cache_path(req.app["config"], key)) / name
-    )
+    cache_path = caches_path / name
```
igboyes: Naming is not great here either.

This path is for a file (`name` is the filename) in the cache not the cache itself as suggested by the name `cache_path`. Change this to something like `cache_file_path`.

igboyes (changes requested):

PR #2864, opened 2023-04-17
chore: remove sigint and sigterm event handlers

SnickerDoodlePop:

PR #2872, opened 2023-04-27
fix: require task name in task spawner cli

SnickerDoodlePop:


virtool/tasks/spawner.py
```diff
@@ -81,6 +81,9 @@ async def shutdown_spawner_app(app):
 
 
 def get_task_from_name(task_name: str) -> Type[BaseTask]:
+
+    logger.info(task_name)
+
     matching_task = [cls for cls in BaseTask.__subclasses__() if cls.name == task_name]
```
ReeceHoffmann, line 87: Pretty sure the task name is already logged elsewhere

virtool/startup.py
```diff
@@ -249,7 +255,7 @@ async def startup_b2c(app: App):
     """
     config = get_config_from_app(app)
 
-    if not config.use_b2c:
+    if not (config.use_b2c if hasattr(config, "use_b2c") else False):
```
ReeceHoffmann, line 258: Was accessing it directly causing problems? This attribute should always be set for any config which calls this startup step

virtool/startup.py
```diff
@@ -198,7 +201,10 @@ async def startup_databases(app: App):
 
 async def startup_routes(app: App):
     logger.debug("Setting up routes")
-    setup_routes(app, dev=app["config"].dev)
+
+    config = app["config"]
+
+    setup_routes(app, dev=config.dev if hasattr(config, "dev") else False)
```
ReeceHoffmann, line 207: Why the change? dev attribute should always be set if the startup path is getting called

virtool/config/cli.py
```diff
@@ -144,8 +145,9 @@ def start_task_runner(**kwargs):
 @tasks.command("spawn")
 @postgres_connection_string_option
 @redis_connection_string_option
-@click.option("--task-name", help="Name of the task to spawn", type=str)
-def spawn_task(task_name: str, **kwargs):
+@click.argument("task_name", type=str)
+def spawn_task(task_name: str = None, **kwargs):
```
ReeceHoffmann, line 149: I don't think a default value for `task_name` makes sense here

virtool/startup.py
```diff
@@ -55,7 +55,10 @@ def get_scheduler_from_app(app: App) -> aiojobs.Scheduler:
 
 
 async def startup_check_db(app: App):
-    if app["config"].no_check_db:
+
+    config = app["config"]
+
+    if (config.no_check_db if hasattr(config, "no_check_db") else False):
```
ReeceHoffmann, line 61: `no_check_db` should always be set for any startup process that calls this function

ReeceHoffmann (changes requested):

virtool/startup.py
SnickerDoodlePop: Some processes in tilt complained `app["config"] as no attribute "dev"`

virtool/startup.py
SnickerDoodlePop: Some processes in tilt complained app["config"] as no attribute "use_b2c"

virtool/startup.py
SnickerDoodlePop: Some processes in tilt complained `app["config"] has no attribute "no_check_db"`

virtool/config/cls.py
```diff
@@ -45,17 +45,26 @@ class TaskRunnerConfig:
...
     postgres_connection_string: str
     redis_connection_string: str
     sentry_dsn: str
+    b2c_client_id: str
+    b2c_client_secret: str
+    b2c_tenant: str
+    b2c_user_flow: str
+    postgres_connection_string: str
+    redis_connection_string: str
+    use_b2c: bool
```
ReeceHoffmann, line 67: I don't think the task runner needs any of this

virtool/config/cli.py
```diff
@@ -132,6 +133,9 @@ def tasks():
 @postgres_connection_string_option
 @redis_connection_string_option
 @sentry_dsn_option
+@no_check_db_option
+@b2c_options
+@dev_option
```
ReeceHoffmann, line 138: Don't think the task runner needs any of this

virtool/config/cli.py
```diff
@@ -144,8 +148,9 @@ def start_task_runner(**kwargs):
 @tasks.command("spawn")
 @postgres_connection_string_option
 @redis_connection_string_option
-@click.option("--task-name", help="Name of the task to spawn", type=str)
+@click.argument("task_name", type=str)
```
ReeceHoffmann, line 151: Why the conversion to argument?

ReeceHoffmann (changes requested):
Looks like there was a major bug introduced in #2864 which results in `start_task_runner` spawning an API instance rather than a task runner. Resolve this bug and I suspect that the new options added to the task runner will not be required

ReeceHoffmann (changes requested):
Just a couple small cleanup things. Otherwise looks good

ReeceHoffmann (approved):

PR #2879, opened 2023-05-11
fix: wait for spawned tasks before shutdown

SnickerDoodlePop:


ReeceHoffmann (changes requested):
As written this code needs a rewrite before it can be merged. There are two major points of concern:

1. As written I believe the tasks will still be immediately killed upon the kill signal being sent. The primary goal of this issue is to attempt to allow the virtool initiated tasks to conclude within a reasonable time frame before exiting to reduce the chances of unnecessary task failures

2. There seems to be some confusion regarding asyncio tasks and virtool tasks. In virtool tasks are essentially short term background jobs. As part of the implementation we use the asyncio tasks API. (ie only some asyncio tasks are these background jobs)

I would suggest accomplishing this by modifying the task runner itself. I would have the main task runner event loop wrapped in a try except block that catches asyncio `CancelledError`. If the error is raised then await the await the task with a time limit and exit after the task is done or the time limit is up. You may also need to shield the underlying task to prevent cancellation when the cancellation error is first propagated.

Task cancellation: https://docs.python.org/3/library/asyncio-task.html#task-cancellation

Time limited await: https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for

SnickerDoodlePop:
DeepSource is complaining about the use of an f-string that was implemented in a previous pr: I am ignoring it.

virtool/tasks/runner.py
```diff
@@ -46,11 +58,18 @@ async def run_task(self, task_id: int):
...
 
-        for cls in BaseTask.__subclasses__():
-            if task.type == cls.name:
-                current_task = await cls.from_task_id(self.app["data"], task.id)
-                await asyncio.get_event_loop().create_task(current_task.run())
+        except asyncio.CancelledError as task_cancelled:
+            raise task_cancelled
+
+        except Exception as err:
+            raise err
```
ReeceHoffmann, line 75: This `try except` block seems redundant to me

virtool/tasks/main.py
```diff
@@ -86,4 +88,8 @@ async def create_task_runner_app(config: TaskRunnerConfig):
 
 def run_task_runner(config: TaskRunnerConfig):
     app = create_task_runner_app(config)
-    aiohttp.web.run_app(app=app, host=config.host, port=config.port)
+
+    with suppress(asyncio.CancelledError):
```
ReeceHoffmann, line 92: I don't think suppressing errors here should be required. I suspect it is just hiding another error that should be dealt with more properly. Also I would be hesitant to suppress asyncio cancelled errors without good reason, as it can prevent the task from dying like it should

virtool/tasks/runner.py
```diff
@@ -23,20 +23,32 @@ def __init__(
...
+
+                await asyncio.gather(*tasks)
 
                 logging.info("Finished task: %s", task_id)
 
         except asyncio.CancelledError:
-            logging.info("Stopped task runner")
+            logging.info("Stopped task runner; awaiting task completion")
+
+            await asyncio.wait_for(asyncio.gather(*tasks), timeout=20)
```
ReeceHoffmann, line 43: My bad on this one, but lets up it to 10 mins. Very long vs the default, but some of the tasks are long running and I want to give them a chance to complete before the container getting killed. Easy enough to increase grace period in Kube.

ReeceHoffmann (changes requested):
This is definitely closer to what we are looking for. That said I am still don't think this will actually prevent premature task cancellation. I would suggest reading up on how asyncio task cancellation works and manually test that tasks are attempting to complete within the given window.

Primarily we need to verify that both:
1. Tasks that can complete within the grace period are completing
2. The process exits immediately after the task is done even if the grace period is not over

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
 
                 await self.run_task(task_id)
 
+                await asyncio.shield(self.current_task)
+
                 logging.info("Finished task: %s", task_id)
 
         except asyncio.CancelledError:
-            logging.info("Stopped task runner")
+            if task_id is not None:
```
igboyes, line 39:
```suggestion
            if self.current_task:
```

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
         self._data = data
         self._tasks_client = tasks_client
         self.app = app
```
igboyes, line 21: I think this unneeded. Can you remove if possible?

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
+        self.current_task = None
```
igboyes, line 22: Add type annotation.
```suggestion
        self._current_task = None
```

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
 
     async def run(self):
-        logging.info("Started task runner")
+        task_id = None
```
igboyes, line 25: I don't think this is necessary.

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
         try:
             while True:
                 logging.info("Waiting for next task")
-
                 task_id = await self._tasks_client.pop()
 
                 await self.run_task(task_id)
 
+                await asyncio.shield(self.current_task)
+
```
igboyes, line 35:
```suggestion
                await asyncio.shield(self.current_task)
                
                self._current_task = None

```

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
                 logging.info("Finished task: %s", task_id)
 
         except asyncio.CancelledError:
-            logging.info("Stopped task runner")
+            if task_id is not None:
+                logging.info("Recieved stop signal; awaiting task completion")
```
igboyes, line 40: This should maybe be two logging calls.

1. Received stop signal where _Stopped task runner_ used to be. It is logged whether or not a task is active.
2. _Waiting for running task to finish/complete_ if there is a current task active.

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
+
+                try:
+                    await asyncio.wait_for(asyncio.shield(self.current_task), 600)
+
+                    logging.info("Finished task: %s", task_id)
+
+                except asyncio.TimeoutError:
+                    logging.info("Task %s timed out", task_id)
```
igboyes, line 48: This should never happen and is considered an error state. Better wording may be _Timed out while waiting for task to complete_.

```suggestion
                except asyncio.TimeoutError:
                    logging.critical("Task %s timed out", task_id)
```

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
+
+                except asyncio.CancelledError:
+                    logging.info("Task %s was cancelled", task_id)
```
igboyes, line 51: I don't think this is very useful to the person reading the logs. What does "cancelled" mean to someone who is just using the CLI and doesn't know anything about how it's implemented?

What acutally happened from their perspective? Another kill signal was received after the first one?

virtool/tasks/runner.py
```diff
@@ -19,24 +19,42 @@ def __init__(
...
 
         except asyncio.CancelledError:
-            logging.info("Stopped task runner")
+            if task_id is not None:
+                logging.info("Recieved stop signal; awaiting task completion")
+
+                try:
+                    await asyncio.wait_for(asyncio.shield(self.current_task), 600)
+
+                    logging.info("Finished task: %s", task_id)
```
igboyes, line 45: You use the wording _complete_ above. Be consistent and try to follow from the wording at line 40.

virtool/tasks/runner.py
```diff
@@ -53,4 +71,9 @@ async def run_task(self, task_id: int):
         for cls in BaseTask.__subclasses__():
             if task.type == cls.name:
                 current_task = await cls.from_task_id(self.app["data"], task.id)
-                await asyncio.get_event_loop().create_task(current_task.run())
+
+                self.current_task = asyncio.get_event_loop().create_task(
```
igboyes, line 75: Please check if this works. I don't think it is necessary to get the event loop.
```suggestion
                self.current_task = asyncio.get_event_loop().create_task(
```

igboyes (changes requested):

virtool/tasks/runner.py
```diff
@@ -48,9 +66,14 @@ async def run_task(self, task_id: int):
...
         for cls in BaseTask.__subclasses__():
             if task.type == cls.name:
                 current_task = await cls.from_task_id(self.app["data"], task.id)
-                await asyncio.get_event_loop().create_task(current_task.run())
+
+                self._current_task = asyncio.create_task(
+                    current_task.run()
+                )
+
+                await asyncio.shield(self._current_task)
```
ReeceHoffmann, line 79: This shield is required since we need to protect the underlying task from being canceled when the asycio canceled error is propagated

virtool/tasks/runner.py
```diff
@@ -14,29 +13,48 @@
...
 
         except asyncio.CancelledError:
-            logging.info("Stopped task runner")
+            if self._current_task:
+                logging.info("Recieved stop signal")
+
+                try:
+                    logging.info("Waiting for task to finish: %s", task_id)
+
+                    await asyncio.wait_for(asyncio.shield(self._current_task), 600)
```
ReeceHoffmann, line 43: This shield concerns me. My impression is that we should only see one cancellation error. Ideally at this point you have already caught any cancelation errors that are being propagated. If your still seeing cancellation errors at this point then we need to think this through, as an error raised at this point would immediately bail from the runner but allow the task to keep running. If that happens good chance aiohttp will shutdown the database connections before the task is done.

virtool/tasks/runner.py
```diff
@@ -14,29 +13,48 @@
...
-
         try:
             while True:
                 logging.info("Waiting for next task")
-
                 task_id = await self._tasks_client.pop()
 
                 await self.run_task(task_id)
 
+                await asyncio.shield(self._current_task)
```
ReeceHoffmann, line 30: I think this shield is redundant as the underlying is already shielding the task we want to let keep running

ReeceHoffmann (commented):
Added comments next to each shield with my thoughts. I am working under the assumption that calling `task.cancel` only propagates one cancelled error under normal circumstances. Multiple canceled error may be getting raised since the whole scheduler is getting closed right now. If that's the case I would add cancel the task runner separately from the scheduler.

ReeceHoffmann (dismissed):
Looks good!

virtool/tasks/runner.py
```diff
@@ -12,45 +11,90 @@
 from sentry_sdk import capture_exception
 
 
+async def shutdown_task_runner(app: Application):
+    """
+    Raise an `asyncio.CancelledError` in the running task runner.
+
+    :param app: Current running application.
+    """
+    await app["task_runner"].close(timeout=float("inf"))
```
igboyes, line 20: 1. Calls to `close()` shouldn't be necessary as the entire `aiojobs` scheduler is closed on application shutdown.

   Is this not the case?

2. All (necessary) shutdown functions need to go in `virtool.shutdown`.

virtool/tasks/runner.py
```diff
@@ -12,45 +11,90 @@
...
+
+
+SHUTDOWN_TIMEOUT: int = 600
```
igboyes, line 23: Needs docstring. Constants should be defined near the top of the module after imports.
```suggestion
SHUTDOWN_TIMEOUT: int = 600
"""Something here"""
```

virtool/tasks/runner.py
```diff
@@ -12,45 +11,90 @@
...
+            try:
+                logging.info("Waiting for task to finish: %s", self._current_task_id)
+
+                await asyncio.wait_for(self._current_task, SHUTDOWN_TIMEOUT)
+
+                logging.info("Finished task: %s", self._current_task_id)
+
+            except asyncio.TimeoutError:
+                logging.critical(
+                    "Task Runner timed out while waiting for task to finish: %s",
```
igboyes, line 96:
```suggestion
                    "Timed out while waiting for task to finish: %s",
```

virtool/tasks/runner.py
```diff
@@ -12,45 +11,90 @@
...
+
+                self._current_task_id = None
+
+    async def _shutdown(self):
+        """
+        Gracefully shutdown the task runner.
+
+        Current task is awaited for SHUTDOWN_TIMEOUT seconds;
+        if it exceeds this time or recieves a second kill signal,
+        the task is forcefully shut down.
```
igboyes, line 82: Max line width is 90. Your docstrings are deviating from this.

virtool/tasks/runner.py
```diff
@@ -12,45 +11,90 @@
...
+        """
+        if self._current_task:
+            logging.info("Received stop signal")
```
igboyes, line 85:
```suggestion
        
        logging.info("Received stop signal")
        
        if self._current_task:
```

virtool/tasks/runner.py
```diff
@@ -12,45 +11,90 @@
...
+    async def _run_task(self):
         """
-        Run task with given `task_id`.
-
-        :param task_id: ID of the task
-
+        Run task given by `self._current_task_id`.
         """
-        task: Task = await get_row_by_id(self.app["pg"], Task, task_id)
+        task: Task = await get_row_by_id(self.app["pg"], Task, self._current_task_id)
```
igboyes, line 56: Do this with data layer instead of direct access to PG.

virtool/tasks/runner.py
```diff
@@ -12,45 +11,90 @@
...
+
+SHUTDOWN_TIMEOUT: int = 600
+
+
 class TaskRunner:
-    def __init__(
-        self, data: DataLayer, tasks_client: AbstractTasksClient, app: Application
-    ):
-        self._data = data
+    def __init__(self, tasks_client: AbstractTasksClient, app: Application):
```
igboyes, line 27: Undo this change. Access to task data can be done entirely through data layer.

igboyes (changes requested):

virtool/tasks/runner.py
ReeceHoffmann: Directly using `scheduler.close()` was resulting in multiple `asycio.cancelledError`s being propagated in short succession. I think this can be fixed by increasing the timeout from the default `0.1s` to a larger value. This will make it apply to all jobs, but that's probably not a huge deal. Inside the component then it leaves the options of either assuming the second `cancelledError` is the result of the timeout or putting a second timeout inside the component that is slightly shorter to reduce the chances of a race condition. I wasn't a fan of either option, so suggested doing a separate shutdown. I am not stuck on either soln, but thought I would throw in some extra context

igboyes: Do you know why more than one `CancelledError` is raised inside the runner?

Looking at the source for `aiojobs`, I don't see how a subsequenty `CancelledError` would be raised. If the timeout is reached, a `TimeoutError` should be raised.

https://github.com/aio-libs/aiojobs/blob/522d10f2de04ee2cc3088fd962047847d40efa97/aiojobs/_job.py#LL117C9-L135C54

igboyes: @markus Please resolve the other comments here. Reece will deal with this when you are done.

SnickerDoodlePop:
Seems like `TaskRunner._app` is now redundant, but it was there before my changes so I don't know if it's appropriate to remove due to possible future use.

igboyes:
> Seems like `TaskRunner._app` is now redundant, but it was there before my changes so I don't know if it's appropriate to remove due to possible future use.

It is unused and can be turfed.

PR #2881, opened 2023-05-12
feat: pg fixture is run once across all pg tests

SnickerDoodlePop:


igboyes (changes requested):
Check snapshots:

1. Why are `id` values changing? Possible that `TRUNCATE` is not being run between tests.
2. Some snapshots have concerning changes where data is lost. See `test_files.py`.

PR #2883, opened 2023-05-26
chore: remove deprecated query middleware

SnickerDoodlePop:


igboyes (commented):
Is `virtool.http.query` still used elsewhere?

PR #2884, opened 2023-05-29
feat: add BLAST search timeout

SnickerDoodlePop:


virtool/blast/task.py
```diff
@@ -64,34 +64,56 @@ async def wait(self):
...
 
         """
-        try:
-            while True:
-                await asyncio.sleep(self.interval)
 
-                blast = await self.data.blast.check_nuvs_blast(
-                    self.analysis_id, self.sequence_index
-                )
+        for _ in range(0, retries):
```
ReeceHoffmann, line 68: Not sure if a `for` loop makes sense for this. We only want to run the blast analysis once, but will run it more times if we have to.

virtool/blast/task.py
```diff
@@ -64,34 +64,56 @@ async def wait(self):
...
+                    await asyncio.sleep(self.interval)
 
-                if blast.ready:
-                    logger.info("Retrieved result for BLAST %s", blast.rid)
-                    break
+                    blast_coroutine = self.data.blast.check_nuvs_blast(
+                        self.analysis_id, self.sequence_index
+                    )
+
+                    blast = await asyncio.wait_for(blast_coroutine, 600)
```
ReeceHoffmann, line 77: I like the usage of `wait_for` but in this case I don't think this will work the way you intended. This will only timeout if a single check takes more than 10 minutes. Would be more appropriate if it was wrapping the the whole checking loop. (An individual check should take seconds at most, what we want to do is retry the blast if the remote run isn't done after 10 minutes)

virtool/blast/task.py
```diff
@@ -53,7 +53,7 @@ async def request(self):
 
         self.rid = blast.rid
 
-    async def wait(self):
+    async def wait(self, retries: int = 3):
```
ReeceHoffmann, line 56: In normal usage this function will never be called with a value for `retries` will only ever by the default value. Take a look at `BaseTask` for a detailed understanding of why. I would personally hardcode it at 3.

ReeceHoffmann (changes requested):
Code needs tweaking before it will behave as desired. Please reach out if you have any questions about the design goals.

virtool/blast/task.py
```diff
@@ -53,45 +53,69 @@ async def request(self):
...
+            try:
+                await asyncio.wait_for(wait_with_timeout(), 10)
+                break
 
-                logger.debug(
-                    "Checked BLAST %s. Waiting %s seconds", blast.rid, self.interval
-                )
+            except asyncio.TimeoutError:
+                if retries < 3:
+                    retries += 1
```
ReeceHoffmann, line 116: Stylistic thing, but I personally check fo `retries >= 3` then bail out if that is the case. Allows continuing to be the default case

virtool/blast/task.py
```diff
@@ -53,45 +53,69 @@ async def request(self):
...
 
+        The BLAST will be retried up to 3 times if a single BLAST
+        search exceeds 10 minutes
         """
-        try:
-            while True:
-                await asyncio.sleep(self.interval)
 
-                blast = await self.data.blast.check_nuvs_blast(
+        async def wait_with_timeout():
```
ReeceHoffmann, line 70: The logic here feels a bit more side-effect prone than it needs to be. I think in this case it would make more sense to remove `wait_with_timeout` from the class and reduce the scope of what it does.  Have it focus exclusively on the core loop along the lines of:


 1.  Wait for an the starting interval
 2. check if the blast is ready
      - if error or ready, return the result
      - else goto next step
 3. increment the interval
 4. goto step 1

Move all the extra logic of parsing what the returned result means or handling errors to the calling loop.

virtool/blast/task.py
```diff
@@ -53,45 +53,69 @@ async def request(self):
...
-                        "Encountered error for BLAST %s: %s", blast.rid, blast.error
-                    )
-                    await self._set_error(blast.error)
-                    break
+        retries: int = 0
 
-                self.interval += 5
+        while True:
+            try:
+                await asyncio.wait_for(wait_with_timeout(), 10)
```
ReeceHoffmann, line 111:
```suggestion
                await asyncio.wait_for(wait_with_timeout(), 600)
```

ReeceHoffmann (changes requested):
Looks much closer to doing what we want. I left a few stylistic comments in the code but there were a few things that didn't really fit elsewhere:

1. Everytime the job timoutes a new blast search should be attempted on NCBI's servers. It will probably make sense to merge the `request` and `wait` steps of the task based on this requirement
2. The waiting interval should reset when each new blast search is started. As written I don't think it does
3. The Blast should be deleted when either:
   - The task is cancelled  **OR**
   - All three attempts have failed due to timeout

ReeceHoffmann (changes requested):

virtool/blast/task.py
```diff
@@ -78,20 +76,61 @@ async def wait(self):
...
+                await self.data.blast.delete_nuvs_blast(
+                    self.analysis_id, self.sequence_index
+                )
+
+                logger.info("Deleted BLAST due to timeout: %s", self.rid)
+
+                blast_timeout_count += 1
+
+                if blast_timeout_count >= 3:
+                    logger.info("BLAST exceeded allowed attempts: %s", self.rid)
```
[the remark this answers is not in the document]
SnickerDoodlePop: yes, on line 111

virtool/blast/task.py
```diff
@@ -41,57 +41,96 @@ def __init__(
 
         self.analysis_id = self.context["analysis_id"]
         self.sequence_index = self.context["sequence_index"]
-        self.interval = 3
-        self.steps = [self.request, self.wait]
+        self.interval: Optional[int] = None
```
igboyes, line 44: Why this change?

virtool/blast/task.py
```diff
@@ -41,57 +41,96 @@ def __init__(
...
         """
-        Wait until the BLAST search completes.
 
-        Keep check the BLAST status on NCBI with increasingly longer intervals between
-        checks.
+        blast_timeout_count: int = 0
 
-        Checks are conducted by the data layer and will store the results or error when
-        the search completes. The task completes when either an error or
+        while True:
```
igboyes, line 61: I don't think your `while...break` pattern is doing anything. Look closely and either fix or reply back with explanation.

Another point, if an exception is raised in your `while` block, it will break the loop. You should put the `try` outside the loop and handle exceptions outside the loop as well.

virtool/blast/task.py
```diff
@@ -41,57 +41,96 @@ def __init__(
...
+
+    async def wait_for_blast_search(self):
+        """
+        Wait until the BLAST search completes.
+
+        Keep check the BLAST status on NCBI with increasingly longer intervals between
+        checks.
+        """
+
+        self.interval = 3
```
igboyes, line 108: Why not put this back in `__init__`? Maybe I'm missing something.

igboyes (changes requested):
Getting close.

igboyes (approved):

PR #2885, opened 2023-05-31
test: reconfigure testing to use separate backing services

ReeceHoffmann:
Changes:
 - Adds docker compose file that brings up backing services needed for testing
    - Services only include the core essentials (mongo, postgres, redis, openfga) and all data is volatile. When the container or volumes goes down the data is removed. (intentional as there is no productive reason to store test data)
    - Compatible with the automatic member dectection of a replicaset
    - Significant increase in testing speed when paired with multiple works vs minikube implementation
  - Changes all default values to assume use the default ports for the docker testing env
     - maintains compatibility with standard ports when passing correct config strings
  - add config option for open_fga host and propagates value through testing env

**Important:** Without configuration pytest assumes that compose will be running and hit the predefined ports. To get a working run a docker environment must be brought up using the new compose file.


tests/docker-compose-test
```diff
@@ -0,0 +1,69 @@
...
+      - migrate
+    ports:
+      - "9004:8080"
+      - "9005:3000"
+
+volumes:
+  postgres_openfga_data:
+    driver_opts:
+      type: tmpfs
+      device: tmpfs
```
igboyes, line 69: Not necessary as it doesn't matter if this is lost.

tests/docker-compose-test
```diff
@@ -0,0 +1,69 @@
+version: "3.1"
```
igboyes, line 1: Rename this file to `docker-compose.yml`.

igboyes (changes requested):

igboyes (approved):

PR #2893, opened 2023-06-09
feat: implement feature flag infrastructure

ChristineWC:


igboyes (changes requested):
This implementation has some problems.

First, read up about why global variables are bad (this is what you are doing here). Like most rules in CS it has exceptions, but this is not one of them. In fact, this is a good example of the problems with globals.

You are assuming the order of execution is going to be:
1. Class attributes defined based on hardcoded flags.
2. Some attributes changed based on CLI.
3. Attributes accessed to do flag checks.

I think your assumption is correct as it stands, but a few changes could make this break spectacularly. The most likely problem would be that the flags are accessed before their values are set from the CLI options. This means you'll only get the hardcoded values.

Redesign this with the following in mind:

The `FeatureFlags` class is a good idea. But, it needs some work. It should be instantiated on server start and updated with the dynamic values from the CLI in the server startup. The flags should be passed around at `app["flags"]`. The route decorators will have to change to get access to these values and you may have to write a middleware. Check how it works for `@policy` and contact me if unclear. I am expecting it.

I think we can forget the feature flags as module constants. I don't see how it can work elegantly. Instead hard-code the flags as class attributes or in the `__init__` method.

It would be really nice to define the possible flag names in one place (probably an enum), though I am not sure this is possible. This would prevent people from trying to access non-existent flags (ie. typos). All examples I found just use simple string keys like:
```py
enabled = flags.check("spaces")
```

Take a look and let me know what you find. If we just use strings, the `check` should return `False` by default and log a warning about an invalid flag being reference. Or, just raise and exception.

virtool/config/cli.py
```diff
@@ -78,12 +78,22 @@ def server():
 @postgres_connection_string_option
 @redis_connection_string_option
 @sentry_dsn_option
-def start_api_server(**kwargs):
+@click.option(
+    "--flags",
+    help="feature flag of the feature to enable",
+    type=click.Choice(
+        ['FF_ADMINISTRATOR_ROLES', 'FF_ML_MODELS', 'FF_SPACES'], case_sensitive=True
+    ),
```
igboyes, line 86: Try this:
```suggestion
    type=FeatureFlag
```

If this doesn't work:
1. Check if there is some other way to make `click` work with an enum.
2. Use a comprehension to iterate through the enum instead of hardcoding the list.

virtool/flags.py
```diff
@@ -0,0 +1,85 @@
+from typing import List, Callable, Dict
+from enum import Enum
+from aiohttp import web
+from aiohttp.web_exceptions import HTTPNotFound
+from aiohttp.web import Request
+
+import virtool.http.routes
+from virtool.types import App
+
+routes = virtool.http.routes.Routes()
```
igboyes, line 10:
```suggestion
routes = Routes()
```

virtool/config/cls.py
```diff
@@ -56,6 +56,7 @@ class ServerConfig:
     redis_connection_string: str
     use_b2c: bool
     sentry_dsn: Optional[str]
+    cli_flags: List[str] = field(default_factory=list)
```
igboyes, line 59: Redundant prefix. No other options start with `cli_`...
```suggestion
    flags: List[str] = field(default_factory=list)
```

virtool/administrators/api.py
```diff
@@ -33,6 +34,7 @@
 
 
 @routes.view("/admin/roles")
+@flag(FlagName.FF_ADMINISTRATOR_ROLES.value)
```
igboyes, line 37: Should look like this. See other comments for explanation.
```suggestion
@flag(FlagName.ADMINISTRATOR_ROLES)
```

igboyes (changes requested):
Hopefully this all makes sense!

virtool/flags.py
```diff
@@ -1,46 +1,50 @@
...
 
 
 class FeatureFlags:
     """
-    Class contains the derived value of the feature flags and cli flag config.
+    Class contains the derived value of the feature flags.
     """
 
-    def __init__(self):
+    def __init__(self, cli_flags: List[FlagName]):
```
igboyes, line 19: The `FeatureFlags` class doesn't care where the flags are passed from (good design). Using `cli_` implies coupling unnecessarily.

```suggestion
    def __init__(self, overrides: List[FlagName]):
```

virtool/flags.py
```diff
@@ -1,46 +1,50 @@
...
-            for constant_name, constant_value in self.constants.items()
-        }
+        for attr_name, constant_value in self.constants.items():
+            setattr(
+                self,
+                attr_name.name,
+                constant_value if attr_name not in cli_flags else True,
+            )
+
+    def check_flag_enabled(self, feature_flag: FlagName):
```
igboyes, line 38: * Docstring including `:param:` please.
* Missing return type.
* Why would you receive `None` for `feature_flag`? The typing explicity contradicts this.

virtool/flags.py
```diff
@@ -1,46 +1,50 @@
...
+        if feature_flag is None:
+            return None
+        return getattr(self, feature_flag.name)
 
 
-def flag(feature_flag: str):
+def flag(feature_flag: FlagName):
     """
     Sets the feature flag attribute for the route.
+    :param feature_flag: feature flag name associated with the route
```
igboyes: This docstring is not very helpful.

Imagine someone wants to use this decorator. First, they can't easily know it is a decorator unless it is mentioned in the docstring. Then, the docstring just tells them some low-level detail instead of what the function will actually accomplish for them.

When writing docstrings, focus on 

```suggestion
    Prevents access to the decorated request handler then the passed ``feature_flag`` is not enabled.
    
    :param feature_flag: feature flag name associated with the route
```

virtool/flags.py
```diff
@@ -1,46 +1,50 @@
...
-    FF_ADMINISTRATOR_ROLES = "FF_ADMINISTRATOR_ROLES "
+    SPACES = "SPACES"
+    ML_MODELS = "ML_MODELS"
+    ADMINISTRATOR_ROLES = "ADMINISTRATOR_ROLES"
 
 
 class FeatureFlags:
     """
-    Class contains the derived value of the feature flags and cli flag config.
+    Class contains the derived value of the feature flags.
```
igboyes, line 16: > A store for feature flags.
> 
> Maintains default values for feature flags and can accept overrides as a list of enabled ``flags``.

virtool/flags.py
```diff
@@ -1,46 +1,50 @@
...
-        return {
-            constant_name: constant_value if constant_name not in cli_flags else True
-            for constant_name, constant_value in self.constants.items()
-        }
+        for attr_name, constant_value in self.constants.items():
+            setattr(
+                self,
+                attr_name.name,
+                constant_value if attr_name not in cli_flags else True,
+            )
```
igboyes: * Just integrate this in `__init__`.
* Change `self.constants` to `self._flags`. They are not constants. The `_` indicates they are private and should not be accessed outside the object; the only way to check a flag is with `check_flag_enabled`.

virtool/flags.py
```diff
@@ -1,46 +1,50 @@
...
+    def set_derived_flag_values(self, cli_flags: List[FlagName]):
         """
-        Derives flag values based on the hardcoded values and the cli flags.
+        :param cli_flags:
         """
-        return {
-            constant_name: constant_value if constant_name not in cli_flags else True
-            for constant_name, constant_value in self.constants.items()
-        }
+        for attr_name, constant_value in self.constants.items():
```
igboyes, line 31: Fix these variable names. These ones are uninformative; what are the variables really.

igboyes (changes requested):
Getting there.

igboyes (approved):

PR #2894, opened 2023-06-09
feat: enable live backend reload

SnickerDoodlePop:


pyproject.toml
```diff
@@ -28,6 +28,7 @@ classifiers = [
 profile = "black"
 
 [tool.poetry.dependencies]
+aiohttp-devtools = "^0.0.2"
```
ReeceHoffmann, line 31: Should be a dev-dependency as we shouldn't be using this in prod

Dockerfile
```diff
@@ -14,5 +14,5 @@ COPY alembic.ini poetry.lock pyproject.toml run.py VERSION* /virtool/
 COPY virtool /virtool/virtool
 COPY assets /virtool/assets
 EXPOSE 9950
-ENTRYPOINT ["python", "run.py"]
+ENTRYPOINT ["adev runserver", "run.py"]
```
ReeceHoffmann, line 17: I would actually leave the dockerfile as is since it provides build instructions for the production ready containers. We only  want hot reloading in dev, so I would instead change the configuration of the `tiltfile` in the `dev` repo.

ReeceHoffmann (changes requested):
If you haven't already please verify in tilt that changes are being reloaded without needing to remake the container. This will require making changes to the `tiltfile` located in dev since the current configuration reloads the whole container whenever a change is made locally

Dockerfile
SnickerDoodlePop: ok; will create a new pr for that repo. I updated the pyproject.toml re. the dev dependency

ReeceHoffmann (dismissed):

ReeceHoffmann (commented):
When you changed it to being a dev dep did you use the poetry cli? Manually moving it might result in a desync between the pyproject.toml and poetry.lock

ReeceHoffmann (approved):

PR #2895, opened 2023-06-09
fix: add task to relist waiting jobs

ReeceHoffmann:
Changes:
 - add new `relist_jobs` task that adds any jobs with the state `waiting` to the redis queue provided they are not already in it. 
   - To handle edge case where a task has been removed from the list but has not yet been set to preparing the task checks the list twice with a 10 second delay between. Only the values in both lists are added back to the redis list 
 - adds barebones `admin/actions` api route which allows invocation of new `relist_jobs` task via api request
   - Possible todos: Add more actions to the list of available ones, define a standard `action` data structure and return type, add action table storing when and who invoked an action, add `get` route to permit fetching of action data

### DIFF
--- a/tests/administrators/test_api.py
+++ b/tests/administrators/test_api.py
@@ -188,3 +188,13 @@ class TestUpdateUser:
             body = await resp.json()
             assert body["force_reset"] is True
             assert body == snapshot
+
+
+@pytest.mark.apitest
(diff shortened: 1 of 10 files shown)


igboyes (approved):
Works for me. Please deploy and run when ready.

PR #2904, opened 2023-06-27
chore: remove use of function on startup

SnickerDoodlePop:
remove use of `recalculate_all_workflow_tags`
in `virtool/migrate::migrate`

add `RecalculateWorkflowTags` as a recurring task

### DIFF
--- a/virtool/mongo/migrate.py
+++ b/virtool/mongo/migrate.py
@@ -25,7 +25,7 @@ async def migrate(app: App):
     :param app: the application object
 
     """
-    await gather(migrate_status(app["db"]), recalculate_all_workflow_tags(app["db"]))
+    await gather(migrate_status(app["db"]))
 
 
 async def recalculate_all_workflow_tags(mongo: Mongo):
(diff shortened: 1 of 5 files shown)


virtool/mongo/migrate.py
```diff
@@ -25,7 +25,7 @@ async def migrate(app: App):
     :param app: the application object
 
     """
-    await gather(migrate_status(app["db"]), recalculate_all_workflow_tags(app["db"]))
+    await gather(migrate_status(app["db"]))
 
 
 async def recalculate_all_workflow_tags(mongo: Mongo):
```
igboyes, line 31: Should this really stay here? Does it even need to be its own function anymore?

virtool/mongo/migrate.py
```diff
@@ -25,7 +25,7 @@ async def migrate(app: App):
     :param app: the application object
 
     """
-    await gather(migrate_status(app["db"]), recalculate_all_workflow_tags(app["db"]))
+    await gather(migrate_status(app["db"]))
```
igboyes, line 28: What purpose does this serve now?

igboyes (approved):

PR #2906, opened 2023-06-28
Markus/vir 1827 use transforms for attaching groups

SnickerDoodlePop:


SnickerDoodlePop:
This pr is UNFINISHED

I have only created the `AttachGroupTransform` and `AttachGroupsTransform` classes so far, along with automated tests for them.

SnickerDoodlePop:
This is a WORK IN PROGRESS

Currently only `AttachGroupTransform` is functional; `AttachGroupsTransform` is a work in progress.

SnickerDoodlePop:
Both transforms are now complete.

ReeceHoffmann:
Ready for review or WIP still?

virtool/groups/transforms.py
```diff
@@ -0,0 +1,135 @@
...
+
+            Returns:
+                A dictionary of complete groups indexed
+                by the input documents' `id` fields
+        """
+        group_ids: List[str | None] = list(
+            {document.get("primary_group", None) for document in documents}
+        )
+
+        group_dict = {
```
ReeceHoffmann, line 63:
```suggestion
        groups = {
```

virtool/groups/transforms.py
```diff
@@ -0,0 +1,135 @@
...
+
+            Returns:
+                A dictionary of lists of complete groups
+                indexed by the input documents' `id` fields
+        """
+        group_ids = list(
+            {group for document in documents for group in document["groups"]}
+        )
+
+        group_dict = {
```
ReeceHoffmann, line 127:
```suggestion
        groups= {
```

tests/groups/test_db.py
```diff
@@ -0,0 +1,83 @@
...
+) -> list[dict] | dict:
+    if quantity <= 0:
+        return None
+    
+    fake_users: list[dict] = []
+
+    for i in range(0, quantity):
+        fake_user: dict = {}
+
+        fake_user.update({"id": f"doc_{i}"})
```
ReeceHoffmann, line 20:
```suggestion
        fake_user = ({"id": f"doc_{i}"})
```
?

tests/groups/test_db.py
```diff
@@ -0,0 +1,83 @@
+from virtool.fake.next import DataFaker
+from virtool.groups.transforms import AttachPrimaryGroupTransform, AttachGroupsTransform
+from virtool.data.transforms import apply_transforms
+
+
+async def create_fake_users_with_groups(
```
ReeceHoffmann, line 6: Docs

ReeceHoffmann (changes requested):
Core logic looks mostly good to me. Docstrings need to be updated to match other modules in the project. `users.session.py` has good examples of auto-documentation compatible docstrings.

tests/groups/test_db.py
```diff
@@ -0,0 +1,85 @@
...
+
+        complete_docs = await apply_transforms(
+            docs, [AttachPrimaryGroupTransform(mongo)]
+        )
+
+        assert complete_docs == snapshot
+
+
+class TestAttachGroupsTransform:
+    async def test_groups_missing(self, snapshot, mongo):
```
ReeceHoffmann, line 66:
```suggestion
    async def test_no_groups(self, snapshot, mongo):
```

tests/groups/test_db.py
```diff
@@ -0,0 +1,85 @@
...
+        doc = (await create_fake_users_with_groups(None, 1, False, False))[0]
+
+        complete_doc = await apply_transforms(doc, [AttachGroupsTransform(mongo)])
+
+        assert complete_doc == snapshot
+
+    async def test_groups_present(self, snapshot, fake2, mongo):
```
ReeceHoffmann, line 73:
```suggestion
    async def test_single_document(self, snapshot, fake2, mongo):
```

tests/groups/test_db.py
```diff
@@ -0,0 +1,85 @@
...
+
+class TestAttachGroupTransform:
+    async def test_primary_group_missing(self, snapshot, mongo):
+        doc = (await create_fake_users_with_groups(None, 1, False, False))[0]
+
+        complete_doc = await apply_transforms(doc, [AttachPrimaryGroupTransform(mongo)])
+
+        assert complete_doc == snapshot
+
+    async def test_primary_group_present(self, snapshot, fake2, mongo):
```
ReeceHoffmann, line 48:
```suggestion
    async def test_single_document(self, snapshot, fake2, mongo):
```

virtool/groups/transforms.py
```diff
@@ -0,0 +1,130 @@
...
+from virtool.utils import base_processor
+from virtool_core.models.group import GroupMinimal
+
+if TYPE_CHECKING:
+    from virtool.mongo.core import Mongo
+
+
+class AttachPrimaryGroupTransform(AbstractTransform):
+    """
+    Attaches a Primary Group to a User document
```
ReeceHoffmann, line 13:
```suggestion
    Attach primary groups to document(s) with a `primary_group` field
```

virtool/groups/transforms.py
```diff
@@ -0,0 +1,130 @@
...
+
+        return {
+            document["id"]: groups.get(document.get("primary_group"))
+            for document in documents
+        }
+
+
+class AttachGroupsTransform(AbstractTransform):
+    """
+    Attaches Groups to a User document
```
ReeceHoffmann, line 75:
```suggestion
    Attach Groups to document(s) containing a `groups` field
```

ReeceHoffmann (changes requested):
Getting close imo. Just some suggested docstring tweaks. Generally did two things:

1. Move details about a specific parameter to descriptions of that parameter 
2. Wrote function description in imperative voice (Attach a groups vs Attaches, reads like a command rather than a description)

virtool/groups/transforms.py
```diff
@@ -0,0 +1,125 @@
...
+
+    async def prepare_many(
+        self, documents: List[Document]
+    ) -> Dict[Union[int, str], Any]:
+        """
+        Bulk prepare groups for attachment to passed documents
+
+        :param documents: A list of input documents with `groups` fields
+        :return: a dictionary of `document["id"]:list[group]`
+        pairs based on each document's `groups` field
```
ReeceHoffmann, line 111: Is this close to 90 chars wide?

ReeceHoffmann (changes requested):
What is here is looking good to me. imo if you remove the `lookup_group(s)_by_id` functions then  this is good with me

virtool/users/db.py
```diff
@@ -310,10 +310,6 @@ async def fetch_complete_user(
         mongo.users.aggregate(
```
ReeceHoffmann, line 310: Looks to me like the aggregation pipeline here is also redudant

virtool/account/data.py
```diff
@@ -294,7 +304,7 @@ async def delete_key(self, user_id: str, key_id: str):
         if delete_result.deleted_count == 0:
             raise ResourceNotFoundError()
 
-    async def login(self, data: CreateLoginRequest) -> Union[str]:
+    async def login(self, data: CreateLoginRequest) -> Union[str, int]:
```
ReeceHoffmann, line 307: Why the type change?

virtool/account/db.py
```diff
@@ -76,7 +78,6 @@ async def fetch_complete_api_key(mongo, key_id: str) -> Optional[APIKey]:
     async for key in mongo.keys.aggregate(
         [
             {"$match": {"id": key_id}},
```
ReeceHoffmann, line 80: Aggregation pipeline seems redundant now

virtool/administrators/data.py
```diff
@@ -89,18 +89,16 @@ async def find(
             per_page,
             client_query,
             sort="handle",
```
ReeceHoffmann, line 91: Redundant pipeline

ReeceHoffmann (changes requested):

tests/groups/test_db.py
```diff
@@ -0,0 +1,85 @@
+from virtool.fake.next import DataFaker
+from virtool.groups.transforms import AttachPrimaryGroupTransform, AttachGroupsTransform
+from virtool.data.transforms import apply_transforms
+
+
+async def create_fake_users_with_groups(
```
igboyes, line 6: **Points (if you were going to retain this design)**
* This will break if someone ever passes `use_primary_group=True` without passing `faker`. Always require `faker`.
* You could make this function a fixture so you can directly inject `fake2` instead of passing it from the test function. You can define another function in the fixture body that takes your other args, then return the inside function.
* The method `fake2.users.create()` can do a better job of what you are doing here. It has keyword arguments to do what you want.
* You can [avoid flags](https://martinfowler.com/bliki/FlagArgument.html) completely by changing the design.

**To do**

* Get rid of all arguments (eg. `faker`, `quantity`, etc). Create a selection of users that have the characteristics you want across your tests and use the same set of users in each test. This reduces complexity and makes the test more robust because you don't have to carefully construct a different set of data using flags for each test case.

  You can then get the raw user documents you need for each test case using `mongo`.
* Make this a fixture. Rename to a more noun-y name instead of verb-y since it isn't a callable anymore.
* Let `fake2` take care of the user IDs instead of creating your own using `i`.

igboyes (changes requested):
This looks great, but needs some simplification in the test code.

tests/groups/test_db.py
```diff
@@ -43,24 +49,31 @@ async def test_multiple_documents(self, snapshot, fake2, mongo):
...
+            incomplete_users,
+            [AttachGroupsTransform(mongo), AttachPrimaryGroupTransform(mongo)],
         )
 
+        for user in complete_users:
+            assert user == snapshot(
+                matcher=path_type(
+                    {"last_password_change": (datetime,), "password": (bytes,)}
+                )
+            )
```
igboyes, line 62: You can do one assertion by using regex path matching.

https://tophat.github.io/syrupy/#assertion-options

It would probably look like:
```py
path_type({
  ".*last_password_change": (datetime,),
  ".*password": (bytes,)
}, regex=True)
```

igboyes (changes requested):
Close.

igboyes (approved):

PR #2910, opened 2023-07-05
chore: merge job processor into fetch_complete_job

ChristineWC:


tests/jobs/test_db.py
```diff
@@ -1,12 +1,12 @@
...
-    """
+async def test_fetch_complete_job(snapshot, mongo, fake2, static_time, test_job):
     user = await fake2.users.create()
     test_job["user"] = {"id": user.id}
-    assert await processor(mongo, test_job) == snapshot
+
+    result = await fetch_complete_job(mongo, test_job)
+
+    assert isinstance(result, Job)
+    snapshot.assert_match(result)
```
igboyes, line 12: Simplify this. It is obvious from the snapshot what the type of the return value is.
```suggestion
    assert await fetch_complete_job(mongo, test_job) == snapshot
```

igboyes (changes requested):

tests/jobs/test_db.py
```diff
@@ -1,9 +1,8 @@
...
 from virtool.jobs.db import fetch_complete_job
 
-async def test_fetch_complete_job(snapshot, mongo, fake2, static_time, test_job):
+
+async def test_fetch_complete_job(snapshot, fake2, mongo, static_time, test_job):
     user = await fake2.users.create()
     test_job["user"] = {"id": user.id}
 
-    assert await fetch_complete_job(mongo, test_job) == snapshot
+    assert repr(await fetch_complete_job(mongo, test_job)) == snapshot
```
igboyes, line 8: Why are you bothering to use `repr`? Wasn't it working without?

igboyes (changes requested):

tests/jobs/test_db.py
ChristineWC: Without using repr, it doesn't include the class names which are in the generated snapshot

igboyes (approved):

PR #2911, opened 2023-07-07
feat: support user permission fetching from PostgreSQL

ReeceHoffmann:
Changes:
  - Removes usage of permission data save in user object
    - No longer reads or writes permission changes to `users` table. Does **NOT** delete any existing data from the users table
  - Add transform that takes a user object with groups and attached populated permissions field
    - transform accepts both group `ids` and groups as dictionaries. 
    - transform accepts data from both postgreSQL and mongo. Any group that exist in either is included. If a group exists in both postgres is assumed to be correct
  -   Updates tests as needed to account for permissions no longer being part of the db definition of a `user`.


tests/subtractions/test_api.py
```diff
@@ -401,10 +403,10 @@ async def test_download_subtraction_files(
         return
 
     fasta_expected_path = (
-            client.app["config"].data_path / "subtractions" / "foo" / "subtraction.fa.gz"
+        client.app["config"].data_path / "subtractions" / "foo" / "subtraction.fa.gz"
     )
     bowtie_expected_path = (
-            client.app["config"].data_path / "subtractions" / "foo" / "subtraction.1.bt2"
+        client.app["config"].data_path / "subtractions" / "foo" / "subtraction.1.bt2"
```
igboyes, line 409: Take opportunity to change this. Provides typing.
```suggestion
        get_config_from_app(client.app).data_path / "subtractions" / "foo" / "subtraction.1.bt2"
```

virtool/account/data.py
```diff
@@ -47,11 +48,16 @@ class AccountData(DataLayerPiece):
...
+        self,
+        db: Mongo,
+        redis: Redis,
+        authorization_client: AuthorizationClient,
+        pg: AsyncEngine,
     ):
         self._authorization_client = authorization_client
         self._db = db
         self._redis = redis
+        self._pg = pg
```
igboyes, line 60: Please change `db` and `_db` to `mongo` and `_mongo`.

virtool/administrators/data.py
```diff
@@ -43,9 +47,12 @@
 class AdministratorsData(DataLayerPiece):
     name = "administrators"
 
-    def __init__(self, authorization_client: AuthorizationClient, mongo: "DB"):
+    def __init__(
+        self, authorization_client: AuthorizationClient, mongo: "DB", pg: AsyncEngine
```
igboyes, line 51:
```suggestion
        self, authorization_client: AuthorizationClient, mongo: "Mongo", pg: AsyncEngine
```

virtool/fake/next.py
```diff
@@ -142,6 +143,7 @@ async def create(
         handle: Optional[str] = None,
         groups: Optional[List[Group]] = None,
         primary_group: Optional[Group] = None,
+        administrator_role: AdministratorRole = None,
```
igboyes, line 146:
```suggestion
        administrator_role: Optional[AdministratorRole] = None,
```

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class AttachPermissionsTransform(AbstractTransform):
+    """
+    Attaches more complete task data to a document with a `task.id` field.
+    """
+
+    def __init__(self, pg: AsyncEngine, mongo: "Mongo"):
+        self._pg = pg
+        self._mongo = mongo
```
igboyes, line 25:
```suggestion
    def __init__(self, mongo: "Mongo", pg: AsyncEngine):
        self._mongo = mongo
        self._pg = pg
```

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    **document,
+                    "permissions": merge_group_permissions(
+                        [
+                            prepared[group_id]
+                            for group_id in self.get_group_ids(document)
+                        ]
+                    ),
+                }
+            )
```
igboyes, line 51: Use a comprehension. There will be a perf benefit and Black will make it readable.

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+                    select(SQLGroup).filter(SQLGroup.legacy_id.in_(group_ids))
+                ),
+            )
+
+            mongo_groups = {group["_id"]: group for group in mongo_result}
+            pg_groups = {
+                group.legacy_id: group.to_dict() for group in pg_result.scalars()
+            }
+
+            return mongo_groups | pg_groups
```
igboyes, line 92: Transaction not necessary for Mongo as nothing is being modified.

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @staticmethod
+    def get_group_ids(user) -> set[str]:
+        groups = get_safely(user, "groups")
+        return {get_safely(group, "id") or group for group in groups}
```
igboyes, line 97: When will group not have `id`? Isn't that concerning?

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+        group_ids = set()
+
+        for user in documents:
+            group_ids.update(self.get_group_ids(user))
+
+        return await self.get_groups_from_ids(list(group_ids))
+
+    async def get_groups_from_ids(self, group_ids: list[str]):
+        if len(group_ids) == 0:
+            return {}
```
igboyes, line 71:
```suggestion
        if not group_ids:
            return {}
```

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+from virtool.types import Document
+from virtool.utils import get_safely
+
+if TYPE_CHECKING:
+    from virtool.mongo.core import Mongo
+
+
+class AttachPermissionsTransform(AbstractTransform):
+    """
+    Attaches more complete task data to a document with a `task.id` field.
```
igboyes, line 20: Take another look.

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+        self, documents: List[Document]
+    ) -> Dict[Union[int, str], Any]:
+        group_ids = set()
+
+        for user in documents:
+            group_ids.update(self.get_group_ids(user))
+
+        return await self.get_groups_from_ids(list(group_ids))
+
+    async def get_groups_from_ids(self, group_ids: list[str]):
```
igboyes, line 69:
```suggestion
    async def fetch_groups_by_ids(self, group_ids: list[str]):
```
Needs a docstring and correct typing.

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+
+            mongo_groups = {group["_id"]: group for group in mongo_result}
+            pg_groups = {
+                group.legacy_id: group.to_dict() for group in pg_result.scalars()
+            }
+
+            return mongo_groups | pg_groups
+
+    @staticmethod
+    def get_group_ids(user) -> set[str]:
```
igboyes, line 95:
```suggestion
    def get_group_ids_from_user(user) -> set[str]:
```
Needs docstring and correct typing.

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+        groups = get_safely(user, "groups")
```
igboyes, line 96: Not sure this is something that we should let fail silently. When are you going to be applying the transform but some documents will by missing the `groups` field?

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+
+        return attached
+
+    async def prepare_one(self, document) -> Optional[Document]:
+        group_ids = self.get_group_ids(document)
+        return await self.get_groups_from_ids(list(group_ids))
+
+    async def prepare_many(
+        self, documents: List[Document]
```
igboyes, line 60:
```suggestion
        self, documents: list[Document]
```
You are inconsistenly using type names list `Dict` and `dict`.

We should try to move to the new style `dict` and `list` instead of `List` and `Dict`. Instead of `Optional` and `Union`, you can also use `str | None` or `int | str`. I am not stuck on every module being updated, but new work should at least be consistent with what is in the module already.

virtool/users/transforms.py
```diff
@@ -0,0 +1,97 @@
...
+                            for group_id in self.get_group_ids(document)
+                        ]
+                    ),
+                }
+            )
+
+        return attached
+
+    async def prepare_one(self, document) -> Optional[Document]:
+        group_ids = self.get_group_ids(document)
```
igboyes, line 56: You call `get_group_ids` in both attach and prepare phases. You could reorganize things to avoid this waste. This may also allow you to avoid redefining `attach_many`.

virtool/users/db.py
```diff
@@ -301,7 +301,10 @@ async def update_keys(
 
 
 async def fetch_complete_user(
-    mongo: "Mongo", authorization_client: "AuthorizationClient", user_id: str
+    pg: AsyncEngine,
+    mongo: "Mongo",
+    authorization_client: "AuthorizationClient",
+    user_id: str,
```
igboyes, line 307: Please alphabetize.

igboyes (changes requested):

igboyes (approved):

PR #2916, opened 2023-07-21
fix: close expired websocket connections

ChristineWC:


virtool/account/api.py
```diff
@@ -352,6 +352,10 @@ async def get(self) -> r204:
             virtool.http.authentication.get_ip(self.request),
         )
 
+        await self.request.app["ws"].close_connections_with_session_id(
+            self.request.cookies.get("session_id")
+        )
```
ReeceHoffmann, line 357: While this works, it tightly couples the WS server to the API, ideally we would disconnect the user via a message over redis

virtool/ws/server.py
```diff
@@ -78,6 +78,24 @@ def remove_connection(self, connection: WSConnection):
...
+                await connection.close(1000)
+
+    async def close_expired_websocket_connections(self):
+        """
+        Closes connections with expired sessions.
+        """
+        while True:
+            for connection in self._connections:
+                if not connection.user_id:
+                    await connection.close(1001)
```
ReeceHoffmann, line 96: You'll need to check if the session is valid to determine if it should be expired. Query the `sessions` data_layer with the `session_id` to check if the session is valid. Keep in mind the session must be authenticated to be valid for a WS connection.

ReeceHoffmann (changes requested):

virtool/ws/server.py
```diff
@@ -78,6 +80,37 @@ def remove_connection(self, connection: WSConnection):
...
+        Periodically closes connections with expired sessions.
+        """
+        session_data = SessionData(self._redis)
+
+        while True:
+            for connection in self._connections:
+                try:
+                    session = await session_data.get_anonymous(connection.session_id)
+                    if session.authentication is None:
+                        await connection.close(1001)
```
ReeceHoffmann, line 109: Couple of points:
 1. We can assume that if a connection exists then it was made using an authenticated session
 2. `get_anonymous` should only have ever returned anonymous (non-authenticated) sessions, this has been fixed on latest

Bring your branch up to date and re-implement around the fixed sessions datalayer. You may need to add a new dataLayer function to check if a `session_id` is associated with an authenticated session without having a session token.

virtool/ws/server.py
```diff
@@ -78,6 +80,37 @@ def remove_connection(self, connection: WSConnection):
         except ValueError:
             pass
 
+    async def handle_redis_ws_messages(self):
+        channel = self._redis.subscribe("close_websocket_connections")
+
+        while await channel.wait_message():
+            session_id = await channel.get(encoding="utf-8")
+            await self.close_connections_with_session_id(session_id)
```
ReeceHoffmann, line 88: Remove the new Redis logic, we will assume that either the client will disconnect itself, or that the scan will do it soon enough for now.

virtool/ws/server.py
```diff
@@ -78,6 +80,37 @@ def remove_connection(self, connection: WSConnection):
...
+
+    async def close_connections_with_session_id(self, session_id):
```
ReeceHoffmann, line 90: Is this used?

ReeceHoffmann (changes requested):

virtool/users/sessions.py
```diff
@@ -181,6 +181,14 @@ async def get_authenticated(self, session_id: str, session_token: str) -> Sessio
...
         return Session(**session)
 
+    async def validate_session(self, session_id: str):
+        """
+        Validates a session with a given ``session_id``.
+
+        :param session_id: the session id
+        """
+        await self._get(session_id)
+
```
ReeceHoffmann, line 191: Good direction, couple of tweaks:

 1. rename along the lines of `check_authenticated`
 2.  return true if session is authenticated, return false if the session does not exist or is not authenticated 

I would avoid throwing errors in this flow as you can give a valid answer regardless of of the passed `session_id` exists or not

ReeceHoffmann (changes requested):

virtool/users/sessions.py
```diff
@@ -181,6 +181,20 @@ async def get_authenticated(self, session_id: str, session_token: str) -> Sessio
 
         return Session(**session)
 
+    async def check_session_is_authenticated(self, session_id: str) -> bool:
+        """
+        Checks whether a session is authenticated a given ``session_id``.
```
ReeceHoffmann, line 186:
```suggestion
        Checks whether a session is authenticated.
```

virtool/ws/server.py
```diff
@@ -78,6 +79,20 @@ def remove_connection(self, connection: WSConnection):
         except ValueError:
             pass
 
+    async def periodically_close_expired_websocket_connections(self):
+        """
+        Periodically closes connections with expired sessions.
```
ReeceHoffmann, line 84:
```suggestion
        Periodically check for and close connections with expired sessions
```

ReeceHoffmann (changes requested):
So close, just a couple docstring tweaks and I think we are good

ReeceHoffmann (approved):

PR #2918, opened 2023-07-26
refactor: use first user data layer check

ChristineWC:


virtool/users/data.py
```diff
@@ -219,9 +229,19 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
+        )
 
         if user is None:
             raise ResourceNotFoundError
 
         return user
+
+    async def check_is_empty(self) -> bool:
+        """
+        Checks if the "users" collection is empty.
```
igboyes, line 243: In retrospect, this naming is not great. Should be `check_if_any_exist`.

Please don't refer to the underlying implementation (users collection; MongoDB) in the docstring. The point of the data layer is to be an abstraction over the data sources.

virtool/users/data.py
```diff
@@ -219,9 +229,19 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
+        """
+        if await self._mongo.users.count_documents({}) == 0:
+            return True
+        return False
```
igboyes, line 247: 1. The query is not efficient... I know I wrote it in the issue. This is going to go through the whole collection to get a count even though we only want to know if the count is greater than `0`.

   Change to `await self._mongo.users.count_documents({}, limit=1) == 0`. This will stop the count as soon as one document is found.
2. The `return` can be cut to a single line.

tests/users/test_data.py
```diff
@@ -174,3 +174,15 @@ async def test_find_or_create_b2c_user(
         assert int(user.handle.split("-")[-1])
 
     assert user == snapshot(matcher=path_type({"handle": (str,)}))
+
+
+async def test_check_is_empty(mongo, users_data, fake2, authorization_client, pg):
```
igboyes, line 179: Needs doc string. What is the test doing.

Please don't combine test cases. Either use `pytest.mark.parametrize` or write two separate tests.

tests/users/test_data.py
```diff
@@ -174,3 +174,15 @@ async def test_find_or_create_b2c_user(
...
+    first_user_exists = not await users_data.check_is_empty()
+
+    assert first_user_exists is False
```
igboyes, line 182: Variable is unnecessary. Using `not` and `False` just adds to cognitive load. Be direct.

Same goes for the other place you do this.
```suggestion
    assert await users_data.check_is_empty() is True
```

virtool/users/api.py
```diff
@@ -126,7 +126,7 @@ async def put(self, data: CreateFirstUserRequest) -> Union[r201[User], r400, r40
         """
         db = self.request.app["db"]
 
-        if await db.users.count_documents({}, limit=1):
+        if not await db.users.check_is_empty():
```
igboyes, line 129: This is wrong. Please write an API test to make sure this will be caught in the future.

igboyes (changes requested):

tests/users/test_api.py
```diff
@@ -295,3 +292,16 @@ async def test_remove_permission(spawn_client, role, status, snapshot):
...
+    """
+    Checks response when there already is a user.
+    """
+    client = await spawn_client(authorize=True, administrator=True)
+
+    data = {"handle": "fred", "password": "hello_world"}
+
+    resp = await client.put("/users/first", data)
+
+    assert resp.status == 409
```
igboyes, line 307:
```suggestion
    resp = await client.put("/users/first",  {"handle": "fred", "password": "hello_world"})

    assert resp.status == 409
```

virtool/users/data.py
```diff
@@ -219,9 +229,17 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
         if user is None:
             raise ResourceNotFoundError
 
         return user
+
+    async def check_if_any_exist(self) -> bool:
+        """
+        Checks if any users exist.
+        """
+        return await self._mongo.users.count_documents({}, limit=1) == 0
```
igboyes, line 245: The name of the function and what it actually does are in contradiction.

igboyes (changes requested):

tests/users/test_data.py
```diff
@@ -174,3 +174,19 @@ async def test_find_or_create_b2c_user(
         assert int(user.handle.split("-")[-1])
 
     assert user == snapshot(matcher=path_type({"handle": (str,)}))
+
+
+async def test_check_user_does_not_exist(users_data):
```
igboyes, line 179: I think parametrization is an ideal solution here rather than having separate tests.

```suggestion
async def test_check_no_users_exist(users_data):
```

tests/users/test_data.py
```diff
@@ -174,3 +174,19 @@ async def test_find_or_create_b2c_user(
...
+    """
+    Checks that users don't exist.
```
igboyes, line 181: Come up with something accurate. What is the test doing?

igboyes (changes requested):

tests/users/test_data.py
```diff
@@ -176,17 +176,15 @@ async def test_find_or_create_b2c_user(
     assert user == snapshot(matcher=path_type({"handle": (str,)}))
 
 
-async def test_check_user_does_not_exist(users_data):
+@pytest.mark.parametrize(
+    "no_user_exists, sample_user_data",
+    [(True, None), (False, {"password": "hello_world", "handle": "bill"})],
```
igboyes, line 181: * Why do you have to parametrize the user `dict`?
* Shouldn't it be (`True`, `False`) instead of (`True`, `None`)?

tests/users/test_data.py
```diff
@@ -176,17 +176,15 @@ async def test_find_or_create_b2c_user(
     assert user == snapshot(matcher=path_type({"handle": (str,)}))
 
 
-async def test_check_user_does_not_exist(users_data):
+@pytest.mark.parametrize(
+    "no_user_exists, sample_user_data",
```
igboyes, line 180:
```suggestion
    "no_users_exists, sample_user_data",
```

igboyes (changes requested):

tests/users/test_api.py
```diff
@@ -295,3 +292,16 @@ async def test_remove_permission(spawn_client, role, status, snapshot):
...
+    """
+    Checks response when there already is a user.
+    """
+    client = await spawn_client(authorize=True, administrator=True)
+
+    resp = await client.put(
+        "/users/first", {"handle": "fred", "password": "hello_world"}
+    )
+
+    assert resp.status == 409
```
ReeceHoffmann, line 307: I think this test should cover two cases:

 1. That no users exist, and a new user is created with the passed info
 2. That a user does exist and no user is created

Since `spawn_client` creates a user, this test is only checking case 2.

virtool/users/data.py
```diff
@@ -219,9 +229,17 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
         if user is None:
             raise ResourceNotFoundError
 
         return user
+
+    async def check_no_users_exist(self) -> bool:
+        """
+        Checks that no users exist.
+        """
+        return await self._mongo.users.count_documents({}, limit=1) == 0
```
ReeceHoffmann, line 245:
```suggestion
    async def check_users_exist(self) -> bool:
        """
        Checks that no users exist.
        """
        return await self._mongo.users.count_documents({}, limit=1) > 0
```
I would rework this to be a check if users exist, then `not` the returned value as needed where called. As written when checking for the first user the logic reads as " if not no users exist" which is a harder to grok than simply just checking if users exist"

ReeceHoffmann (changes requested):

virtool/users/data.py
```diff
@@ -219,9 +229,17 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
+        )
 
         if user is None:
             raise ResourceNotFoundError
 
         return user
+
+    async def check_no_users_exist(self) -> bool:
+        """
+        Checks that no users exist.
```
ReeceHoffmann, line 243: Needs params

virtool/users/data.py
```diff
@@ -219,9 +229,18 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
 
         if user is None:
             raise ResourceNotFoundError
 
         return user
+
+    async def check_users_exist(self) -> bool:
+        """
+        Checks that users exist.
+        :returns: True if users exist otherwise False
```
ReeceHoffmann, line 244:
```suggestion
        Check for existing users.
        
        :returns: True if users exist otherwise False
```

tests/users/test_data.py
```diff
@@ -174,3 +174,13 @@ async def test_find_or_create_b2c_user(
...
+
+
+async def test_check_no_users_exist(users_data):
+    """
+    Verifies user existence checks work as expected with and without existing users.
+    """
+    assert await users_data.check_users_exist() is False
+
+    await users_data.create(password="hello_world", handle="bill")
+    assert await users_data.check_users_exist() is True
```
ReeceHoffmann, line 186: Comparison of `bool is True` is redundant since the return value is a bool, aka if it was `True` the comparison returns `True`, If `False` it returns `False`

tests/users/test_data.py
```diff
@@ -174,3 +174,13 @@ async def test_find_or_create_b2c_user(
         assert int(user.handle.split("-")[-1])
 
     assert user == snapshot(matcher=path_type({"handle": (str,)}))
+
+
+async def test_check_no_users_exist(users_data):
```
ReeceHoffmann, line 179: Naming should match underlying

tests/users/test_data.py
ReeceHoffmann, line 186: Should only do one case per test. Either us pytest's parameterise or separate tests wrapped in a class

tests/users/test_api.py
```diff
@@ -295,3 +292,25 @@ async def test_remove_permission(spawn_client, role, status, snapshot):
...
+
+    await mongo.users.delete_one({"handle": "bob"})
+
+    resp = await client.put(
+        "/users/first", {"handle": "fred", "password": "hello_world"}
+    )
+
+    assert resp.status == 201
```
ReeceHoffmann, line 316: Cases should be separate test or done with parameterise

ReeceHoffmann, line 316: Please include a snapshot of the response body for the successful case. I don't think we really use it on the FE but returning the newly created user document is part of the API defn

ReeceHoffmann (changes requested):
Close, few testing tweaks and it should be good

virtool/users/api.py
```diff
@@ -139,13 +138,11 @@ async def put(self, data: CreateFirstUserRequest) -> Union[r201[User], r400, r40
...
 
-        session_id, session, token = await get_data_from_req(
-            self.request
-        ).sessions.create_authenticated(
+        _, token = await get_data_from_req(self.request).sessions.create_authenticated(
             virtool.http.authentication.get_ip(self.request), user.id
         )
 
-        self.request["client"].authorize(session, is_api=False)
+        session_id = self.request.cookies.get("session_id")
```
ReeceHoffmann, line 145: You should get the session_id from the session returned when calling `created_authenticated`. As written I think this will keep the session_id as the old id from whatever session existed before. Would be a good idea to manually test to make sure the initial sign on flow works as expected. The latest dev version has a new parameter to allow spinning up of a temporary virtool instance

tests/users/test_api.py
```diff
@@ -295,3 +292,23 @@ async def test_remove_permission(spawn_client, role, status, snapshot):
...
+
+
+@pytest.mark.parametrize("first_user_exists, status", [(True, 409), (False, 201)])
+async def test_first_user_view(
+    spawn_client, mongo, first_user_exists, status, snapshot
+):
+    """
+    Checks response when first user exists and does not exist.
+    """
+    client = await spawn_client(authorize=True, administrator=True)
```
ReeceHoffmann, line 304: Why? Given that this is the first user route I think the most realistic route would be to have a client with no perms at all.

ReeceHoffmann (changes requested):

tests/users/test_api.py
```diff
@@ -295,3 +292,23 @@ async def test_remove_permission(spawn_client, role, status, snapshot):
...
+async def test_first_user_view(
+    spawn_client, mongo, first_user_exists, status, snapshot
+):
+    """
+    Checks response when first user exists and does not exist.
+    """
+    client = await spawn_client()
+
+    if not first_user_exists:
+        await mongo.users.delete_one({"handle": "bob"})
```
ReeceHoffmann, line 307:
```suggestion
        await mongo.users.delete_many({})
```
Prevent this test from causing problems the client users is ever changed

tests/users/test_data.py
```diff
@@ -174,3 +174,18 @@ async def test_find_or_create_b2c_user(
         assert int(user.handle.split("-")[-1])
 
     assert user == snapshot(matcher=path_type({"handle": (str,)}))
+
+
+class TestUserExistenceChecks:
```
ReeceHoffmann, line 179:
```suggestion
class TestCheckUsersExist:
```
Class name should closely match function name being tested

tests/users/test_data.py
```diff
@@ -174,3 +174,18 @@ async def test_find_or_create_b2c_user(
...
+    async def test_no_users_exist(self, users_data):
+        """
+        Verifies that user existence check returns False when no users exist.
```
ReeceHoffmann, line 182:
```suggestion
        Verify that the user existence check returns False when no users exist.
```

ReeceHoffmann (changes requested):

PR #2920, opened 2023-08-03
feat: restrict allowed sequence characters

SnickerDoodlePop:
The OTU DataLayer now restricts sequence creation / updating to only contain characters in [ATCGRYKM \n]

### DIFF
--- a/virtool/otus/data.py
+++ b/virtool/otus/data.py
@@ -2,6 +2,7 @@ import asyncio
 from copy import deepcopy
 from pathlib import Path
 from typing import Optional, Tuple, Mapping
+from pydantic import constr
 
 from pymongo.results import DeleteResult
 from virtool_core.models.enums import HistoryMethod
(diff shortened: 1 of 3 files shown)


SnickerDoodlePop:
Utils function `validate_sequence` was replaced with a functor `ValidateSequence` to take advantage of Pydantic `BaseModel`.

igboyes (changes requested):
Do this at the request handler by modifying the request model `CreateSequenceRequest`.

https://github.com/virtool/virtool/blob/4d3d505fcb83602a942f5535a606928b16469c00/virtool/otus/api.py#L363

https://github.com/virtool/virtool/blob/4d3d505fcb83602a942f5535a606928b16469c00/virtool/otus/oas.py#L35

igboyes (changes requested):
Apply requested changes in both locations.

igboyes (approved):

PR #2922, opened 2023-08-08
feat: allow `GroupsData.get` from pg or mongo

SnickerDoodlePop:


virtool/groups/data.py
```diff
@@ -47,17 +52,32 @@ async def find(self) -> List[GroupMinimal]:
             async for document in self._mongo.groups.find()
         ]
 
-    async def get(self, group_id: str) -> Group:
+    async def get(self, group_id: str | int) -> Group:
```
ReeceHoffmann, line 55: I think this function can be simplified. I would divide it into two main cases:

1. the id is an int -> Then you know the group must be on postgres as int ids mean nothing to mongo. Raise an error if no group is found
2. The id is a string -> then you need to check both DBs. I would rewrite so it fetches the group from both DBs at the same time along with the users list

tests/groups/test_data.py
```diff
@@ -34,6 +33,26 @@ async def test_create(
     assert doc["_id"] == group.id == row.legacy_id
 
 
+class TestGet:
```
ReeceHoffmann, line 36: Add cases for where no group can be found with both ID types

tests/groups/test_data.py
```diff
@@ -34,6 +33,26 @@ async def test_create(
...
+    async def test_id_is_str(self, groups_data: GroupsData, fake2, snapshot):
```
ReeceHoffmann, line 37:
```suggestion
    async def test_legacy_id(self, groups_data: GroupsData, fake2, snapshot):
```

tests/groups/test_data.py
```diff
@@ -34,6 +33,26 @@ async def test_create(
...
+        fake_group = await fake2.groups.create()
+
+        await fake2.users.create(groups=[fake_group])
+
+        group = await groups_data.get(fake_group.id)
+
+        assert group == snapshot
+
+    async def test_id_is_int(self, groups_data: GroupsData, fake2, snapshot):
```
ReeceHoffmann, line 46:
```suggestion
    async def test_single_user(self, groups_data: GroupsData, fake2, snapshot):
```
Since we are working towards postgres first support for groups an integer id should be treated as the default case

ReeceHoffmann (changes requested):

virtool/groups/data.py
```diff
@@ -47,19 +52,37 @@ async def find(self) -> List[GroupMinimal]:
...
+                id=pg_group.legacy_id,
+                name=pg_group.name,
+                permissions=Permissions(**pg_group.permissions),
+                users=await fetch_group_users(self._mongo, pg_group.legacy_id),
+            )
+
+        elif mongo_group:
+            return mongo_group
+
+        else:
```
ReeceHoffmann, line 84: Redundant else

virtool/groups/data.py
```diff
@@ -47,19 +52,37 @@ async def find(self) -> List[GroupMinimal]:
             async for document in self._mongo.groups.find()
         ]
 
-    async def get(self, group_id: str) -> Group:
+    async def get(self, group_id: str | int) -> Group:
```
ReeceHoffmann, line 55: Looking closer, I think the new logic for getting the group should all be in `fetch_complete_group`. As written functions which that rely on `fetch_complete_group` will fail to grab data from postgres. You will have to modify the signature of `fetch_complete_group` to include postgres. While doing so, please work on streamlining the logic. I think there is significant room to streamline and reduce the number of `if` statements to increase readability.

virtool/groups/data.py
```diff
@@ -47,19 +52,37 @@ async def find(self) -> List[GroupMinimal]:
...
-        if group:
-            return group
+        if type(group_id) is int:
+            pg_group = await get_row_by_id(self._pg, SQLGroup, group_id)
 
-        raise ResourceNotFoundError()
+            if pg_group is None:
+                raise ResourceNotFoundError()
+
+        elif type(group_id) is str:
```
ReeceHoffmann, line 69: Can be `else` b/c type should only ever be int or str

ReeceHoffmann (changes requested):

igboyes (approved):

PR #2927, opened 2023-08-11
fix: close non-checked-in-pg-connections

ChristineWC:


ChristineWC:
In _virtool/assets/alembic/env.py, tests/fixtures/migration.py_ and _tests/fixtures/pg.py_ there are other instances where `connect() `and `begin() `are used. However, data definition language (ddl) operations are used in those blocks. DDL operations are inherently synchronous and are context dependent. Interchanging those blocks with` async with AsyncSession(pg) as session:` in those instances results in errors about using DDL operations.

PR #2928, opened 2023-08-16
fix: move virtool http ws to virtool ws route

ChristineWC:


virtool/ws/route.py
```diff
@@ -1,24 +1,26 @@
-import logging
+"""
+This module manages Websocket connection requests.
```
ReeceHoffmann, line 2:
```suggestion
Provides handlers for managing Websocket related requests.
```

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #2929, opened 2023-08-21
fix: fix last run task task retrieval

ChristineWC:


ChristineWC:
**Bug:** In the test_register test, `assert task_spawner_service.registered[0].last_triggered == last_run_task.created_at` causes an assertion error because the last_run_task.created_at is earlier than the last_triggered time. 

The `find` method in TasksData uses a SQL query to retrieve the list of tasks. By default, this would return the results in the sequence they were initially added. However, it does not guarantee that the order will be the same every time. To guarantee that the results are returned in a specific order would require the order by clause.

igboyes (approved):

PR #2932, opened 2023-08-23
refactor: move analysis creation to data layer

ChristineWC:


igboyes (changes requested):
This issue is more complicated than it looks.

* Data layer methods need to return Pydantic data objects. The `create()` method should have the same return type as `get()`.
* There is no testing for this now that you deleted `test_create`.
* Is `str` still the right type for `workflow`?

tests/analyses/test_data.py
```diff
@@ -0,0 +1,59 @@
...
+from tests.users.test_data import users_data
+
+
+async def test_create(app, mongo, snapshot, static_time, users_data):
+    """
+    Tests that an analysis is created with the expected fields.
+    """
+    subtractions = ["subtraction_1", "subtraction_2"]
+
+    await users_data.create(password="hello_world", handle="bill")
```
igboyes, line 13: Use the `fake2` fixture to do this.

tests/analyses/test_data.py
```diff
@@ -0,0 +1,59 @@
...
+    user_info = await mongo.users.find_one({"handle": "bill"})
```
igboyes, line 14: A variable name ending with non-specifics like `info`, `data`, etc is a variable naming smell.

---

The full user object is returned when you create user with the data layer or `fake2`. No need to do another query directly to Mongo.

```py

user_id = user.id
```

tests/samples/test_api.py
```diff
@@ -930,15 +930,20 @@ async def test_analyze(
...
 
     if error != "400_reference":
-        await client.db.references.insert_one({"_id": "foo"})
+        await client.db.references.insert_one(
+            {
+                "_id": "foo",
+                "name": "foo",
+                "data_type": "genome",
+            }
+        )
```
igboyes, line 946: Please use realistic names as described for the data layer tests.

It makes it easier to scan over the snapshots if they change.

virtool/subtractions/db.py
```diff
@@ -182,3 +182,21 @@ def lookup_nested_subtractions(
             }
         },
     ]
+
+
+async def get_subtraction_names(subtraction_ids: List, db) -> List[dict]:
```
igboyes, line 187: 1. Let's upgrade the signature of this function.
     ```suggestion
     async def get_subtraction_names(mongo: "Mongo", subtraction_ids: List[str]) -> List[str]:
     ```

2. Refactor this function to be more efficient. It likely possible to use `Mongo.find` to massively improve perf. If not, `asyncio.gather()` is the less desirable option.
3. The docstring is not accurate.
4. The return type can be more specific.

igboyes (changes requested):

igboyes:
Will be resolved by @mattcurts in another PR.

PR #2934, opened 2023-08-24
refactor: remove job rights

SnickerDoodlePop:
>remove class definitions for `JobRights` and `JobRightsDomain`
>replace occurences of `JobRights` with `dict`
>update snapshots to reflect absent job rights


SnickerDoodlePop:
Deepsource is complaining about an async method being overridden as non-async; I am ignoring this.

virtool/http/client.py
```diff
@@ -63,7 +61,7 @@ def has_permission(self, permission: str) -> bool:
 
 
 class JobClient(AbstractClient):
-    def __init__(self, job_id, rights: JobRights):
+    def __init__(self, job_id, rights: dict):
```
igboyes, line 64: Is `rights` used for anything in `JobClient` now? It should be removed if not.

virtool/jobs/data.py
```diff
@@ -168,7 +163,7 @@ async def create(
         workflow: str,
         job_args: Document,
         user_id: str,
-        rights: JobRights,
+        rights: dict,
```
igboyes, line 166: Remove argument. It is always going to be an empty dict.

virtool/jobs/db.py
```diff
@@ -186,7 +186,7 @@ async def create_job(
         "workflow": workflow,
         "args": job_args,
         "key": None,
-        "rights": rights.as_dict(),
+        "rights": rights,
```
igboyes, line 189: Remove argument. It is never going to be a non-empty dict.
```suggestion
        "rights": {},
```

igboyes (changes requested):

igboyes (approved):

PR #2935, opened 2023-08-25
fix: return `UserNested` from `"/users"` endpoint

SnickerDoodlePop:


virtool/users/api.py
```diff
@@ -159,7 +151,7 @@ async def put(self, data: CreateFirstUserRequest) -> Union[r201[User], r400, r40
 @routes.view("/users/{user_id}")
 class UserView(PydanticView):
     @policy(AdministratorRoutePolicy(AdministratorRole.USERS))
-    async def get(self, user_id: str, /) -> Union[r200[User], r403, r404]:
+    async def get(self, user_id: str, /) -> Union[r200[UserNested], r403, r404]:
```
ReeceHoffmann, line 154: Should just be `User`.

virtool/users/api.py
```diff
@@ -93,14 +83,16 @@ async def post(self, data: CreateUserRequest) -> Union[r201[User], r400, r403]:
...
                 data.handle, data.password, data.force_reset
             )
         except ResourceConflictError as err:
             raise HTTPBadRequest(text=str(err))
 
         return json_response(
-            user,
+            UserNested(
+                id=user.id, administrator=user.administrator, handle=user.handle
+            ),
```
ReeceHoffmann, line 95: Should just be `User`

virtool/users/api.py
```diff
@@ -64,17 +56,15 @@ async def get(
             mongo_query,
             self.request.query,
             sort="handle",
-            projection=virtool.users.db.PROJECTION,
```
ReeceHoffmann, line 67: Please move DB access to the users data layer

ReeceHoffmann (changes requested):

PR #2940, opened 2023-09-07
test: update pytest-aiohttp 1.0.4 to 1.0.5

mattcurts:
Didn't seem to resolve many warnings, and warnings about loop being deprecated are still present.


igboyes (changes requested):
1. Change commit message to `test: update pytest-aiohttp 1.0.4 to 1.0.5`. We generally don't use the brackets feature, though I'd like to develop a convention for that. The words would be more general like `test(deps)` or `fix(compatibility)`.
2. Can you take a look at the GitHub issues for this package and see if there is a way to resolve the warnings? [I found this for starters](https://github.com/aio-libs/pytest-aiohttp/issues/13).

mattcurts:
Looked through the GitHub issues are couldn't find any way to resolve the issues without just silencing the warnings. It appears that the warnings are coming from other packages (based on the usage of the warning function).

I tried to update the relevant packages but that didn't have an impact, I also tried updating all the packages that we are using, and that still had no effect on eliminating the warnings. So I set the packages back to the versions they were using before.

The packages that appear to be causing the warning are:
aiohttp_devtools
asyncpg
uvloop
Screenshot from 2023-09-07 12-02-52

igboyes:
OK. We'll merge this anyway.

PR #2941, opened 2023-09-07
Matthewcurtis1999/vir 1855 update root handler with correct documentation links

mattcurts:
ensured current doc urls match the endpoint "docs" and all items on the side bar have an endpoint,


mattcurts:
also had to update the snapshot for tests/http/test_root.py to get test-unit CI to pass.

igboyes (approved):

PR #2943, opened 2023-09-11
Matthewcurtis1999/vir 1866 move virtoolsamplesdbfinalize to data layer

mattcurts:


virtool/samples/data.py
```diff
@@ -187,6 +191,64 @@ async def get(self, sample_id: str) -> Sample:
 
         return Sample(**document)
 
+    async def finalize(
+        self,
+        sample_id: str,
+        quality: Dict[str, Any],
+        _run_in_thread: callable,
+        data_path: Path,
+    ) -> Dict[str, Any]:
```
igboyes, line 200: Needs to return a `Sample`.

virtool/samples/api.py
```diff
@@ -273,16 +272,14 @@ async def finalize(req):
...
-        req.app["pg"],
+    await get_data_from_req(req).samples.finalize(
         sample_id,
         data["quality"],
         to_thread,
         get_config_from_req(req).data_path,
     )
 
-    return json_response(await virtool.samples.db.get_sample(req.app, sample_id))
+    return json_response(await get_data_from_req(req).samples.get(sample_id))
```
igboyes, line 282: This will look like the following after the change to `finalize()`.
```suggestion
    return json_response(sample)
```

virtool/samples/data.py
```diff
@@ -187,6 +191,64 @@ async def get(self, sample_id: str) -> Sample:
...
+        """
+        Finalize a sample document by setting a ``quality`` field
```
igboyes, line 202:
```suggestion
        Finalize a sample by setting a ``quality`` field
```

virtool/samples/data.py
```diff
@@ -187,6 +191,64 @@ async def get(self, sample_id: str) -> Sample:
...
+        and ``ready`` to ``True``
+
+        :param sample_id: the id of the sample
+        :param quality: a dict contains quality data
```
igboyes, line 206:
```suggestion
        :param quality: a dict containing quality data
```

virtool/samples/data.py
```diff
@@ -187,6 +191,64 @@ async def get(self, sample_id: str) -> Sample:
...
+        :param _run_in_thread: the application thread running function
```
igboyes, line 207: I'm not sure this is actually used. It should be removed if not.

virtool/samples/data.py
```diff
@@ -187,6 +191,64 @@ async def get(self, sample_id: str) -> Sample:
...
+        :param data_path: the application data path settings
```
igboyes, line 208: There may be a `self._config` attribute  which is the preferred way to access configuration.

igboyes (changes requested):

igboyes (changes requested):
You have set the desired return type for the `finalize()` method, but it doesn't actually return an object of that type.

You need to return the return value of `self.get()`. In your last set of commits the call to this method was lost completely.

PR #2945, opened 2023-09-12
Move analysis creation to data layer

mattcurts:
Continuing from the other PR.

Fixed up all the changes that were missing, and had to exclude created_at and updated_at time from snapshot for test_create.

Unsure if I should've updated check_rights on line 443 in samples/api.py to use a data layer function.

Otherwise I believe that the division of responsibility between the data layer and the API is being respected.


mattcurts:
I was not able to generate the logs to test it, due to issues with installing HMMs, but I believe I solved the issue.

PR #2949, opened 2023-09-15
feat!: remove unpaginated upload search

mattcurts:
Did I need to have a BREAKING CHANGE Footer or is the ! enough?


igboyes:
> Did I need to have a BREAKING CHANGE Footer or is the ! enough?

I have always used the footer, but let's do this an see how it looks in the release notes.

@ReeceHoffmann Is this going to break UI?

ReeceHoffmann:
It will break the UI. The main `File` views will handle the change just fine, but the sample and subtraction views currently expect full unpaginated data. I can make the updates a priority on our end

igboyes:
@ReeceHoffmann The ML changes are the priority. You can give this to Rastin if you want. I'll let this PR stew if it can't be merged now.

PR #2952, opened 2023-09-19
Matthewcurtis1999/vir 1860 adopt structlog

mattcurts:
Changed all logging in virtool over to structlog.
I have removed coloredlogs (using logging basicConfig).
we can get it printing without color if we want.
the pathname shows what context variables will look like when passed into log.info("message",x=1).

image

Further changes would be refactoring the log messages to pull out context variables wherever possible.
We can also adjust the length of padding given for the event message.


igboyes:
I am doing this in a different branch using @mattcurts changes to core.

PR #2954, opened 2023-09-21
Matthewcurtis1999/vir 1852 setup mypy

mattcurts:
opened the PR early to test that CI is working as expected.


mattcurts:
canceled tests to avoid running tests that are not needed to run right away. 
CI Task now works the same as local.

igboyes (approved):

PR #2956, opened 2023-09-26
Extend syrupy to assert valid ISO times

mattcurts:
opening early to check if the updating breaks things.


mattcurts:
didn't mean to close but newer PR is here: https://github.com/virtool/virtool/pull/2957.

PR #2958, opened 2023-10-03
Matthewcurtis1999/vir 1872 extend syrupy to assert valid iso times

mattcurts:
Set up the template for the extension.
To use the extension you need to pass and call snapshot_recent instead of snapshot.

### DIFF
--- a/tests/fixtures/snapshot_date.py
+++ b/tests/fixtures/snapshot_date.py
@@ -1,0 +1,28 @@
+import datetime
+from typing import Any
+
+import pytest
+from syrupy.matchers import path_type
+
+
+def parse_time(data):
+    if not isinstance(data, datetime.datetime):
+        raise ValueError("timestamp in wrong format")
+
+    if datetime.datetime.utcnow() - data < datetime.timedelta(seconds=30):
+        return "approximately_now"
+    else:
+        return "not_approximately_now"
+
+
+@pytest.fixture
+def snapshot_recent(snapshot):
+    return snapshot.with_defaults(
+        matcher=path_type(
+            mapping={
+                "created_at": (datetime.datetime, Any),
+                "uploaded_at": (datetime.datetime, Any),
+            },
+            replacer=lambda data, _: parse_time(data),
+        )
+    )
--- a/tests/fixtures/snapshot_date.py
+++ b/tests/fixtures/snapshot_date.py
@@ -14,8 +14,7 @@ def validate_time(timestamp: datetime.datetime | Any):
 
     if datetime.datetime.utcnow() - timestamp < datetime.timedelta(seconds=30):
         return "approximately_now"
-    else:
-        return "not_approximately_now"
+    return "not_approximately_now"
 
 
 @pytest.fixture
--- a/tests/fixtures/snapshot_date.py
+++ b/tests/fixtures/snapshot_date.py
@@ -43,11 +43,12 @@ def snapshot_recent(snapshot):
     return snapshot.with_defaults(
         matcher=path_type(
             mapping={
-                "applied_at": (datetime.datetime, str, Any),
-                "created_at": (datetime.datetime, str, Any),
-                "updated_at": (datetime.datetime, str, Any),
-                "uploaded_at": (datetime.datetime, str, Any),
+                ".*applied_at": (datetime.datetime, str, Any),
+                ".*created_at": (datetime.datetime, str, Any),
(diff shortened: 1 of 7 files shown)


tests/fixtures/snapshot_date.py
```diff
@@ -0,0 +1,32 @@
+import datetime
+from typing import Any
+
+import pytest
+from syrupy.matchers import path_type
+
+
+def validate_time(timestamp: datetime.datetime | Any):
```
igboyes, line 8: Why are we accepting `Any` here if a `ValueError` is thrown if the value is not a `datetime`?

The raised exception should actually be `TypeError` and a more accurate message is 'Timestamp must be datetime'. Format would be more appropriate for a ISO time `str`.

tests/fixtures/snapshot_date.py
```diff
@@ -0,0 +1,32 @@
...
+        raise ValueError("timestamp in wrong format")
+
+    if datetime.datetime.utcnow() - timestamp < datetime.timedelta(seconds=30):
+        return "approximately_now"
+    return "not_approximately_now"
+
+
+@pytest.fixture
+def snapshot_recent(snapshot):
+    return snapshot.with_defaults(
```
igboyes, line 22: * Extend this so that if a `str` is found for the path, we try to parse it as an ISO formatted time using Arrow. If this fails, we can throw a `ValueError`.

  The representation of the time should show up in the snapshot value. This will allow us to avoid a ISO string getting put in MongoDB instead of a `datetime`. For example, `approximately_now_isoformat` or `approximately_now_datetime`.

  If the value is not a `datetime` or `str`, we should throw `TypeError`. I think it's best to limit the scope of this fixture to time representations we actually use.
* Add docstring describing usage that is suitable for guiding devs using this fixture.

tests/fixtures/snapshot_date.py
```diff
@@ -0,0 +1,32 @@
...
+        matcher=path_type(
+            mapping={
+                "created_at": (datetime.datetime, Any),
+                "uploaded_at": (datetime.datetime, Any),
+                "applied_at": (datetime.datetime, Any),
+                "updated_at": (datetime.datetime, Any),
```
igboyes, line 28: Sorry to nitpick, but sort alphabetically so it is easy to scan for potential future growth.

tests/test_fixtures.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def test_time_recent():
+    timestamp = datetime.datetime.utcnow()
+    assert validate_time(timestamp) == "approximately_now"
+
+
+def test_time_not_recent():
+    timestamp = datetime.datetime.utcnow() - datetime.timedelta(minutes=1)
+    assert validate_time(timestamp) == "not_approximately_now"
```
igboyes, line 21: * Test the actual `snapshot_recent` fixture.
* Once you update things to support ISO formatted strings, make sure these tests cover ISO vs incorrect format and datetime vs str type.

tests/fixtures/snapshot_date.py
igboyes, line 8: You may be able to avoid the extra `lambda` below with:
```suggestion
def validate_time(timestamp: datetime.datetime | Any, _):
```

tests/fixtures/snapshot_date.py
```diff
@@ -0,0 +1,32 @@
...
+            },
+            replacer=lambda timestamp, _: validate_time(timestamp),
```
igboyes, line 30: With previous comment, this should be possible:
```suggestion
            replacer=validate_time,
```

igboyes (changes requested):
A few enhancements and tweaks needed. Thanks for figuring this out!

tests/fixtures/snapshot_date.py
mattcurts: Had to default set it to None

mattcurts:
Current issue: it doesn't work when the timestamp is nested in another dictionary.

example:
        "example": {
            "id": 1,
            "active": True,
            "color": "red",
            "message": "Administrative instance message",
            "created_at": "2021-11-24T19:40:03.320000Z",
            "updated_at": "2021-11-24T19:40:03.320000Z",
            "user": {"id": "ian", "handle": "ianboyes", "administrator": True},
        }
    }
    This would not change the timestamps.

mattcurts:
I can't find a good solution to the timestamp nesting problem I mentioned above

igboyes (approved):

PR #2959, opened 2023-10-06
fix: correct latest_release to be non None when possible

mattcurts:


virtool/ml/data.py
```diff
@@ -67,7 +67,18 @@ async def list(self) -> MLModelListResult:
...
+                        created_at=one.releases[0].created_at,
+                        download_url=one.releases[0].download_url,
+                        github_url=one.releases[0].github_url,
+                        name=one.releases[0].name,
+                        published_at=one.releases[0].published_at,
+                        ready=one.releases[0].ready,
+                        size=one.releases[0].size,
+                    )
+                    if one.releases
+                    else None,
```
igboyes, line 81: I think this can be done in the SQL query instead of pulling all the releases for each model and getting the first.

https://stackoverflow.com/a/66116924

igboyes (changes requested):

igboyes (approved):

PR #2961, opened 2023-10-10
Matthewcurtis1999/vir 1873 make error shapes consistent between web and job apis

mattcurts:
added a snapshot test so we know if the fix breaks.

### DIFF
--- a/tests/jobs/test_api.py
+++ b/tests/jobs/test_api.py
@@ -19,7 +19,6 @@ _job_response_matcher = path_type(
 class TestFind:
     async def test_basic(
         self,
-        data_layer: DataLayer,
         fake2: DataFaker,
         snapshot,
         spawn_client: ClientSpawner,
@@ -294,14 +293,13 @@ class TestPing:
             seconds=1
         )
 
-    async def test_not_found(self, snapshot, spawn_client: ClientSpawner):
+    async def test_not_found(self, spawn_job_client):
         """Test that a 404 is returned when the job doesn't exist."""
-        client = await spawn_client(authenticated=True)
+        client = await spawn_job_client(authorize=True)
 
-        resp = await client.put("/jobs/foo/ping", {})
+        resp = await client.put("/jobs/foo/ping", data={})
 
         assert resp.status == 404
-        assert await resp.json() == snapshot
 
 
 @pytest.mark.apitest
--- a/tests/jobs/test_api.py
+++ b/tests/jobs/test_api.py
@@ -294,17 +294,14 @@ class TestPing:
             seconds=1
         )
 
-    async def test_not_found(self, spawn_client: ClientSpawner):
+    async def test_not_found(self, snapshot, spawn_client: ClientSpawner):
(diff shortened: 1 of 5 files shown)


virtool/api/response.py
```diff
@@ -88,13 +88,20 @@ async def error_middleware(req: web.Request, handler: Callable):
     :param req: the incoming request
     :param handler: the next middleware to be executed
     """
+    logger = logging.getLogger()
     try:
+        logger.error("BEFORE HANDLE")
         return await handler(req)
 
     except web.HTTPException as exc:
+        logger.error("CAUGHT ERROR")
```
igboyes, line 97: * Don't leave these dev logs in your PR.
* We always instantiate `logger` at the top of the module.

virtool/api/response.py
```diff
@@ -88,13 +88,20 @@ async def error_middleware(req: web.Request, handler: Callable):
...
         data = {"id": "_".join(exc.reason.lower().split(" ")), "message": exc.text}
 
+        if exc.reason == "Not Found":
+            data["message"] = "Not found"
+
         if isinstance(exc, (InvalidQuery, InvalidInput)):
             data["errors"] = exc.errors
 
+        logger.error(data)
```
igboyes, line 106: Not appropriate for PR. This will be useless log spam in production.

virtool/api/response.py
```diff
@@ -88,13 +88,20 @@ async def error_middleware(req: web.Request, handler: Callable):
...
+    logger = logging.getLogger()
     try:
+        logger.error("BEFORE HANDLE")
         return await handler(req)
 
     except web.HTTPException as exc:
+        logger.error("CAUGHT ERROR")
         data = {"id": "_".join(exc.reason.lower().split(" ")), "message": exc.text}
 
+        if exc.reason == "Not Found":
```
igboyes, line 100: Please add a comment to say why this is being done. I can see it being opaque in the future.

tests/jobs/test_api.py
```diff
@@ -293,17 +294,14 @@ async def test_ok(self, fake2: DataFaker, spawn_job_client):
             seconds=1
         )
 
-    async def test_not_found(self, spawn_client: ClientSpawner):
+    async def test_not_found(self, snapshot, spawn_client: ClientSpawner):
```
igboyes, line 297: This is muddling things a bit.

This test should use `spawn_job_client`. This is a mistake in how the test was written and is the reason `404: Not Found` is being returned in `message`. The path is not actually configured in the router because `/jobs/:id/ping` is only available on the jobs API. This `404` is coming from the router instead of the request handler.

Please revert changes to this test and change `spawn_client` to `spawn_job_client`.

igboyes (changes requested):
* `logger` can be removed. What you are logging is not desirable in production.
* Add a test at `tests/api/test_response.py` to check that the desired response body is received for a request to static path that isn't configured in the router (eg. `/foobar`). You are not effectively testing that your change works.

igboyes (approved):

PR #2968, opened 2023-10-13
Matthewcurtis1999/vir 1880 move recalculate workflow tags to data layer

mattcurts:
I'm confused about this objective: 

Emit the sample update event using the emit decorator instead of emits.


igboyes:
This turned out to not be a great idea. Closing.

PR #2973, opened 2023-10-16
feat: start writing users to postgres

igboyes:


tests/groups/test_mongo.py
```diff
@@ -17,6 +17,11 @@ async def test_update_member_users_and_api_keys(
     user = await fake2.users.create(groups=[group_1, group_2])
 
     async with AsyncSession(pg) as session:
+        await session.execute(
+            delete(user_group_associations).where(
+                user_group_associations.c.group_id == group_2.id
+            )
+        )
         await session.execute(delete(SQLGroup).where(SQLGroup.id == group_2.id))
```
igboyes, line 25: I think there is a way to make it so when the delete the group, the rows in the assocation table are automatically removed a well.

It may just work out of the box with SQLAlchemy.

Otherwise, take a look at cascading (`CASCADE`).

virtool/users/pg.py
```diff
@@ -28,7 +28,7 @@ class SQLUser(Base):
     b2c_given_name: Mapped[str] = ""
     b2c_family_name: Mapped[str] = ""
     b2c_oid: Mapped[str] = ""
-    force_reset: Mapped[bool] = False
+    force_reset: Mapped[bool] = Column(Boolean, default=False)
```
igboyes, line 31: Was the old line not working? I am a bit surprised. I thought it was valid new-style model definition using `Mapped`. It would like to figure out why it doesn't work.

This is work a try, but I don't want to do it without being clear on why a simple `= False` is not working for the default value.
```suggestion
    force_reset: Mapped[bool] = mapped_column(default=False)
```

virtool/groups/data.py
```diff
@@ -150,6 +151,20 @@ async def delete(self, group_id: int):
...
+                )
+            )
+
+            if user_associations:
+                # If the group is associated with any users, remove the associations
+                await pg_session.execute(
+                    delete(user_group_associations).where(
+                        user_group_associations.c.group_id == group_id
+                    )
+                )
```
igboyes, line 167: Again, this can probably be super-simplified using `CASCADE`.

Generally, Postgres/SQL is a lot more features to make our lives easier when compared with Mongo. I don't know them all. If you find yourself manually joining or filtering data in code or writing a lot of logic like this block, check if there are better ways to do it using features of Postgres.

On top of that, SQLAlchemy may have helpful functionality for making it easier to use these features of Postgres.

tests/users/test_data.py
```diff
@@ -188,37 +198,41 @@ async def test_administrator(
...
+                user.id, UpdateUserRequest(force_reset=True)
+            )
+        async with (AsyncSession(pg) as session):
+            row = await session.get(SQLUser, 1)
+            assert user.force_reset is True
+            assert row.force_reset is True
+            assert await get_one_field(mongo.users, "force_reset", user.id) is True
+            user = await data_layer.users.update(
+                user.id, UpdateUserRequest(force_reset=False)
+            )
```
igboyes, line 312: Move this back one indent. It should not be in the session context manager.

tests/users/test_data.py
```diff
@@ -188,37 +198,41 @@ async def test_administrator(
...
-            user.id, UpdateUserRequest(force_reset=False)
-        )
-
-        assert user.force_reset is False
-        assert await get_one_field(mongo.users, "force_reset", user.id) is False
+            assert row.force_reset is False
+            assert user.force_reset is False
+            user = await data_layer.users.update(
+                user.id, UpdateUserRequest(force_reset=True)
+            )
```
igboyes, line 304: Please use whitespace to separate concepts to some extent. One line below a `with` block is a good place to start.
```suggestion
            )

```

tests/users/test_data.py
```diff
@@ -236,13 +250,27 @@ async def test_set_groups(
             exclude=props("password"),
             matcher=_last_password_change_matcher,
         )
+        async with (AsyncSession(pg) as session):
+            groups = await session.execute(
+                select(user_group_associations)
+                .join(SQLUser)
+                .where(user_group_associations.c.user_id == SQLUser.id)
```
igboyes, line 346: I feel like this should be done the other way around.
```select(SQLUser).where(SQLUser.id == user_id).join(...)```

igboyes (commented):
Thanks.

I think the Postgres parts can be simplified and made more robust by exploiting the features of the database.

I don't think the association should be directly queried during tests. It makes more sense to test the query against the users table as that is what actually happens in the wild.

tests/groups/test_mongo.py
mattcurts: Fixed by adding ondelete to the foreign key

virtool/users/pg.py
mattcurts: The old line wasn't working for test_force_reset as the SQLUser did not store force reset without it being a column or mapped_column. See picture below, and note that all b2c is not stored in the SQLUser as well.

The Mapped[bool] is only a type hint that is really only useful to the reader, unless it gets checked with mypy or other static analysis tools.

The mapped_column does work to pass the test and not fail any other existing test.

image

igboyes:
Superseded by #2976.

PR #2976, opened 2023-10-31
feat: start writing users to postgres

mattcurts:
created own PR to allow for review functions to work better

The old line wasn't working for test_force_reset as the SQLUser did not store force reset without it being a column or mapped_column. See picture below, and note that all b2c is not stored in the SQLUser as well.

image

The Mapped[bool] is only a type hint that is really only useful to the reader, unless it gets checked with mypy or other static analysis tools.

The mapped_column does work to pass the test and not fail any other existing test.

Did you want me to make all columns mapped?

### DIFF
--- a/tests/groups/test_data.py
+++ b/tests/groups/test_data.py
@@ -1,5 +1,6 @@
 import pytest
-from sqlalchemy.ext.asyncio import AsyncEngine
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
 from virtool_core.models.group import Permissions
 
 from virtool.data.errors import ResourceConflictError, ResourceNotFoundError
@@ -8,6 +9,8 @@ from virtool.fake.next import DataFaker
 from virtool.groups.oas import UpdateGroupRequest
 from virtool.groups.pg import SQLGroup
 from virtool.pg.utils import get_row_by_id
+from virtool.users.oas import UpdateUserRequest
+from virtool.users.pg import SQLUserGroup
 from virtool.users.utils import generate_base_permissions
 
 
@@ -108,17 +111,33 @@ async def test_update_permissions(
 
 
 class TestDelete:
-    async def test_ok(
-        self,
-        data_layer: DataLayer,
-        pg: AsyncEngine,
-    ):
+    async def test_ok(self, data_layer: DataLayer, pg: AsyncEngine, fake2):
         """Test that deletion of a group removes it from both databases."""
+        user = await fake2.users.create()
         group = await data_layer.groups.create("Test")
+        await data_layer.users.update(user.id, UpdateUserRequest(groups=[group.id]))
+
+        async with AsyncSession(pg) as session:
+            user_associations = (
+                await session.execute(
+                    select(SQLUserGroup).where(SQLUserGroup.group_id == group.id)
+                )
+            ).all()
+
+        assert len(user_associations) == 1
 
         await data_layer.groups.delete(group.id)
 
         assert await get_row_by_id(pg, SQLGroup, group.id) is None
+
+        async with AsyncSession(pg) as session:
+            users = (
+                await session.execute(
+                    select(SQLUserGroup).where(SQLUserGroup.group_id == group.id)
+                )
+            ).all()
+
+        assert len(users) == 0
 
     async def test_not_found(self, data_layer: DataLayer):
         """Test that a ResourceNotFoundError is raised when the group does not exist."""
--- a/tests/references/__snapshots__/test_api.ambr
+++ b/tests/references/__snapshots__/test_api.ambr
@@ -1920,13 +1920,13 @@
   dict({
     'body': '''
       The release consists of a gzipped JSON file containing:
-      
+
       - a `data_type` field with value _genome_
       - an `organism` field with value _virus_
       - the `version` name (eg. *v0.2.0*)
       - a timestamp with the key `created_at`
       - virus data compatible for import into Virtool v2.0.0+
-      
+
       Scripts have been updated to follow upcoming convention changes in Virtool v3.0.0.
     ''',
     'content_type': 'application/gzip',
--- a/tests/users/__snapshots__/test_data.ambr
+++ b/tests/users/__snapshots__/test_data.ambr
@@ -1,8 +1,8 @@
 # serializer version: 1
-# name: TestCreate.test_force_reset[False]
+# name: TestCreate.test_first
   dict({
     'active': True,
-    'administrator_role': None,
+    'administrator_role': <AdministratorRole.FULL: 'full'>,
     'b2c': None,
     'b2c_display_name': None,
     'b2c_family_name': None,
@@ -12,7 +12,7 @@
     'groups': list([
     ]),
     'handle': 'bill',
-    'last_password_change': datetime,
+    'last_password_change': 'approximately_now_datetime',
     'permissions': dict({
       'cancel_job': False,
       'create_ref': False,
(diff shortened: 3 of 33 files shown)


mattcurts:
How do you want me to handle the high  cyclomatic complexity that is causing deepsource to fail?

tests/references/__snapshots__/test_api.ambr
```diff
@@ -616,297 +616,6 @@
...
   })
 # ---
-# name: TestCreateOTU.test[True--True][db]
-  list([
-    dict({
-      '_id': 'bf1b993c',
-      'abbreviation': '',
-      'isolates': list([
-      ]),
-      'last_indexed_version': None,
```
ReeceHoffmann, line 626: What's going on here? Seem unrelated to the PR

tests/users/test_api.py
```diff
@@ -334,7 +334,7 @@ async def test_create_first_user(
     client = await spawn_client()
 
     if not first_user_exists:
-        await mongo.users.delete_many({})
+        await data_layer.users.delete_all()
```
ReeceHoffmann, line 337: imo I would leave this as it was. I can't see a practical use for `delete_all` in production and triggering it by accident would be catastrophic

virtool/users/data.py
```diff
@@ -161,39 +170,98 @@ async def create(
...
+                    b2c_display_name=b2c_user_attributes.display_name
+                    if b2c_user_attributes
+                    else "",
+                    b2c_given_name=b2c_user_attributes.given_name
+                    if b2c_user_attributes
+                    else "",
+                    b2c_family_name=b2c_user_attributes.family_name
+                    if b2c_user_attributes
+                    else "",
+                    b2c_oid=b2c_user_attributes.oid if b2c_user_attributes else "",
```
ReeceHoffmann, line 204: Can this be handled more simply with a single `if` or short circuit evaluation?

virtool/users/data.py
```diff
@@ -161,39 +170,98 @@ async def create(
...
+            await self._mongo.users.delete_one({"_id": legacy_id}, session=mongo)
+
+    async def delete_all(self):
+        """
+        Delete all users.
+         :return:
+        """
+        async with both_transactions(self._mongo, self._pg) as (mongo, pg):
+            await pg.execute(delete(SQLUser))
+            await self._mongo.users.delete_many({}, session=mongo)
```
ReeceHoffmann, line 228: Is this used elsewhere in the codebase? This seems like real landmine

virtool/users/data.py
```diff
@@ -161,39 +170,98 @@ async def create(
...
+            )
+
         return await self.get(document["_id"])
 
+    async def delete(self, legacy_id: str):
+        """
+        Delete the user with the matching legacy_id.
+
+         :param legacy_id: legacy id of the user to delete
+         :return:
```
ReeceHoffmann, line 215: Nitpick: is the left hand justification of this docstring correct? looks off to me but could be correct

virtool/users/data.py
```diff
@@ -161,39 +170,98 @@ async def create(
...
+                session=mongo,
+            )
+            user = SQLUser(
+                legacy_id=document["_id"],
+                handle=handle,
+                password=document["password"],
+                force_reset=False,
+                last_password_change=now,
             )
+            pg.add(user)
```
ReeceHoffmann, line 262: nitpick: I think you only use `user` here once so I would just define it as it is used

virtool/users/data.py
```diff
@@ -161,39 +170,98 @@ async def create(
...
             document = await create_user(
-                self._mongo, handle, password, False, session=session
+                self._mongo, handle, password, False, session=mongo
             )
-
             await self._mongo.users.update_one(
                 {"_id": document["_id"]},
                 {"$set": {"administrator": True}},
-                session=session,
+                session=mongo,
```
ReeceHoffmann, line 253: Won't this be covered by your call to `set_administrator_role` below?

virtool/users/data.py
```diff
@@ -292,91 +354,120 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
+                try:
+                    updates.update(
+                        await compose_groups_update(self._pg, data["groups"])
+                    )
+                    for _id in data["groups"]:
+                        await pg.execute(
+                            insert(user_group_associations).values(
+                                user_id=user.id, group_id=_id
+                            )
+                        )
```
ReeceHoffmann, line 419: Can this be streamlined to a single request? Also what happens in the case that a user is no longer part of a group?

virtool/users/data.py
```diff
@@ -292,91 +354,120 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
                         self._mongo,
                         self._pg,
                         data.get("groups", []),
                         data["primary_group"],
                         user_id,
                     )
+                    await pg.execute(
+                        insert(user_group_associations).values(
+                            user_id=user.id, group_id=data["primary_group"]
+                        )
```
ReeceHoffmann, line 436: What happens to the users previous primary group?

ReeceHoffmann (changes requested):
Overall looks like a good start. I probably didn't catch everything, so expect a second round of review. If you haven't please also take a good look through the codebase for any other modules that might be modifying the mongo `users` table.

tests/users/test_api.py
mattcurts: needed to also delete postgres Users, so I have just moved delete all into this function as it was the only place it was called.

virtool/users/data.py
mattcurts: streamlined. I neglected that, but I have added the removal of groups.

virtool/users/data.py
mattcurts: updated and now getting removed from table

tests/references/__snapshots__/test_api.ambr
mattcurts: I guess I accidentally updated this snapshot, I have undone this change.

mattcurts:
I've split update into two (update and update_legacy), as I noticed that in test it will work fine due to everything being created in both mongo and pg, but in production I would've needed a check to ensure ``user is not None`` before every update else an error would be raised.

This way once we have done the transition, we should be able to deprecate update_legacy with minimal change to the existing code.

ReeceHoffmann:
I'm gonna take a better look through tomorrow, but is the GH version up to date with your local? The two update functions look intermingled to me still. Also naming thing, I would specify `mongo` instead of legacy since that's really what it is doing

mattcurts:
yeah GH is up to date with local. The legacy one is only called if a user can't be found for the id, and it's the same code as the normal one just without any user postgres stuff to deal with. I will update the name maybe ``update_mongo_only``?

virtool/users/data.py
```diff
@@ -292,91 +362,265 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
+                        .where(
+                            user_group_associations.c.group_id == user.primary_group_id
+                        )
+                    )
+
+                    await pg_session.execute(
+                        insert(user_group_associations).values(
+                            user_id=user.id, group_id=data["primary_group"]
+                        )
+                    )
```
ReeceHoffmann, line 590: I believe correct behaviour is to error if the user is not currently or won't be a member of the group selected as the primary group

virtool/users/data.py
```diff
@@ -292,91 +362,265 @@ async def update(self, user_id: str, data: UpdateUserRequest):
...
+                user.invalidate_sessions = True
+
+            if "groups" in data:
+                try:
+                    updates.update(
+                        await compose_groups_update(self._pg, data["groups"])
+                    )
+
+                    await pg_session.execute(
+                        delete(user_group_associations)
```
ReeceHoffmann, line 539: You should be able to manage the the groups associated via SQLAlchemy User model ORM. Clearing all the associations and rewriting them seems clunky

ReeceHoffmann (changes requested):

virtool/users/data.py
```diff
@@ -161,39 +171,103 @@ async def create(
...
+        async with both_transactions(self._mongo, self._pg) as (
+            mongo_session,
+            pg_session,
+        ):
+            await pg_session.execute(
+                delete(SQLUser).where(SQLUser.legacy_id == legacy_id)
+            )
+            await self._mongo.users.delete_one(
+                {"_id": legacy_id}, session=mongo_session
+            )
```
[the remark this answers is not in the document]
mattcurts: Ian wanted me to "Update the user create, update, and delete methods, as well as any other methods that affect users.  For now, we will not be modifying how user data is read." but there wasn't a delete before, so I added it. But I have now removed it again because I see the issues with it.

virtool/users/data.py
```diff
@@ -252,37 +320,39 @@ async def set_administrator_role(
...
+        async with both_transactions(self._mongo, self._pg) as (
+            mongo_session,
+            pg_session,
+        ):
+            await update_legacy_administrator(self._mongo, user_id, role, mongo_session)
+            await pg_session.execute(
+                update(SQLUser)
+                .where(SQLUser.legacy_id == user_id)
+                .values(administrator=role == AdministratorRole.FULL)
+            )
```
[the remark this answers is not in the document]
mattcurts: I thought we wanted to keep logic out of db files unless it was legacy? Updated to used gather

virtool/users/data.py
mattcurts: I'm confused , this is the update function so primary group may be being changed?

mattcurts: nvm understood now

virtool/users/data.py
```diff
@@ -214,26 +274,18 @@ async def find_or_create_b2c_user(
...
-        handle = "-".join(
-            [
-                b2c_user_attributes.given_name,
-                b2c_user_attributes.family_name,
-                str(random.randint(1, 100)),
-            ]
+        handle = await generate_handle(
+            self._mongo.users,
+            b2c_user_attributes.given_name,
+            b2c_user_attributes.family_name,
```
ReeceHoffmann, line 280: nitpick: I'd move this into `create_b2c` since you only need it for a new user

virtool/users/data.py
ReeceHoffmann: Did you have a chance to try this via ORM?

ReeceHoffmann (changes requested):

tests/users/__snapshots__/test_data.ambr
```diff
@@ -493,6 +626,18 @@
     'primary_group': None,
   })
 # ---
+# name: TestUpdate.test_set_groups[pg_1]
+  <SQLUser(id='1', legacy_id='bf1b993c', active='True', administrator='False', b2c_display_name='', b2c_given_name='', b2c_family_name='', b2c_oid='', force_reset='False', handle='qtaylor', invalidate_sessions='False', last_password_change='datetime', primary_group_id='None')>
```
ReeceHoffmann, line 630: Should these models include the `groups` field as well?

ReeceHoffmann (changes requested):
I think this is super close, comments are mostly small tweaks. That said, can you verify if deleting a group fully correctly removes the association from the table? might need to tweak the groups model or `data_layer` to ensure this happens

tests/users/test_data.py
```diff
@@ -120,37 +199,44 @@ async def test_administrator(
...
+        assert row.force_reset is False
 
     async def test_set_groups(
         self,
         data_layer: DataLayer,
         fake2: DataFaker,
-        mongo: Mongo,
+        pg: AsyncEngine,
         snapshot: SnapshotAssertion,
     ):
```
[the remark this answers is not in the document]
mattcurts: datalayer.user.update() calls self.get() at the end of the function so calling datalayer.users.get() after updating would return the same as the already present return value from datalayer.user.update().

I will work on parameterizing it to reduce the logic.

tests/users/__snapshots__/test_data.ambr
mattcurts: SQLUser lazy loads the groups relationship, so groups will not show up in the snapshots

tests/groups/test_data.py
```diff
@@ -120,6 +122,20 @@ async def test_ok(
...
+                    await session.execute(
+                        select(user_group_associations).where(
+                            user_group_associations.c.group_id == group.id
+                        )
+                    )
+                )
+                .scalars()
+                .all()
+            )
+        assert len(users) == 0
```
ReeceHoffmann, line 137: Won't this be true before and after deletion? with no associated users there should never be an entry in this table

tests/users/test_data.py
```diff
@@ -13,49 +15,75 @@
...
+
+
+@pytest.fixture
+async def groups(groups_param, fake2):
+    groups = []
+    for _ in range(groups_param):
+        groups.append(await fake2.groups.create())
+    return groups
+
+
```
ReeceHoffmann, line 41: What's the advantage over `@parameterise`?

tests/users/test_data.py
```diff
@@ -120,115 +212,117 @@ async def test_administrator(
...
         snapshot: SnapshotAssertion,
+        groups,
     ):
         """
         Test that setting ``groups`` works as expected.
         """
-        group_1 = await fake2.groups.create()
-        group_2 = await fake2.groups.create()
-
         user = await fake2.users.create()
```
ReeceHoffmann, line 260: I would be tempted always assign a group to the user to start. That way you can test removal of groups as well as addition

ReeceHoffmann (dismissed):

tests/users/test_data.py
mattcurts: misunderstood how I had to get the 0,1, or 2 groups. Condensed into 1 function now. but we need this to simulate the actual groups vs just group id's.

ReeceHoffmann: Is there a blocker on doing it in the main test function? something like:
`groups = await asyncio.gather([fake2.groups.create() for group in range(groups_count) ])`

tests/groups/test_data.py
mattcurts: updated by adding user and group to user before deletion.

tests/users/test_data.py
```diff
@@ -13,49 +15,70 @@
...
     {"last_password_change": (datetime.datetime,)}
 )
 
 
+@pytest.fixture(params=[0, 1, 2])
+async def groups(fake2, request):
+    groups = []
+    for _ in range(request.param):
+        groups.append(await fake2.groups.create())
+    return groups
```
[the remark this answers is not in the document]
mattcurts: I put it in a fixture to make it easier to count the number of groups but it better to probably inline the logic.

igboyes (dismissed):

virtool/users/db.py
```diff
@@ -78,3 +81,100 @@ async def compose_groups_update(
         raise ResourceConflictError(f"Non-existent groups: {', '.join(repr_ids)}")
 
     return {"groups": group_ids}
+
+
+async def update_mongo_user(
```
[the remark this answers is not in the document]
mattcurts: @ReeceHoffmann recommended it to reduce the complexity of the update function, given we eventually will be no longer on mongo.

virtool/users/data.py
```diff
@@ -252,33 +307,38 @@ async def set_administrator_role(
...
+            mongo_session,
+            pg_session,
+        ):
+            await asyncio.gather(
+                update_legacy_administrator(self._mongo, user_id, role, mongo_session),
+                pg_session.execute(
+                    update(SQLUser)
+                    .where(SQLUser.legacy_id == user_id)
+                    .values(administrator=role == AdministratorRole.FULL)
+                ),
```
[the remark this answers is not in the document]
ReeceHoffmann: Frontend only use administratorRoles at this point, made the transition a while ago.

virtool/users/db.py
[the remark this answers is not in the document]
igboyes: We will leave this as is, seeing as it will soon be deleted.

This pattern for building updates comes up frequently. You are going to have to write dual support for MongoDB + Postgres in the future. Follow this pattern:

* Check `if field in update request`
  * Add to Mongo `update`.
  * Update Postgres model object.
* Check `if field in update request`
  * Add to Mongo `update`.
  * Update Postgres model object.
* Check `if field in update request`
  * Add to Mongo `update`.
  * Update Postgres model object.
* Commit changes
  * Call `update_one` for Mongo.
  * Nothing required for Postgres if you are in `both_transactions`. Commit will happen when context manager closes.

This lets us see the parallel actions happening across the two databases. Everything happening is right in front of our faces.

---

Here is the current form:

* Check `if field in update request`
  * Update Postgres model object.
* Check `if field in update request`
  * Update Postgres model object.
* Check `if field in update request`
  * Update Postgres model object.
* Commit changes
  * Call another function that copies exactly the above series of conditionals but does the Mongo-flavored actions. There is also a giant interface for this function that not only takes the Mongo and Postgres objects, but the entire users data object.
  * Nothing required for Postgres if you are in `both_transactions`. Commit will happen when context manager closes.

The one argument for this implementation is you can just delete the Mongo function and its call when Mongo support is dropped. I don't think that justifies the indirection and addition of another function.

I don't want to see the users data object being passed around by dependency injection. Right now we have largely maintained a nice layered structure:

API -> Data -> direct database access

Functions that do direct database access only need references to the database client objects (`mongo`, `pg`). In `update_mongo_user`, we are not only taking the database clients, but the entire user data object. This muddles our hierarchy and pushes us back into spaghetti territory.

Hope this makes sense.

virtool/users/data.py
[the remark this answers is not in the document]
igboyes: Thanks @ReeceHoffmann. Then there are two PRs / issues to be tackled:

1. We just need to make sure `user.administrator` is not used in any logic in the backend. Drop all use of `user.administrator` or the `administrator` field from Mongo. Can be done with a complete survey of all text matches to "administrator".
2. We need to make sure all users with `administrator == true`  have the corresponding roles in OpenFGA. I think this was done in the "promote administrators" task. This should be dropped and replaced with a migration.

Merge these, then this PR can be rebased onto `main`.

mattcurts:
good for review, this is the only thing left that I believe needs fixing https://github.com/virtool/virtool/pull/2976/#discussion_r1406782146

PR #2978, opened 2023-11-07
Matthewcurtis1999/vir 1902 include jobminimal with list of analysis

mattcurts:


virtool/analyses/data.py
```diff
@@ -69,83 +69,88 @@ def __init__(self, db: Mongo, config, pg: AsyncEngine):
...
+                        "user": True,
+                    },
+                    "total_count": {"$arrayElemAt": ["$total_count.total_count", 0]},
+                    "found_count": {"$arrayElemAt": ["$found_count.found_count", 0]},
+                }
+            },
+        ]
+
+        if sample_id is not None:
+            pipeline.insert(0, {"$match": {"sample.id": sample_id}})
```
ReeceHoffmann, line 128: 1. I would avoid injecting a step like this, harder to debug and a nightmare if you have more than one conditional
2. I think `total_count` should probably still reflect the total number of analyses

virtool/analyses/data.py
```diff
@@ -69,83 +69,88 @@ def __init__(self, db: Mongo, config, pg: AsyncEngine):
...
-                    self._mongo, document["sample"]["id"], client, write=False
+                virtool.samples.db.check_rights_error_check(
+                    self._mongo,
+                    sample_id if sample_id is not None else document["sample"]["id"],
+                    client,
+                    write=False,
                 )
                 for document in data
             ]
         )
```
ReeceHoffmann, line 145: Probably worth a refactor, no need to send so many requests if you already know the `sample_id` is consistent.

ReeceHoffmann (changes requested):
Close, but I think `sample_id` should only have an impact on `found_count` and not on `total_count`.

virtool/analyses/data.py
```diff
@@ -69,86 +69,106 @@ def __init__(self, db: Mongo, config, pg: AsyncEngine):
...
         documents = await apply_transforms(
             [base_processor(d) for d in documents],
             [AttachJobTransform(self._mongo), AttachUserTransform(self._mongo)],
         )
 
+        print(sample_id)
+        print(data)
+        print(per_document_can_read)
+        print(documents)
+
```
ReeceHoffmann, line 171:
```suggestion

```

virtool/analyses/data.py
```diff
@@ -69,86 +69,106 @@ def __init__(self, db: Mongo, config, pg: AsyncEngine):
...
+                virtool.samples.db.check_rights_error_check(
+                    self._mongo,
+                    sample_id,
+                    client,
+                    write=False,
+                )
+            ]
+            found_count = sum(
+                document["sample"]["id"] == sample_id for document in data
+            )
```
ReeceHoffmann, line 143: Filtering needs to be done as part of the pipeline to get accurate results

virtool/analyses/data.py
```diff
@@ -69,86 +69,106 @@ def __init__(self, db: Mongo, config, pg: AsyncEngine):
...
+        else:
+            can_read = [
+                virtool.samples.db.check_rights_error_check(
```
ReeceHoffmann, line 146: A permission error raised here is a problem for clients who haven't asked for a specific sample_id.

ReeceHoffmann (changes requested):

tests/samples/__snapshots__/test_api.ambr
```diff
@@ -3789,79 +3783,20 @@
...
-            'name': 'Malus domestica',
-          }),
-        ]),
-        'user': dict({
-          'administrator': False,
-          'handle': 'zclark',
-          'id': '7cf872dc',
-        }),
-        'workflow': 'pathoscope_bowtie',
-      }),
```
ReeceHoffmann, line 3858: This seems suspicious, you sure excluding these is correct?

igboyes (approved):

PR #2981, opened 2023-11-22
Matthewcurtis1999/vir 1908 add post route for uploading analysis files

mattcurts:


tests/analyses/test_api.py
```diff
@@ -27,6 +27,13 @@ def files(test_files_path, tmp_path):
     return data
 
 
+@pytest.fixture
+def more_files(test_files_path, tmp_path):
```
igboyes, line 31: This does the exact same thing as `files`. Get this down to single fixture. The best solution might be to return a function from the fixture.

igboyes (approved):

PR #2987, opened 2023-12-08
Matthewcurtis1999/vir 1904 attributeerror nonetype object has no attribute pop

mattcurts:
Added return type hints to all functions that have @emits, and I think the issue is that ``get_json()`` which gets assigned to ``received`` is a shortcut for ``get()`` is either the message received or None signifying that the channel has been unsubscribed and no more messages will be received.

Added a logging message and then subscribe to the events channel before continuing in the loop, waiting for the next event.

https://aioredis.readthedocs.io/en/v1.3.0/api_reference.html


mattcurts:
no clue why CI is flipping the order of test snapshot

virtool/data/events.py
```diff
@@ -203,6 +203,13 @@ async def __anext__(self) -> Event:
             try:
                 received = await self._channel.get_json()
 
+                if received is None:
+                    logger.error(
+                        "Channel has been unsubscribed, trying to re-subscribe"
+                    )
+                    (self._channel,) = await self._redis.subscribe(self._channel_name)
+                    continue
+
```
igboyes, line 212: Remove this. We are already handling `ChannelClosedError` more rigourously below.

This problem here is that `None` is being put in the queue. Update the `emit` function such that it avoids ever putting `None` in the queue. Log a warning if it is attempted so we can fix the problem.

igboyes (changes requested):
The `emit` function needs to avoid ever putting `None` in the queue. Update this function so it logs a warning and does nothing if `None` is received.

virtool/data/events.py
```diff
@@ -92,6 +92,10 @@ def emit(data: BaseModel, domain: str, name: str, operation: Operation):
 
 
     """
+
+    if data is None:
+        logger.warning("emit event with no data")
+
```
igboyes, line 98: Need an `else` here so that nothing is emitted in warning case.

igboyes (changes requested):

igboyes (approved):

PR #3000, opened 2024-01-02
Fix analysis tests

mattcurts:


mattcurts:
had to re-add the total counting as the pipeline does not count ones that don't have matching sample_id and total count would be wrong

virtool/analyses/data.py
```diff
@@ -96,7 +96,7 @@ async def find(
             {"$match": {"sample.id": sample_id} if sample_id is not None else {}},
             {
                 "$facet": {
-                    "total_count": [{"$count": "total_count"}],
```
igboyes, line 99: Please don't change how this is implemented versus `main`.

There may have been some confusion as a fix was merged to remediate very slow responses for this method in production. It turns out that `count_documents({})` is extremely slow for a collection of 1000s of large documents.

virtool/analyses/data.py
```diff
@@ -146,21 +146,24 @@ async def find(
...
+            apply_transforms(
+                [base_processor(d) for d in documents],
+                [
+                    AttachJobTransform(self._mongo),
+                    AttachReferenceTransform(self._mongo),
+                    AttachSubtractionsTransform(self._mongo),
+                    AttachUserTransform(self._mongo),
+                ],
+            ),
+            self._mongo.analyses.count_documents({}),
```
igboyes, line 159: This needs to go. Rely on the aggregation as in `main`.

igboyes (changes requested):

igboyes (approved):

PR #3001, opened 2024-01-02
Matthewcurtis1999/vir 1905 dont finalize an already finalized sample

mattcurts:


tests/samples/test_data.py
```diff
@@ -126,6 +174,26 @@ async def test_finalize(
...
+        "length": [3237],
+        "sequences": [7091],
+    }
+
+    await data_layer.samples.finalize("test", quality)
+
+    try:
+        await data_layer.samples.finalize("test", quality)
+    except ResourceConflictError:
+        pass
```
igboyes, line 197: Use `pytest.raises`.

igboyes (changes requested):
Please closely review changes to snapshots.

mattcurts:
The Artifacts, Reads, and subtractions missing are intentional, as I am using a simpler fixture that doesn't have those attached and has ``ready`` set to false to ensure that we can test that it gets updated and that we are stopped when trying to finalize it when ``ready`` is true.

igboyes (approved):

PR #3008, opened 2024-01-09
Matthewcurtis1999/vir 1635 use python 311 or 312 with virtool

mattcurts:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10-buster as build
+FROM python:3.12-rc-buster as build
```
igboyes, line 1: I believe `rc` here means "release candidate". Is there not a full release available?

bowtie2-inspect
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
```
igboyes, line 1: Please move this file to `assets`.

tests/analyses/test_api.py
```diff
@@ -382,14 +382,11 @@ async def test_remove(
...
         )
 
-    m_remove = mocker.patch("virtool_core.utils.rm")
-
     resp = await client.delete("/analyses/foobar")
 
     match error:
         case None:
             await resp_is.no_content(resp)
-            assert m_remove.called_with("data/samples/baz/analyses/foobar", True)
```
igboyes, line 392: This is eliminating part of the test. Am I missing something?

tests/references/test_api.py
```diff
@@ -669,7 +664,7 @@ async def test_update(
         assert await resp.json() == snapshot(
             name="json", matcher=path_type({".*etag": (str,)}, regex=True)
         )
-        assert m_enqueue.called_with(1)
+
```
igboyes, line 667: This is eliminating part of the test. Am I missing something?

tests/analyses/test_api.py
mattcurts: the mock was never being called as checked with m_remove.call_count being 0 and m_remove.call_args (the way to check arguments passed to magic mocks) being None. Since the mock was never used, I removed the code to clean it up.

PR #3014, opened 2024-01-23
Matthewcurtis1999/vir 1948 resolve datetime deprecation warning

mattcurts:
if we wanted to keep utc, then we would have to make sure all timestamps have the timezone UTC attached.

### DIFF
--- a/tests/fixtures/snapshot_date.py
+++ b/tests/fixtures/snapshot_date.py
@@ -21,14 +21,14 @@ def validate_time(timestamp: datetime.datetime | str | Any, _=None):
 
     """
     if isinstance(timestamp, datetime.datetime):
-        if datetime.datetime.now(datetime.timezone.utc) - timestamp < datetime.timedelta(seconds=30):
+        if arrow.utcnow().naive - timestamp < datetime.timedelta(seconds=30):
             return "approximately_now_datetime"
         return "not_approximately_now_datetime"
     if isinstance(timestamp, str):
         try:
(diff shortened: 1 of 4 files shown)


tests/fixtures/snapshot_date.py
```diff
@@ -21,13 +21,13 @@ def validate_time(timestamp: datetime.datetime | str | Any, _=None):
 
     """
     if isinstance(timestamp, datetime.datetime):
-        if datetime.datetime.utcnow() - timestamp < datetime.timedelta(seconds=30):
+        if datetime.datetime.now() - timestamp < datetime.timedelta(seconds=30):
```
igboyes, line 24: I am bit confused about this.

The `datetime.now()` method returns a naive `datetime` object with the local time. The method `utcnow()` returns a naive `datetime` object in UTC time.

The timestamp value should be a naive datetime in UTC, so how is it possible that it is only 30 seconds different from a naive datetime in local time.

I am assuming you read this: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

My expectation was that the code about should look like:
```py
if datetime.datetime.now(timezone.utc) - timestamp < datetime.timedelta(seconds=30):
    ...
```
The only explanation I can think of is that `timestamp` is in local time. It shouldn't be.

mattcurts: changed from datime.now to arrow.utcnow().naivie to keep it in UTC and Naive

igboyes (approved):

PR #3023, opened 2024-02-20
Matthewcurtis1999/vir 1927 drop the mongo attribute on spawned test clients

mattcurts:
chasing some last bugs


mattcurts:
I had to change the ``assert_called_with`` because there were 2 different mongo objects, the one created in the api/datalayer that is normally used and the one created by the test function. ANY allows us to focus the concern on the other parameters that we wanted to ensure the functions were called with.

igboyes (changes requested):
This is good, but a bunch of unwanted extra indentation has been added. Please fix.

igboyes (approved):

PR #3024, opened 2024-02-23
Restore reference updates

ReeceHoffmann:
Fixes a bug preventing reference updates from working.


igboyes (approved):
This change was made because

PR #3025, opened 2024-02-23
fix: add retry attempts before erroring

mattcurts:


virtool/data/events.py
```diff
@@ -51,7 +52,16 @@ class _InternalEventsTarget:
...
 
     def emit(self, event: Event):
-        self.q.put_nowait(event)
+        retries = 3
+        for attempt in range(retries-1):
+            try:
+                self.q.put_nowait(event)
+                return
+            except asyncio.QueueFull:
+                sleep(5)
```
igboyes, line 61: Must use `asyncio.sleep` to avoid blocking the application

virtool/data/events.py
```diff
@@ -51,7 +52,16 @@ class _InternalEventsTarget:
...
+                continue
```
igboyes, line 62: Doesn't have any effect.

virtool/data/events.py
```diff
@@ -51,7 +52,16 @@ class _InternalEventsTarget:
     q = asyncio.Queue(maxsize=1000)
 
     def emit(self, event: Event):
-        self.q.put_nowait(event)
+        retries = 3
+        for attempt in range(retries-1):
```
igboyes, line 56: * Unnecessary variable `retries`.
* Unused variable `attempt`.
* Current code only tries two times, whereas I believe we discussed three tries.

```suggestion
        for _ in range(3):
```

igboyes (changes requested):

igboyes (approved):

PR #3026, opened 2024-02-26
Prevent error when setting sample rights

ReeceHoffmann:
Changes:
  -  No longer requires setting the group when changing sample rights to prevent a 500 error
  - Adds check that the sample rights API works as expected

### DIFF
--- a/tests/samples/test_api.py
+++ b/tests/samples/test_api.py
@@ -1352,3 +1352,99 @@ async def test_download_artifact(error, tmp_path, spawn_job_client, pg):
     assert (
         get_config_from_app(client.app).data_path / "samples" / "foo" / "fastqc.txt"
     ).read_bytes() == await resp.content.read()
+
+
+class TestChangeSampleRights:
+    async def test_update_group_id(
+        self,
+        fake2,
+        get_sample_data,
+        mongo,
+        snapshot,
+        spawn_client,
+    ):
+        group = await fake2.groups.create()
+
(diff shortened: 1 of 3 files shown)


tests/samples/test_api.py
```diff
@@ -1352,3 +1352,42 @@ async def test_download_artifact(error, tmp_path, spawn_job_client, pg):
     assert (
         get_config_from_app(client.app).data_path / "samples" / "foo" / "fastqc.txt"
     ).read_bytes() == await resp.content.read()
+
+
+@pytest.mark.apitest
```
igboyes, line 1357: We're getting rid of this since the test suite is running in one job again.

tests/samples/test_api.py
```diff
@@ -1352,3 +1352,42 @@ async def test_download_artifact(error, tmp_path, spawn_job_client, pg):
...
+    "update_group, group_settings, all_settings",
+    [
+        (True, False, False),
+        (False, True, False),
+        (False, False, True),
+        (True, True, True),
+    ],
+    ids=["group", "group_settings", "all_settings", "all"],
+)
+async def test_change_rights(
```
igboyes, line 1368: Please change this to a class-based test with methods for the four cases.

I know I have used parametrization a lot previously, but I have discovered that the tests are difficult to grok and update after writing.

igboyes (changes requested):

igboyes (approved):

PR #3028, opened 2024-02-26
Include jobs field in samples

ReeceHoffmann:
Changes:
 - Changes the `data.samples.find` to include a complete job as part of the search results
 - Updates `find` tests as well as other tests to check for proper job inclusion


igboyes (approved):
Very nice!

PR #3052, opened 2024-05-09
fix: determine image version at build time

ReeceHoffmann:
## Changes

- remove real time version determination from git
- Image version now determined during docker image build.
- remove CI for setting the version during publishing

## Compatibility

Sources of breaking changes:

### API

### Migration

### Configuration

 New version breaks compatibility with dev.

### Services

  OpenFGA.

### Notes

Compatibility with the current version of `dev` is broken due to needing to access `.git` at build time. 
Production images will work as expected with no breaking changes.

PR fix: https://github.com/virtool/dev/pull/29

## Pre-Review Checklist
* [x] All changes are tested.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-tags: "true"
+          fetch-depth: 0
       - name: Login to Registry
         uses: docker/login-action@v3
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,6 +23,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-tags: "true"
       - name: Login to Registry
         uses: docker/login-action@v3
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,8 +23,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Version
-        run: echo ${{ github.event.release.tag_name }} > VERSION
       - name: Login to Registry
         uses: docker/login-action@v3
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,6 +25,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          ref: github.event.release.tag_name
       - name: Login to Registry
         uses: docker/login-action@v3
         with:
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,8 @@ RUN poetry install --without dev --no-root && rm -rf "$POETRY_CACHE_DIR"
 
 FROM python:3.12-bookworm as version
 COPY .git .
-RUN git describe --tags --abbrev=0 #This line is for testing ONLY remove before merge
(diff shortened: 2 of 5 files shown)


tests/test_version.py
```diff
@@ -1,27 +1,11 @@
...
-    if source == "git":
-        mocker.patch(
-            "subprocess.check_output", return_value=bytes("1.0.13", encoding="utf-8")
-        )
-    else:
-        mocker.patch(
-            "subprocess.check_output",
-            side_effect=subprocess.CalledProcessError(1, "none"),
-        )
+import virtool.version
```
igboyes, line 3:
```
from virtool.version import determine_server_version
```

Dockerfile
```diff
@@ -9,13 +9,18 @@ WORKDIR /app
 COPY poetry.lock pyproject.toml ./
 RUN poetry install --without dev --no-root && rm -rf "$POETRY_CACHE_DIR"
 
+FROM python:3.12-bookworm as version
+COPY .git .
+RUN git describe --tags --abbrev=0 > VERSION
```
igboyes, line 14: What happens when the current HEAD doesn't have a tag? 25.0.1-abc123?, where abc123 is short commit hash?

.github/workflows/publish.yml
```diff
@@ -23,8 +23,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Version
-        run: echo ${{ github.event.release.tag_name }} > VERSION
+        with:
+          fetch-depth: 0
```
igboyes, line 27: I just realized we should be providing the `tag_name` as the value for the `ref` option for the [checkout action](https://github.com/actions/checkout). It is currently possible to release a build based on a commit after the `tag_name`.

Correct me if I'm wrong.

ReeceHoffmann: Hmmm yes I think so. Since checkout will always grab the latest version (afaik) it is possible that if another release commit gets merged between when the release was triggered and actually building the container. The result there would be the `VERSION` and container files being the updated version. The `VERSION` file would I think still be accurate to the actual contents, but the containers release tag could be misleading.

I assume there must be a way to check out the exact commit version, will investigate and get back to you. Pretty sure it would be no big deal to edit the dockerfile to use either the git commit history or the GH event tag

Dockerfile
ReeceHoffmann: So it depends, if only the current version has been fetch and it lacks a tag then the checkout fails since it needs to find a tag somewhere in the history. Otherwise it will just grab the most recent tag. 

When on a version with a tag, it is just the current tag. `eg. 25.0.1` 

When the current HEAD doesn't have a tag, but one exist earlier in the commit history, it will give the most recent tag currently w/o the short hash. `eg. 25.0.1`  Removing  `--abbrev=0` makes it something like `24.1.0-6-g58b98e072` which matches the style for a pre-release [semver rule 9.](https://semver.org/#spec-item-9) but this may end up appearing in release builds.

The best option for including short hashes might be to go back to using the release tag from GH for release builds, then use this in a non-release context

ReeceHoffmann:
Converting to draft while a test CI changes in the build step

Dockerfile
ReeceHoffmann: Updated:  The tag for non-head commits is now formatted as: `24.1.0-pre+g58b98e072`

igboyes (changes requested):

igboyes (approved):

PR #3065, opened 2024-05-24
Convert ensureindexfilestask to migration

ReeceHoffmann:
## Changes

- Move functionality of `EnsureFileIndexTask` into a migration
   - Replicates functionality of EnsureFileIndex in new migration re-using much of the previous code
   - updates test to work with
   - Adds `data_path` to the migration context object. 
      - CLI options already existed for getting the data path but were previously ignored
- Removes usage of `EnsureFileIndexTask` from data faker 

## Compatibility

Sources of breaking changes:

### API

### Migration

- [ ] My changes do not require a newer migration than the currently required migration..

### Configuration

- [ ] My changes do not change configuration value names or types.

**NOTE:** The data path variable in the CLI was previously ignored. No new CLI options were technically added, 
but in practice new arguments will need to be passed if the default option for data storage is not correct
 - Dev already passes the data_path as a env variable

### Services

  OpenFGA.

### Notes

## Pre-Review Checklist
* [x] All changes are tested.


igboyes (changes requested):
Change typing to use literals (eg. `Dict` -> `dict`).

Looks good otherwise.

PR #3066, opened 2024-05-28
Reecehoffmann/vir 2004 set subtractions field in analysis documents

ReeceHoffmann:
~~**NOTE:** This PR is a draft until it can be updated to work with the other migration in review~~ Resolved

## Changes

 - Adds a migration to create the `deleted` field on subtraction documents
    - Adds test to ensure that `deleted` is set to false if and only if no  `deleted` field exists on that subtraction

## Compatibility

Sources of breaking changes:

### API

### Migration

### Configuration

### Services

  OpenFGA.

### Notes

## Pre-Review Checklist
* [x] All changes are tested.


igboyes (changes requested):
Naming is a bit wonky. Need to change branch name I think.

PR #3080, opened 2024-06-06
refactor: rename fake2 fixture to fake

Bryce-Davidson:
## Changes

- Rename fake2 fixture to fake 

## Compatibility

Sources of breaking changes:

### API

### Migration

### Configuration

### Services

  OpenFGA.

### Notes

## Pre-Review Checklist
* [x] All changes are tested.

### DIFF
--- a/tests/account/test_api.py
+++ b/tests/account/test_api.py
@@ -128,7 +128,7 @@ async def test_update_settings(data, status, spawn_client, resp_is, snapshot):
 
 
 async def test_get_api_keys(
-    fake2: DataFaker,
+        fake: DataFaker,
     mongo: Mongo,
     spawn_client: ClientSpawner,
     snapshot,
@@ -136,7 +136,7 @@ async def test_get_api_keys(
 ):
     client = await spawn_client(authenticated=True)
 
-    group = await fake2.groups.create()
+    group = await fake.groups.create()
 
     await mongo.keys.insert_many(
         [
@@ -177,7 +177,7 @@ class TestCreateAPIKey:
         has_perm,
         req_perm,
         data_layer: DataLayer,
-        fake2: DataFaker,
+            fake: DataFaker,
         mocker,
         snapshot,
         spawn_client: ClientSpawner,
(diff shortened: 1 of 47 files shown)


tests/account/test_api.py
```diff
@@ -128,15 +128,15 @@ async def test_update_settings(data, status, spawn_client, resp_is, snapshot):
 
 
 async def test_get_api_keys(
-    fake2: DataFaker,
+        fake: DataFaker,
```
igboyes, line 131: There is something weird going on with the indentation here. This is not expected from Ruff.
I am guessing this could be a side-effect of a find-replace.

tests/groups/test_data.py
```diff
@@ -144,31 +143,31 @@ async def test_update_permissions(
...
     assert group.permissions == Permissions(
-        **{**generate_base_permissions(), "modify_subtraction": True}
+        **{**generate_base_permissions(), "modify_subtraction": True},
     )
     assert await get_row_by_id(pg, SQLGroup, group.id) == snapshot(name="pg_removed")
 
 
 class TestDelete:
-    async def test_ok(self, data_layer: DataLayer, pg: AsyncEngine, fake2):
+    async def test_ok(self, data_layer: DataLayer, pg: AsyncEngine, fake):
```
igboyes, line 161: Please add typing where it is missing, in files you touch.
```suggestion
    async def test_ok(self, data_layer: DataLayer, pg: AsyncEngine, fake: DataFaker):
```

igboyes (changes requested):
Very minor issues. Thanks.

tests/groups/test_data.py
Bryce-Davidson: Trying to do this elegantly is taking a while. Pycharm refactor only takes care of rename. No way to ensure import of type in all files.

Might need to manual it in the morning. 
Or get better at grep and bash

igboyes: Don't worry about it. Best effort when you see it while editing. 

Let's get this merged

igboyes (approved):

PR #3082, opened 2024-06-12
Bryce davidson/vir 1985 remove create fake subtraction functions

Bryce-Davidson:
## Changes

- removed direct use of mongo for /dev route fake subtraction

## Compatibility

Sources of breaking changes:

### API

### Migration

### Configuration

### Services

  OpenFGA.

### Notes

## Pre-Review Checklist
* [x] All changes are tested.


virtool/subtractions/fake.py
```diff
@@ -24,41 +21,27 @@ async def create_fake_fasta_upload(app: App, user_id: str) -> Tuple[int, str]:
...
-    upload_name: str,
-    subtraction_id: str,
-    user_id: str,
-):
+) -> None:
+    """Create a finalized subtraction in the database and file system.
+
+    :param app: the application object
+    """
+    layer = get_data_from_app(app)
```
Bryce-Davidson, line 34: Could this procedure be it's own `create_data_faker` function?

Bryce-Davidson:
Could also remove that function entirely as it has no other incoming calls

igboyes:
> Could also remove that function entirely as it has no other incoming calls

Yes. Please remove those functions. The whole module should be able to go.

igboyes (approved):

PR #3085, opened 2024-06-17
feat: allow getting client IP from configurable header

ReeceHoffmann:
## Changes

 - Adds the `real_ip_header` config option to `ServerOptions`
   - Option is available for both the regular API and jobsAPI
   - If the value is not specified the `req.remote` is logged. Otherwise the contents or lack therefore of is logged

## Compatibility

Sources of breaking changes:

### API

### Migration

### Configuration

- [ ] My changes do not change configuration value names or types.

Changes **do** change configuration options by adding `real_ip_header`. If left blank the server will continue to use `req.remote` for logging the clients ip. 

### Services

  OpenFGA.

### Notes

## Pre-Review Checklist
* [ ] All changes are tested. - We don't currently have a framework for testing our middlewares. Happy to work on it if determined to be a priority. Change has been manually tested


virtool/config/options.py
```diff
@@ -129,6 +129,13 @@ def b2c_options(func):
     is_flag=True,
 )
 
+real_ip_header_option = click.option(
+    "--real_ip_header",
```
igboyes, line 133:
```suggestion
    "--real-ip-header",
```

igboyes (changes requested):

igboyes (approved):

PR #3091, opened 2024-06-21
Reecehoffmann/vir 2009 migrate sample files

ReeceHoffmann:
## Changes

- Add a migrate sample files migration which:
   - Will determine sample legacy status if undetermined and write that to the DB
   - Compress uncompressed legacy sample files
   - migrate the sample file field over to postgres
 - Removes tasks, helpers and related tests that handled this job before

## Pre-Review Checklist
* [x] All changes are tested.

### DIFF
--- a/assets/revisions/rev_lcq797n5ryxk_migrate_sample_files.py
+++ b/assets/revisions/rev_lcq797n5ryxk_migrate_sample_files.py
@@ -135,9 +135,6 @@ async def compress_sample_reads(
 
 async def upgrade(ctx: MigrationContext):
     async for sample in ctx.mongo.samples.find({"files": {"$exists": True}}):
-        files = sample["files"]
-        sample["files"] = [files] if isinstance(files, dict) else files
-
         if "is_legacy" not in sample:
             sample["is_legacy"] = check_is_legacy(sample)
             ctx.mongo.samples.update_one(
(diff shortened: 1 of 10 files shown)


assets/revisions/rev_lcq797n5ryxk_migrate_sample_files.py
```diff
@@ -0,0 +1,334 @@
...
+        return [
+            sample_path / "reads_1.fastq",
+            sample_path / "reads_2.fastq",
+        ]
+
+    return [sample_path / "reads_1.fastq"]
+
+
+async def compress_sample_reads(
+    db: AsyncIOMotorDatabase,
```
igboyes, line 85:
```suggestion
    motor_database: AsyncIOMotorDatabase,
```

assets/revisions/rev_lcq797n5ryxk_migrate_sample_files.py
```diff
@@ -0,0 +1,334 @@
...
+    :return: whether the sample read files are compressed
+    """
+    files = sample.get("files")
+
+    if not files or not sample.get("is_legacy") or sample.get("is_compressed"):
+        return True
+
+    names = [file["name"] for file in files]
+
+    return all(name in {"reads_1.fq.gz", "reads_2.fq.gz"} for name in names)
```
igboyes, line 62: Can be combined pretty easily.

assets/revisions/rev_lcq797n5ryxk_migrate_sample_files.py
```diff
@@ -0,0 +1,334 @@
...
+    )
+
+
+def check_is_compressed(sample: dict[str, any]) -> bool:
+    """Check if a sample has compressed read files.
+
+    :param sample: the sample document
+    :return: whether the sample read files are compressed
+    """
+    files = sample.get("files")
```
igboyes, line 55: This shouldn't be tripped because of the pre-existing Mongo query.

assets/revisions/rev_lcq797n5ryxk_migrate_sample_files.py
```diff
@@ -0,0 +1,334 @@
...
+            )
+
+        if not check_is_compressed(sample):
+            sample["files"] = await compress_sample_reads(
+                ctx.mongo,
+                ctx.data_path,
+                sample["files"],
+                sample["paired"],
+                sample["_id"],
+            )
```
igboyes, line 152: I think the Mongo changes here should also happen in the below Mongo transaction.

igboyes (changes requested):
Overall, I think the code can be simplified and maybe number of functions reduced compared to what it was originally.

assets/revisions/rev_lcq797n5ryxk_migrate_sample_files.py
ReeceHoffmann: I think this probably makes sense, but I need to be careful with getting this right. The key advantage to the way it is done now is that the files will never be deleted before the new `file` location is saved to a database. With merging them I potentially lose track of which samples have been compressed during this migration and which were already compressed. That said, I think blindly attempting to delete the uncompressed reads for each sample shouldn't cause any problems

ReeceHoffmann:
Review has been addressed. One notable change to the assumptions made is that the migration will now write `is_compressed: True` to all mongodb records that are processed. (before `is_compressed` only got written to records where the files were at some point not compressed) Given that `is_compressed` has never been consistently written to all records that are compressed, there is probably a good argument for stripping it out completely though.

igboyes (approved):

PR #3096, opened 2024-07-02
Migrate legacy jobs to latest data structure

ReeceHoffmann:
## Changes

- Migrates legacy jobs to new data structure
  - Adds field `key` with value `None` if the field is undefined
  - Adds field `acquired` with value `False` if the field is undefined
  - Adds field `archived` with value `False` if the field is undefined
  - renames field `task` to `workflow` if the `task` field exists
  - Ensures that all status progression is represented as integers
    - runs if for any given status: (`state=complete` AND `progress= 1`) OR `type(progress) == double`
    - If the above checks are met for any status, all status updated to have their progress multiplied by 100 and truncated to an integer.

## Pre-Review Checklist
* [x] All changes are tested.


assets/revisions/rev_u4tep9xyu4hc_update_job_documents.py
```diff
@@ -23,8 +23,14 @@
 
 async def upgrade(ctx: MigrationContext):
     ctx.mongo.jobs.update_many({"key": {"$exists": False}}, {"$set": {"key": None}})
-    ctx.mongo.jobs.update_many({"acquired": None}, {"$set": {"acquired": False}})
-    ctx.mongo.jobs.update_many({"archived": None}, {"$set": {"archived": False}})
+    ctx.mongo.jobs.update_many(
+        {"acquired": {"$exists": False}},
+        {"$set": {"acquired": False}},
+    )
```
igboyes, line 29: I think this should be set to `True`.

Any jobs that don't have this field will have been completed in the past, which also means they were acquired.

igboyes (changes requested):

igboyes (approved):

PR #3097, opened 2024-07-03
Bryce davidson/vir 1986 use subtraction faking

Bryce-Davidson:
## Pre-Review Checklist
* [x] All changes are tested.


tests/subtractions/test_db.py
```diff
@@ -1,61 +1,83 @@
...
-    await mongo.subtraction.insert_many(
-        [{"_id": "foo", "name": "Foo"}, {"_id": "bar", "name": "Bar"}], session=None
-    )
+    ]
+
+    for document in documents:
+        result = await apply_transforms(document, [AttachSubtractionsTransform(mongo)])
 
-    result = await apply_transforms(documents, [AttachSubtractionsTransform(mongo)])
+        assert result == snapshot
```
igboyes, line 36: The way this was before was faster.

igboyes (changes requested):

tests/subtractions/test_db.py
Bryce-Davidson: Not sure how to pass the created subtraction Id to the parameterized test. I guess I should just write a separate function to pass to the `pytest.mark.parametrize` so the tests can be parallelized?

Bryce-Davidson: Do you want me to break this up into multiple tests? There is no straight forward way to get code execution within the `pytest.mark.parametrize` that works nicely with the new fake API.

igboyes: This test isn't parametrized is it?

I am just saying the last few lines can look like this instead:
```py
assert await apply_transforms(document, [AttachSubtractionsTransform(mongo)]) == snapshot
```

The `apply_transforms` function can take either single document or list of documents as its first argument.

tests/subtractions/test_api.py
```diff
@@ -144,361 +108,555 @@ async def test_get_from_job(fake: DataFaker, spawn_job_client, snapshot_recent):
...
+        test_dir: Path = tmp_path / "files"
+        test_dir.mkdir(exist_ok=True)
+
+        file_directory: Path = test_dir
+        subtraction_directory: Path = tmp_path / "subtractions"
+
+        return self.Context(
+            file_directory=file_directory,
+            subtraction_directory=subtraction_directory,
         )
```
igboyes, line 213: This is only used in `test_create` and doesn't need to be a fixture.

---
In fact, I don't think it needs to exist at all.

When you use `spawn_job_client` an application server is created with it own temporary data directory (same as the `--data-path` configuration value used when starting the server.

The `data_path` is accessible using the eponymous fixture. This means the following can be simplified:
```py
# Initial
assert os.listdir(
    ctx.subtraction_directory / subtraction.id,
) == [
    self.VALID_SUBTRACTION_FILE_NAME,
]

# With ``data_path`` fixture.
assert os.listdir(data_path / "subtractions" / subtraction.id) == [
    self.VALID_SUBTRACTION_FILE_NAME,
]
```

Let me know if I am missing some nuance in this case.

tests/subtractions/test_api.py
```diff
@@ -144,361 +108,555 @@ async def test_get_from_job(fake: DataFaker, spawn_job_client, snapshot_recent):
...
+        - test_subtraction_file_invalid_name_error: Ensures proper handling of invalid file names.
+        - test_subtraction_file_name_conflict_error: Verifies handling of file name conflicts.
+    """
+
+    VALID_SUBTRACTION_FILE_NAME = "subtraction.1.bt2"
+    INVALID_SUBTRACTION_FILE_NAME = "reference.1.bt2"
+
+    @dataclass
+    class Context:
+        file_directory: Path
```
igboyes, line 196: No used as far as I can tell.

tests/subtractions/test_api.py
```diff
@@ -144,361 +108,555 @@ async def test_get_from_job(fake: DataFaker, spawn_job_client, snapshot_recent):
...
+        user = await fake.users.create()
+        upload = await fake.uploads.create(
+            user=user,
+            upload_type=UploadType.subtraction,
+        )
+        subtraction = await fake.subtractions.create(user=user, upload=upload)
+
+        Path.unlink(
+            data_path / "subtractions" / subtraction.id / "subtraction.1.bt2",
+        )
```
igboyes, line 632:
```suggestion
        (data_path / "subtractions" / subtraction.id / "subtraction.1.bt2").unlink()
```

tests/subtractions/test_db.py
```diff
@@ -2,60 +2,87 @@
...
-    [
-        {"id": "sub_1", "subtractions": ["foo", "bar"]},
-        [
-            {"id": "sub_1", "subtractions": ["foo", "bar"]},
-            {"id": "sub_2", "subtractions": ["foo"]},
-            {"id": "sub_3", "subtractions": []},
-        ],
-    ],
+    "case",
+    ["single", "multiple"],
```
igboyes, line 15: I see what you were saying about this test.

I think the most elegant solution is to convert to a class-based test with `test_single` and `test_multi` methods.

igboyes (changes requested):

tests/subtractions/test_db.py
Bryce-Davidson: With the parameterization it has now it runs in parallel when using`xdist`

tests/subtractions/test_api.py
Bryce-Davidson: Yes, sorry. An artifact of when I thought I needed it in all tests. removing.

tests/subtractions/test_db.py
igboyes: It should either way.

igboyes (approved):

PR #3100, opened 2024-08-23
Determine file size during file migration

ReeceHoffmann:
## Changes

Adjust the file migration to work with the legacy data

Changes:

- Attempt to recover the size of any files where size was not included in the original db
  - if the file cannot be found, then write assume file size is zero since the file was likely deleted
- Add a whitelist to only migrate files of type `reads`, `subtractions`, `references` and `HMMs`
  - This whitelist includes all types currently considered valid based on the postres table
  - Based on the legacy database the only file type excluded is the `sample_replacement` type.

## Pre-Review Checklist
* [x] All changes are tested.


ReeceHoffmann:
re: the conversation about excluding `sample_replacement`. I think excluding them is fine. Samples never directly reference the `file` instead reference with `SQLSampleReads` table which is populate for legacy samples at the time of running the `migrate_sample_files` migration using the data in the `from` field of the `sample` document

PR #3128, opened 2024-11-14
fix: prevent migration failure when pathoscope analysis lack `subtracted_count`

ReeceHoffmann:
Changes the existing migration to handle the case where `subtracted_count` DNE. When no `subtracted_count` exists then it is fully excluded from the newly nested version of results as well

## Pre-Review Checklist
* [x] All changes are tested.


igboyes (commented):
I think we should get rid of `buffered_bulk_writer` here and just use `update_one` serially.

igboyes (approved):

PR #3131, opened 2024-11-21
Fix analysis file migration

ReeceHoffmann:
Due to long run times during NUVs analysis compression the mongodb cursor would expire after 30 minutes of no activity. 

changes:

- introduces a persistent session context manger that creates a session and refreshes it every 10 minutes
- Modifies the existing migration to use the persistent session to prevent timeout due to long compression times

## Pre-Review Checklist
* [x] All changes are tested.


igboyes (changes requested):
`persistent_session` is going to cause problems

igboyes (approved):

PR #3132, opened 2024-11-21
Upgrade subtraction `created_at`, `nickname` and `file.name` fields

ReeceHoffmann:
In order to get the legacy subtraction up to date with modern virtool data models a few updates are required to some of the legacy subtractions.

Changes:

1. `created_at` was missing in all legacy subtractions. There is no reliable way to derive this information. Best attempt at restoring this value is by using whichever is oldest of the time the subtractions index was first added to the file system or the last date the index was modified. \
**NOTE:** It is possible that this method will result in highly inaccurate timestamps if the user has moved the copied or otherwise changed the birth time of the file.

3. `nickname` is missing from some legacy subtractions. Any subtractions without a nickname have the default value of `""` set
4. `file.name` is missing for subtractions where the provided name was used directy as a an id without modification. (no-prepending of a random string) In cases where file name is missing the `file.id` is copied into `file.name`

Pre-Review Checklist
* [x] All changes are tested.

### DIFF
--- a/assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
+++ b/assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
@@ -37,7 +37,7 @@ async def upgrade(ctx: MigrationContext):
             ctx.data_path
             / "subtractions"
             / subtraction["_id"].replace(" ", "_")
-            / "reference.1.bt2"
+            / "subtraction.1.bt2"
         ).stat()
 
         created_at = (
@@ -81,7 +81,7 @@ async def test_upgrade(ctx: MigrationContext, snapshot):
 
     subtraction_path = ctx.data_path / "subtractions" / "legacy"
     subtraction_path.mkdir(exist_ok=True, parents=True)
-    index_file = subtraction_path / "reference.1.bt2"
+    index_file = subtraction_path / "subtraction.1.bt2"
     index_file.write_text("subtraction_index")
 
     await upgrade(ctx)
--- a/assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
+++ b/assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
@@ -34,7 +34,10 @@ async def upgrade(ctx: MigrationContext):
         {"created_at": {"$exists": False}},
     ):
         index_stats = (
-            ctx.data_path / "subtractions" / subtraction["_id"] / "reference.1.bt2"
+            ctx.data_path
+            / "subtractions"
+            / subtraction["_id"].replace(" ", "_")
+            / "reference.1.bt2"
         ).stat()
 
         created_at = (
--- a/assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
+++ b/assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
@@ -25,12 +25,12 @@ required_alembic_revision = None
 
 
 async def upgrade(ctx: MigrationContext):
-    await ctx.mongo.subtractions.update_many(
+    await ctx.mongo.subtraction.update_many(
         {"nickname": {"$exists": False}},
         {"$set": {"nickname": ""}},
     )
 
-    async for subtraction in ctx.mongo.subtractions.find(
+    async for subtraction in ctx.mongo.subtraction.find(
         {"created_at": {"$exists": False}},
     ):
         index_stats = (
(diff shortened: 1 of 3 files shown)


assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
```diff
@@ -0,0 +1,91 @@
...
+            {
+                "_id": "legacy",
+                "file": {"id": "legacy_file_id", "name": None},
+            },
+        ],
+    )
+
+    subtraction_path = ctx.data_path / "subtractions" / "legacy"
+    subtraction_path.mkdir(exist_ok=True, parents=True)
+    index_file = subtraction_path / "subtraction.1.bt2"
```
igboyes, line 84:
```suggestion
    index_file_path = subtraction_path / "subtraction.1.bt2"
```

assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
```diff
@@ -0,0 +1,91 @@
...
+    index_file.write_text("subtraction_index")
+
+    await upgrade(ctx)
+
+    assert [
+        subtraction async for subtraction in ctx.mongo.subtraction.find({})
+    ] == snapshot(matcher=path_type({".*created_at": (datetime.datetime,)}, regex=True))
```
igboyes, line 91:
```suggestion
    assert await ctx.mongo.subtraction.find({}).to_list(None) == snapshot(matcher=path_type({".*created_at": (datetime.datetime,)}, regex=True))
```

assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
```diff
@@ -0,0 +1,91 @@
...
+    await ctx.mongo.subtraction.update_many(
+        filter={"file.name": None},
+        update=[
+            {
+                "$set": {
+                    "file.name": "$file.id",
+                },
+            },
+        ],
+    )
```
igboyes, line 63: Needs a transaction around it. Multi updates are not atomic.

assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
```diff
@@ -0,0 +1,91 @@
...
+            else index_stats.st_mtime
+        )
+
+        await ctx.mongo.subtraction.update_one(
+            {"_id": subtraction["_id"]},
+            {"$set": {"created_at": datetime.datetime.fromtimestamp(created_at)}},
+        )
+
+    await ctx.mongo.subtraction.update_many(
+        filter={"file.name": None},
```
igboyes, line 55: Is this the same sort of thing as line 29? `filter=` is not used anywhere else in the codebase so leave out if this is equivalent.

assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
```diff
@@ -0,0 +1,91 @@
...
+
+        created_at = (
+            index_stats.st_ctime
+            if index_stats.st_ctime < index_stats.st_mtime
+            else index_stats.st_mtime
+        )
+
+        await ctx.mongo.subtraction.update_one(
+            {"_id": subtraction["_id"]},
+            {"$set": {"created_at": datetime.datetime.fromtimestamp(created_at)}},
```
igboyes, line 51: I think there is an issue with local versus UTC time here.

* Make sure the local time from the filesystem is converted to UTC.
* Use `arrow`. Avoid `datetime` as much as possible.

assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
```diff
@@ -0,0 +1,91 @@
...
+
+# Change this if an Alembic revision is required to run this migration.
+required_alembic_revision = None
+
+
+async def upgrade(ctx: MigrationContext):
+    await ctx.mongo.subtraction.update_many(
+        {"nickname": {"$exists": False}},
+        {"$set": {"nickname": ""}},
+    )
```
igboyes, line 31: Needs a transaction around it. Multi updates are not atomic.

igboyes (changes requested):

assets/revisions/rev_oxu8ghlvuqmh_update_subtraction_nicknames_created_at_and_file_names.py
ReeceHoffmann: Same as above

igboyes (commented):
I think each of the `update_many` should be in their own transactions. The loop doesn't need to be in one.

Thoughts?

igboyes (approved):

PR #3134, opened 2024-12-04
Support user type flag

ReeceHoffmann:
Changes:
- Adds a `alembic` migration to add a `type` field to the `users` table in postgres
- Adds a `virtool` migration to add a `type` field to the `users` collection in mongodb
- Enforces usage of latest migration at startup
- updates users data layer to handle the new parameter
   - find only returns users of type `user`
   - create functions assume type of `user` but in some cases allow specification of the type
   - update function only allows updating of `user` type users
   - Fetching a single user ignores the flag
- Users of type `bot` are blocked from authenticating via session or logging on 

## Pre-Review Checklist
* [x] All changes are tested.

### DIFF
--- a/assets/alembic/versions/e52b2748f384_add_user_type_field.py
+++ b/assets/alembic/versions/e52b2748f384_add_user_type_field.py
@@ -1,0 +1,42 @@
+"""add user type field
+
+Revision ID: e52b2748f384
+Revises: b79c1e6cf56c
+Create Date: 2024-12-03 18:03:12.062643+00:00
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "e52b2748f384"
+down_revision = "b79c1e6cf56c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    user_type_enum = postgresql.ENUM("user", "bot", name="usertype")
+    user_type_enum.create(op.get_bind())
+
+    op.add_column(
+        "users",
+        sa.Column(
+            "type",
+            postgresql.ENUM("user", "bot", name="usertype"),
+            nullable=True,
+        ),
+    )
+
+    op.execute("UPDATE users SET type = 'user'")
+
+    op.alter_column("users", "type", nullable=False)
+
+
+def downgrade() -> None:
+    op.drop_column("users", "type")
+
+    user_type_enum = postgresql.ENUM("user", "bot", name="usertype")
+    user_type_enum.drop(op.get_bind())
--- a/tests/account/test_api.py
+++ b/tests/account/test_api.py
@@ -128,7 +128,7 @@ async def test_update_settings(data, status, spawn_client, resp_is, snapshot):
 
 
 async def test_get_api_keys(
-        fake: DataFaker,
+    fake: DataFaker,
     mongo: Mongo,
     spawn_client: ClientSpawner,
     snapshot,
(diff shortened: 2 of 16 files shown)


tests/account/test_api.py
```diff
@@ -525,6 +526,31 @@ async def test_login(
...
+    client = await spawn_client()
+
+    await mongo.users.insert_one(
+        {
+            "user_id": "abc123",
+            "handle": "foobar",
+            "password": hash_password("p@ssword123"),
+            "type": "bot",
+        },
+    )
```
igboyes, line 543: Doesn't the `fake` fixture work for this?

If not, please update to support `type` if that is what you need for this test.

virtool/account/data.py
```diff
@@ -387,10 +387,14 @@ async def login(self, data: CreateLoginRequest) -> str:
         # the database and/or password are invalid.
         document = await self._mongo.users.find_one({"handle": data.username})
 
-        if not document or not await validate_credentials(
-            self._mongo,
-            document["_id"],
-            data.password,
+        if (
+            not document
+            or document["type"] == UserType.bot
```
igboyes, line 392: Safer.
```suggestion
            or document["type"] != UserType.user
```

virtool/api/authentication.py
```diff
@@ -155,6 +157,9 @@ async def authenticate_with_session(req: Request, handler: Callable) -> Response
     if not user.active:
         raise APIUnauthorized("User is deactivated", error_id="deactivated_user")
 
+    if user == UserType.bot:
+        raise HTTPForbidden("Cannot authenticate system users")
```
igboyes, line 161: Please use `APIForbidden`.

virtool/users/data.py
```diff
@@ -404,9 +411,14 @@ async def update(self, user_id: str, data: UpdateUserRequest) -> User:
         :param data: the update data object
         :return: the updated user
         """
-        if not await id_exists(self._mongo.users, user_id):
+        user = await self._mongo.users.find_one({"_id": user_id})
```
igboyes, line 414:
```suggestion
        user = await self._mongo.users.find_one({"_id": user_id}, projection=["type"])
```

virtool/users/data.py
```diff
@@ -404,9 +411,14 @@ async def update(self, user_id: str, data: UpdateUserRequest) -> User:
...
+
+        if not user:
             raise ResourceNotFoundError("User does not exist")
 
+        if user["type"] == UserType.bot:
+            raise HTTPForbidden("Cannot update system users")
```
igboyes, line 420: Use `APIForbidden`.

virtool/users/pg.py
```diff
@@ -36,6 +37,11 @@ class SQLUserGroup(Base):
     )
 
 
+class UserType(str, SQLEnum):
+    user = "user"
+    bot = "bot"
+
```
igboyes, line 43: Docstrings all around.

igboyes (changes requested):

assets/alembic/versions/e52b2748f384_add_user_type_field.py
```diff
@@ -0,0 +1,42 @@
...
+    user_type_enum.create(op.get_bind())
+
+    op.add_column(
+        "users",
+        sa.Column(
+            "type",
+            postgresql.ENUM("user", "bot", name="usertype"),
+            nullable=True,
+        ),
+    )
```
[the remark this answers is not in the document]
ReeceHoffmann: On board with the rename of `bot -> system`, not 100% on the use case for unknown. Is the use case for the users that have been previously deleted but are have resource relations?

PR #3136, opened 2024-12-11
chore(migration): copy all missing users from mongo to postgres

ReeceHoffmann:
Changes:
- adds migration that copies any mongo users not yet in postgres into postgres

## Pre-Review Checklist
* [x] All changes are tested.

### DIFF
--- a/assets/revisions/rev_e83uz1iz5tw2_copy_users_to_postgres.py
+++ b/assets/revisions/rev_e83uz1iz5tw2_copy_users_to_postgres.py
@@ -11,7 +11,6 @@ from sqlalchemy import Select, insert, select
 from sqlalchemy.ext.asyncio import AsyncSession
 from syrupy import SnapshotAssertion
 
-from tests.fixtures.core import StaticTime
 from virtool.groups.pg import SQLGroup
 from virtool.migration import MigrationContext
 from virtool.users.pg import SQLUser, SQLUserGroup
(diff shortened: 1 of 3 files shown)


igboyes (changes requested):
Please add typing everywhere it is missing. All new code needs to be fully typed.

assets/revisions/rev_e83uz1iz5tw2_copy_users_to_postgres.py
```diff
@@ -0,0 +1,175 @@
...
+                password=user["password"],
+                settings=user["settings"],
+            )
+
+            session.add(sql_user)
+
+            try:
+                await session.flush()
+            except IntegrityError:
+                continue
```
igboyes, line 59: Are you pretty confident about this or do you think we should check for the expected error content and reraise if it's not what we expect?

assets/revisions/rev_e83uz1iz5tw2_copy_users_to_postgres.py
```diff
@@ -0,0 +1,175 @@
...
+                )
+
+            await session.commit()
+
+
+class TestUpgrade:
+    """Verify that users are correctly moved to postgres."""
+
+    base_user: Document
+    fetch_users: Callable[[], Coroutine[None, None, Sequence[SQLUser]]]
```
igboyes, line 82: I think `Awaitable` might work better than `Coroutine`.

igboyes (approved):

PR #3400, opened 2026-04-22
fix: store bulk reference history diffs in postgresql

ReeceHoffmann:
## Summary
- Aligns the bulk reference write paths (`populate_insert_only_reference`, `populate_imported_reference`) with the two-phase history storage pattern already used by the single-record path.
- History diffs are bulk-inserted into `history_diffs` with asyncpg-safe chunking and committed before any Mongo writes; the Mongo history documents carry the `"postgres"` sentinel in place of the inline diff.
- Rollback cleans up `history_diffs` rows alongside the scoped Mongo cleanup so Mongo and PostgreSQL stay in sync when a bulk insert fails.

Closes VIR-2292.

## Test plan
- [x] `mise run test -- tests/references -n4`
- [x] `mise run test -- -n4` (full suite)
- [x] New `test_populate_insert_only_reference_rollback` in `tests/references/test_db.py` forces a Mongo failure after the PG commit and asserts both PostgreSQL `history_diffs` rows and all reference-scoped Mongo state are cleaned up.
- [x] Reference snapshot files updated to reflect the `"postgres"` sentinel in bulk-path history documents.

### DIFF
--- a/tests/references/__snapshots__/test_data.ambr
+++ b/tests/references/__snapshots__/test_data.ambr
@@ -273,46 +273,7 @@
       '_id': 'otu_id_1.0',
       'created_at': datetime.datetime(2015, 10, 6, 20, 0),
       'description': 'Remoted Cherry virus B',
-      'diff': list([
-        list([
-          'change',
-          '',
-          list([
-            None,
-            dict({
-              '_id': 'otu_id_1',
-              'abbreviation': '',
-              'created_at': datetime.datetime(2015, 10, 6, 20, 0),
-              'imported': True,
-              'isolates': list([
-                dict({
-                  'default': True,
-                  'id': 'iso_id_1',
-                  'source_name': 'S3',
-                  'source_type': 'strain',
-                }),
-              ]),
-              'issues': None,
-              'last_indexed_version': None,
-              'lower_name': 'cherry virus b',
-              'name': 'Cherry virus B',
-              'reference': dict({
-                'id': 'ref_id',
-              }),
-              'remote': dict({
-                'id': 'otu_1',
-              }),
-              'schema': list([
-              ]),
-              'user': dict({
-                'id': 1,
-              }),
-              'verified': True,
-              'version': 0,
-            }),
-          ]),
-        ]),
-      ]),
+      'diff': 'postgres',
       'index': dict({
         'id': 'unbuilt',
         'version': 'unbuilt',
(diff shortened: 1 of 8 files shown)


tests/references/__snapshots__/test_data.ambr
```diff
@@ -273,46 +273,7 @@
       '_id': 'otu_id_1.0',
       'created_at': datetime.datetime(2015, 10, 6, 20, 0),
       'description': 'Remoted Cherry virus B',
-      'diff': list([
-        list([
-          'change',
-          '',
-          list([
-            None,
-            dict({
```
ReeceHoffmann: Are the diffs themselves still checked somewhere?

virtool/references/data.py
```diff
@@ -1004,6 +1007,13 @@ async def populate_imported_reference(
             for otu in data.otus
         ]
 
+        diff_rows = [
+            {"change_id": insertion.history.id, "diff": insertion.history.diff}
+            for insertion in insertions
+        ]
```
igboyes, line 1013: Couldn't this translate to massive memory usage.

virtool/references/data.py
```diff
@@ -1012,27 +1022,32 @@ async def populate_imported_reference(
...
+            await bulk_delete_diffs(self._pg, [row["change_id"] for row in diff_rows])
+
             await asyncio.gather(
-                self._mongo.otus.delete_many({"reference.id": ref_id}),
                 self._mongo.history.delete_many({"reference.id": ref_id}),
-                self._mongo.references.delete_one({"_id": ref_id}),
                 self._mongo.sequences.delete_many({"reference.id": ref_id}),
+                self._mongo.otus.delete_many({"reference.id": ref_id}),
             )
+
```
igboyes, line 1049: Why not one gather?

virtool/references/data.py
ReeceHoffmann: Fair point, I'm actually not totally sure. This line mihht not add to much more since the string is already in memory in "insertions" but I can rework the prepare_otu_insertions as a generator to avoid either of them being made

virtool/references/data.py
ReeceHoffmann: Possibly pointless, but I was thinking about mid delete failure recover-ability

1. Postgres first since deleting the diffs out of postgres will be a major pain since we have to check the change_id in postgres against every history record in mongo to clean up after the fact
2. Everything except the reference next since they are all keyed by the reference id
3. Reference last since we can us its id to identify most other things to delete. 

I am perfectly happy to group steps 2 and 3, since for automatic recovery it in some way easier if the reference is deleted. Point 3 is only really helpful if we want to have users initiate the failed import deletion. If the reference persists it kinda complicates failure state cleanup, since we can reliably detect a failed install with the current schema. I generally want to clean up how reference install/update failure is handled as the current failure states are a bit problematic.

igboyes (approved):

PR #3416, opened 2026-04-28
feat(samples): block analyses on archived references and return 409

ReeceHoffmann:
## Summary

- Reject new analyses targeting an archived reference with `ResourceConflictError("Reference is archived")`, raised from `SamplesData.has_resources_for_analysis_job` after the existence check.
- Map the analyze handler's `ResourceError` to `APIConflict` (409) instead of `APIBadRequest` (400), aligning the endpoint with the codebase-wide convention for conflict-style failures.

## Behaviour change — 400 → 409

This intentionally flips the response status for **all** existing `ResourceConflictError` reasons surfaced by `POST /samples/{id}/analyses`, not just the new archived case:

- `Reference does not exist`
- `No ready index`
- `Subtractions do not exist: ...`
- `Reference is archived` *(new)*

Any client checking specifically for 400 on these reasons will need to be updated to expect 409.

## Tests

- `tests/samples/test_data.py::TestHasResourcesForAnalysisJob` — direct unit coverage of the guard (ok / missing / archived / missing-subtraction).
- `tests/samples/test_api.py::test_analyze` — parametrize ids renamed `400_*` → `409_*`, new `409_archived` case, assertions switched to `resp_is.conflict(...)`.
- `tests/analyses/test_api.py::test_get_archived_reference` — read-side smoke test confirming `AttachReferenceTransform` still resolves reference metadata for an analysis whose reference is archived.

Closes VIR-2299.

### DIFF
--- a/tests/samples/test_api.py
+++ b/tests/samples/test_api.py
@@ -1151,7 +1151,7 @@ class TestAnalyze:
         return client
 
     @staticmethod
-    async def _seed(
+    async def _insert_analysis_resources(
         mongo: Mongo,
         static_time,
         *,
@@ -1217,7 +1217,7 @@ class TestAnalyze:
         snapshot: SnapshotAssertion,
         static_time,
     ):
-        await self._seed(mongo, static_time)
+        await self._insert_analysis_resources(mongo, static_time)
 
         resp = await self._post(analyze_client)
 
@@ -1232,7 +1232,7 @@ class TestAnalyze:
         resp_is,
         static_time,
     ):
-        await self._seed(mongo, static_time, reference=False)
+        await self._insert_analysis_resources(mongo, static_time, reference=False)
 
         resp = await self._post(analyze_client)
 
@@ -1245,7 +1245,7 @@ class TestAnalyze:
         resp_is,
         static_time,
     ):
-        await self._seed(mongo, static_time, archived=True)
+        await self._insert_analysis_resources(mongo, static_time, archived=True)
 
         resp = await self._post(analyze_client)
 
@@ -1258,7 +1258,7 @@ class TestAnalyze:
         resp_is,
         static_time,
     ):
-        await self._seed(mongo, static_time, index=False)
+        await self._insert_analysis_resources(mongo, static_time, index=False)
 
         resp = await self._post(analyze_client)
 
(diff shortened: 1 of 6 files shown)


igboyes (dismissed):

tests/samples/test_api.py
```diff
@@ -1133,85 +1133,175 @@ async def test_find_analyses(
...
+                {
+                    "_id": "test",
+                    "name": "Test",
+                    "created_at": static_time.datetime,
+                    "all_read": True,
+                    "all_write": True,
+                    "ready": True,
+                },
+            )
+
```
igboyes, line 1193: Refactor into class-based test.

tests/samples/test_api.py
```diff
@@ -1133,85 +1133,175 @@ async def test_find_analyses(
...
+        static_time,
+        *,
+        reference: bool = True,
+        archived: bool = False,
+        index: bool = True,
+        index_ready: bool = True,
+        subtraction: bool = True,
+        sample: bool = True,
+    ) -> None:
+        if reference:
```
igboyes: I really hate this, but I guess we can let it pass…

igboyes (changes requested):
Try to refactor these nasty parametrized API tests to classes as you go.

tests/samples/test_api.py
ReeceHoffmann: Yeah your right, fixed

tests/samples/test_data.py
```diff
@@ -386,3 +386,66 @@ async def test_non_group_member_denied(
...
+            {
+                "_id": "test_index",
+                "reference": {"id": "test_ref"},
+                "ready": True,
+                "version": 1,
+            },
+        )
+        await mongo.subtraction.insert_one(
+            {"_id": "subtraction_1", "name": "Subtraction 1"},
+        )
```
igboyes: Use gather to speed up.

tests/samples/test_data.py
```diff
@@ -386,3 +386,66 @@ async def test_non_group_member_denied(
...
+        await mongo.indexes.insert_one(
+            {
+                "_id": "test_index",
+                "reference": {"id": "test_ref"},
+                "ready": True,
+                "version": 1,
+            },
+        )
+        await mongo.subtraction.insert_one(
```
igboyes: I believe there is a fake class for this.

igboyes (changes requested):

igboyes (approved):

PR #3420, opened 2026-04-29
feat: support archived filter on reference and index list endpoints

ReeceHoffmann:
## Summary

- Adds tri-state `archived` query param (omit | `include` | `only`) to `GET /references/v1` and `GET /indexes`. Default behavior is active-only.
- Documents a new project-wide `{field}=include|only` lifecycle filter convention in `AGENTS.md`, with `archived` as the canonical example.
- Extracts `compose_archived_filter` in `virtool/references/db.py` as the single source of truth for the lifecycle → Mongo filter mapping; `compose_base_find_query` and the indexes find paths both use it.
- The `/indexes` filter applies to the index's reference (covers both `ready=false` paginated and `ready=true` aggregation paths). This means the analysis-creation form's `useCompatibleIndexes` hook silently excludes archived-reference indexes with no virtool-ui change required.
- Also gitignores the locally-generated `virtool/spec/openapi.json` artefact.

Closes VIR-2296.


virtool/references/db.py
```diff
@@ -249,23 +249,51 @@ async def check_source_type(mongo: "Mongo", ref_id: str, source_type: str) -> bo
...
+
+    :param archived: lifecycle filter mode
+    :return: a Mongo filter dict for the ``archived`` field
+
+    """
+    if archived == "only":
+        return {"archived": True}
+    if archived == "include":
+        return {}
+    return {"archived": False}
```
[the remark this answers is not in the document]
ReeceHoffmann: No: Migration's must have occurred to add the field prior to this version of the code base being run. This is enforced in code

tests/indexes/test_api.py
```diff
@@ -29,8 +29,19 @@
...
+        [
+            (None, {"bar", "foo"}),
+            ("include", {"bar", "foo", "qux"}),
+            ("only", {"qux"}),
+        ],
+        ids=["default", "include", "only"],
+    )
     async def test(
         self,
+        archived_param: str | None,
```
igboyes: Can this just be archived?

AGENTS.md
```diff
@@ -141,6 +147,19 @@ Use **"delete"** for permanently destroying an entity. Use **"remove"** for
...
 removing an isolate from an OTU).
 
+### Lifecycle Filter Convention
+
+For list endpoints that expose a lifecycle field (e.g. `archived`), filter on
+that field with a tri-state query parameter named after the field:
+
+- absent → only entities where the field is `False` (the active set)
+- `{field}=include` → both states
+- `{field}=only` → only entities where the field is `True`
```
igboyes: This breaks with previous convention doesn't it.

1. Not set: both state.
2. True: archived only
3. False: not archived only
4. null: same as 1

igboyes (approved):

PR #3432, opened 2026-05-07
feat(caches): add caches module foundation

ReeceHoffmann:
## Summary

- introduce `virtool/caches/` with content-addressed cache keys (SHA-256 over canonical params + normalized semver), a Postgres-backed `CacheRepo`, on-disk blob path helper, and `CacheType` enum
- create `caches` table via Alembic migration `c3a7e9b4d1f2`; `parent_id` is intentionally not a foreign key (Mongo `_id` reference, orphans cleaned out-of-band)
- `CacheRepo.get` time-buckets `last_accessed_at` writes (5 min) to keep eviction ordering useful without a write per read; `put` uses `ON CONFLICT (key) DO NOTHING` so racing writers converge on a single row
- foundation only — higher-level concerns (eviction, blob lifecycle, jobs API integration) land in follow-ups


virtool/caches/data.py
```diff
@@ -0,0 +1,116 @@
...
+            cache = (
+                await session.execute(select(SQLCache).where(SQLCache.key == key))
+            ).scalar_one_or_none()
+
+            if cache is None:
+                return None
+
+            now = virtool.utils.timestamp()
+
+            if now - cache.last_accessed_at >= LAST_ACCESSED_BUCKET:
```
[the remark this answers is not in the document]
ReeceHoffmann: image

ReeceHoffmann:
fyi: gemini's assertions about the time zones are incorrect. Gemini's claims are based around virtool using local time zones, whereas virtool actually uses naive time zones across the codebase

assets/alembic/versions/c3a7e9b4d1f2_create_caches_table.py
```diff
@@ -0,0 +1,54 @@
...
+
+CACHE_TYPES = (
+    "reference_mapping_index",
+    "subtraction_mapping_index",
+    "sample_trimmed_reads",
+)
+
+
+def upgrade() -> None:
+    cache_type_enum = sa.Enum(*CACHE_TYPES, name="cachetype")
```
igboyes: Don't use pg enums. It turns it is generally advised not to use them and enforce the rule in code. They make changing the variants challenging.

virtool/caches/data.py
```diff
@@ -0,0 +1,191 @@
...
+        self._pg = pg
+        self._storage = storage
+
+    async def get(
+        self,
+        cache_type: CacheType,
+        parent_id: str,
+        tool_name: str,
+        tool_version: str,
+        params: dict[str, Any],
```
igboyes: This is still tightly coupled to how caches would currently be used in workflows.

It is quite possible we wind up with a cache object that is the result of two tools. Maybe you could argue this is already the case.

virtool/caches/data.py
```diff
@@ -0,0 +1,191 @@
...
+                .where(SQLCache.key == key)
+                .returning(SQLCache.blob_uuid),
+            )
+            blob_uuid = result.scalar_one_or_none()
+            await session.commit()
+
+        if blob_uuid is not None:
+            await self._storage.delete(_blob_key(blob_uuid))
+
+    async def delete_by_parent(self, parent_id: str, cache_type: CacheType) -> int:
```
igboyes: Are these method used anywhere?

Delete by key is okay. Delete by parent_id is problematic because the parent ID type is going to change with a migration to Postgres.

Overall, I wonder if we should just focus on reaping by last_accessed_at and avoid implementing direct deletion. If a use case emerges, we can add the functionality.

virtool/caches/utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+import hashlib
+import json
+from typing import Any
+
+from semver import VersionInfo
+
+from virtool.caches.types import CacheType
+
+_KEY_FIELD_SEPARATOR = "\x00"
```
igboyes: Can just be defined in the function.

virtool/caches/utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+def normalize_semver(tool_version: str) -> str:
+    """Return ``tool_version`` with build metadata stripped.
+
+    The pre-release segment is preserved so that ``1.2.3-rc.1`` and
+    ``1.2.3`` produce different keys. Build metadata (``+...``) is dropped
+    because it does not change tool behavior.
+    """
+    version = VersionInfo.parse(tool_version.lstrip("v"))
+    return str(version.replace(build=None))
```
igboyes: I don't see what the benefit of this behaviour is, only the risk and additional complexity.

virtool/caches/data.py
```diff
@@ -0,0 +1,191 @@
...
+A coarse bucket keeps eviction ordering useful without a write on every read.
+"""
+
+
+_CACHE_KEY_CONSTRAINT = "cache_key"
+"""Name of the unique constraint on ``caches.key``.
+
+Pinned in the migration, the SQLAlchemy model, and here so we can distinguish
+the expected duplicate-key race from any other integrity violation.
+"""
```
igboyes: Same constant should be used here and in model.

__table_args__ = (UniqueConstraint("key", name="cache_key"),)

tests/caches/test_data.py
```diff
@@ -0,0 +1,390 @@
...
+
+class TestBlobKey:
+    def test_pins_v1_namespace(self):
+        assert (
+            _blob_key("3d8f1c527a4e4b9c9a521c8e7d3b0a91")
+            == "caches/v1/3d8f1c527a4e4b9c9a521c8e7d3b0a91"
+        )
+
+
+class TestPut:
```
igboyes: Name not appropriate.

virtool/caches/types.py
```diff
@@ -0,0 +1,9 @@
+from virtool.pg.utils import SQLEnum
+
+
+class CacheType(str, SQLEnum):
```
igboyes: This should be parent_type. Otherwise we are directly coupling ourselves to the use cases we already have.

I am not sure if we even really need parent_type.

tests/caches/test_data.py
igboyes: The data layer tests focus a lot on the database and storage instead of whether the data layer behaves correctly.

For creating a cache, it should check the return value (done) but also that getting after returns the same thing. Instead of diving down to blob, check that the data layer is able to return the data itself.

tests/caches/test_data.py
```diff
@@ -0,0 +1,390 @@
...
+            {},
+        )
+
+        got = await data_layer.caches.get(
+            CacheType.sample_trimmed_reads,
+            "sample_alpha",
+            "skewer",
+            "0.2.2",
+            {},
+        )
```
igboyes: Does static_time do anything here?

tests/caches/test_data.py
```diff
@@ -0,0 +1,390 @@
...
+        assert got is not None
+        assert got.id == cache.id
+
+    async def test_does_not_touch_within_bucket(
+        self,
+        data_layer: DataLayer,
+        pg: AsyncEngine,
+        static_time,
+        mocker,
+    ):
```
igboyes, line 134: Can't tell what this is doing.

virtool/caches/data.py
```diff
@@ -0,0 +1,191 @@
...
+from virtool.caches.utils import build_stored_params, derive_key
+from virtool.data.domain import DataLayerDomain
+from virtool.data.errors import CacheAlreadyExistsError
+from virtool.pg.utils import extract_constraint_name
+from virtool.storage.protocol import StorageBackend
+
+LAST_ACCESSED_BUCKET = timedelta(minutes=5)
+"""Max staleness of ``last_accessed_at`` before ``get`` refreshes it.
+
+A coarse bucket keeps eviction ordering useful without a write on every read.
```
igboyes: I find "bucket" confusing. Should be margin or range or something. Or is Claude right on this naming?

igboyes (commented):
Let's discuss some of these design decisions.

virtool/caches/models.py
```diff
@@ -0,0 +1,32 @@
...
+from virtool.models.base import BaseModel
+
+
+class Cache(BaseModel):
+    """A reusable artifact keyed by a SHA-256 of its inputs."""
+
+    id: int
+    key: str
+    storage_key: str
+    type: str
```
igboyes: What is the point of this now?

virtool/caches/models.py
```diff
@@ -0,0 +1,32 @@
...
+    params: dict[str, Any]
+    parent_id: str
```
igboyes: This could be an int in the near future. I am wondering what the point is.

virtool/workflow/models.py
```diff
@@ -0,0 +1,13 @@
...
+class WorkflowCacheParams(BaseCacheParams):
+    """Cache params shared by every workflow-produced cache entry.
+
+    Workflows subclass this to declare the additional keys that contribute
+    to their cache key. ``tool_name`` and ``tool_version`` are required
+    because every workflow cache is keyed on the tool that produced it.
+    """
+
+    tool_name: str
+    tool_version: str
```
igboyes: I thought we were using workflow name and version. We came up with some cases where tool-specific caches break down quickly.

igboyes (changes requested):

igboyes (approved):

PR #3527, opened 2026-06-04
feat(subtractions): dual-write subtraction mutations to postgres

igboyes:
## Summary

- Wraps all subtraction create, update, delete, and finalize operations in dual transactions so changes are written to both MongoDB and PostgreSQL simultaneously
- Uses `both_transactions` to keep Mongo and Postgres in sync, laying the groundwork for a future Mongo-free read path for subtractions
- Adds a snapshot test verifying the Postgres row is correctly populated after finalization


igboyes:
@coderabbitai review

PR #3571, opened 2026-06-09
fix: release reserved uploads on all sample-deletion paths

igboyes:
## Summary

- Reserved uploads were only released in the `job_remove` API handler, keyed off `SQLSampleReads` rows — so deletions triggered by other paths (e.g. admin delete) and uploads whose workflow-written reads row lacked an `upload_id` were never released.
- Moved upload-release logic into `SamplesData.delete`, using the sample's `uploads` array directly, so every deletion path releases reserved uploads atomically in the same transaction.
- Similarly fixed `SamplesData.finalize` to key upload cleanup off the `uploads` array instead of joining through `SQLSampleReads`, ensuring cleanup works even when reads rows have no `upload_id`.

### DIFF
--- a/tests/samples/__snapshots__/test_data.ambr
+++ b/tests/samples/__snapshots__/test_data.ambr
@@ -256,7 +256,24 @@
         'name': 'Pear',
       }),
     ]),
-    'uploads': None,
+    'uploads': list([
+      dict({
+        'created_at': 'not_approximately_now_datetime',
+        'id': 1,
+        'name': 'test.fq.gz',
+        'ready': True,
+        'removed': True,
+        'removed_at': 'not_approximately_now_datetime',
+        'reserved': False,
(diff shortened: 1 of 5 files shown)


igboyes:
Thanks @sourcery-ai. Addressed in 7455ce247:

- **Redundant Mongo read in `delete`** — fixed. `SamplesData.delete` now reads upload ids from the already-fetched `sample.uploads` instead of a separate `get_one_field` call, dropping the extra round-trip.
- **Read-outside-transaction race** — intentionally not changing. A sample's `uploads` array is written once at creation and never mutated afterward, so the read-then-mutate window cannot diverge. Pulling the read into the transaction with `mongo_session` would add a query to guard a race that can't occur. (`finalize` similarly has no surrounding Mongo transaction to move the read into — its `update_one` is standalone.)

PR #3594, opened 2026-06-30
Add index build and wfindex v2 compatibility foundations

ReeceHoffmann:
This PR introduces SQLite-backed index artifacts (`index.sqlite`) for workflow execution, replacing the previous in-memory JSON parsing. Additionallity it introduces task based index generation producing a legacy references.json file. Task based index generation is not active in this PR.

### Key Changes
  - **Task-backed Index Generation**: Task-backed index builds generate and store `reference.json.gz` as a compressed JSON artifact.
  - **Workflow Runner SQLite Conversion**: The workflow loader (`WFIndex`) reads from a pre-built `index.sqlite` if available. If the SQLite artifact is missing (always for now), the workflow runner downloads the fallback `reference.json.gz` and dynamically converts it into SQLite format locally on the runner.
  - **SQLite Artifact Schema & Builder**: Added the SQLAlchemy schema and database builder to handle indexing and batch querying of OTUs and sequences.
  - **Database Schema Upgrades**: Added an Alembic migration to allow `sqlite` in index file types and registered a database revision to backfill the `task` field on legacy index documents


virtool/indexes/models.py
```diff
@@ -12,7 +12,13 @@ class IndexNested(BaseModel):
     version: int
 
 
+class IndexBuild(BaseModel):
+    status: str
```
igboyes: This should be nailed down I think.

Job is using the modern wording: `state`. Coalesce on that instead of `status`. Map to a conserved set of variants. The job state and task status variants are not conserved.

---

Alternatively, I wonder if we should even have this `build` field. Could we just have a `state` field on the index itself with variants `CREATED | PENDING | READY` or similar. This would supersede the `ready` field, which could be dropped in the future.

The task will be quick and I don't know if the progress is all that helpful.

If we wanted to improve feedback in the future, it would be easier once everything is in Postgres.

virtool/workflow/data/indexes.py
```diff
@@ -40,26 +43,18 @@ class WFIndex:
...
+    otu_source: "OTUSource"
+    """The source used to asynchronously iterate indexed OTUs."""
 
     @property
     def fasta_path(self) -> Path:
-        """The path to the complete FASTA file for the reference index in the workflow's
-        work directory.
-
-        """
+        """The path to the FASTA file for legacy analysis workflows."""
```
igboyes: Have we not already stopped fetching FASTAs for workflows?

igboyes (changes requested):

virtool/workflow/data/indexes.py
ReeceHoffmann: NuVs currently uses the fasta file. I have a branch that was working that was going to change that, but I don't see a reason not to just wait and make move directly to the new index version once it is merged

virtool/indexes/index_sqlite.py
```diff
@@ -0,0 +1,296 @@
...
+)
+from sqlalchemy.engine import URL, Connection, Engine
+from sqlalchemy.pool import ConnectionPoolEntry
+
+from virtool.api.custom_json import datetime_to_isoformat
+
+INDEX_SQLITE_FILE_NAME = "index.sqlite"
+COMPRESSED_INDEX_SQLITE_FILE_NAME = f"{INDEX_SQLITE_FILE_NAME}.gz"
+INDEX_SQLITE_FORMAT = "virtool-index-sqlite"
+INDEX_SQLITE_FORMAT_VERSION = "1"
```
igboyes: I think version and format should be in filename. If we want multiple iterations of this format associated with an index, just having `index.sqlite` could be a problem.

virtool/indexes/index_sqlite.py
```diff
@@ -0,0 +1,296 @@
...
+    connection: Connection,
+    otu_id: str,
+    isolate_id: str,
+    sequence: Mapping[str, Any],
+    schema_names: set[str],
+) -> None:
+    sequence_id = _get_id(sequence)
+    segment = sequence.get("segment")
+
+    if schema_names and segment not in schema_names:
```
igboyes: This enforces an invariant that Virtool's data doesn't satisfy, and I think it will break analyses on existing references.

`segment` is optional on a sequence — `check_sequence_segment` only validates it when one is supplied, and `virtool/otus/utils.py::verify` (the gate for inclusion in an index) checks only for empty OTUs, empty isolates, empty sequences, and isolate count consistency. A `segment: None` sequence on a schema'd OTU is legal and common today.

`_validate_required_schema_sequences` below has the same problem from the other direction: `required` defaults to true, so it demands every isolate carry a sequence for every schema segment. Nothing enforces that today either.

Because the `index` workflow fixture now builds this SQLite file at runtime on every analysis job, any reference violating either rule will crash every workflow that touches it — on existing production data, not just at migration time.

I'd drop both validations from the writer, or downgrade them to a `logger.warning`. The index build isn't the place to start enforcing a data contract no other code path enforces. If we do want to enforce it, that should be its own PR, with a survey of production references first.

virtool/indexes/index_sqlite.py
```diff
@@ -0,0 +1,296 @@
...
+    return reference_id
+
+
+def _insert_otu(
+    connection: Connection,
+    reference_id: str,
+    otu: Mapping[str, Any],
+) -> None:
+    otu_id = _get_id(otu)
+    schema_names = {item["name"] for item in otu["schema"]}
```
igboyes: `otu["schema"]` here and `otu["abbreviation"]` below will `KeyError` on legacy OTU documents that predate those fields. Patched OTUs come straight out of history, so they can be very old.

virtool/indexes/index_sqlite.py
```diff
@@ -0,0 +1,296 @@
...
+    event,
+    insert,
+)
+from sqlalchemy.engine import URL, Connection, Engine
+from sqlalchemy.pool import ConnectionPoolEntry
+
+from virtool.api.custom_json import datetime_to_isoformat
+
+INDEX_SQLITE_FILE_NAME = "index.sqlite"
+COMPRESSED_INDEX_SQLITE_FILE_NAME = f"{INDEX_SQLITE_FILE_NAME}.gz"
```
igboyes: Following up on my note about putting the version and format in the filename: worth being aware the name is currently duplicated. It's defined here as `INDEX_SQLITE_FILE_NAME` / `COMPRESSED_INDEX_SQLITE_FILE_NAME`, but `check_index_file_type` matches on a bare `".sqlite.gz"` literal. If the name grows a version component, that second one silently stops matching and the file gets typed as `bowtie2`.

Relatedly, `INDEX_SQLITE_FORMAT_VERSION` currently lives only inside the artifact's `metadata` table, so a reader has to open the SQLite file to learn which format it holds. And since `index_files` is keyed on `(index, name)` and `upsert_index_file` overwrites by name, a v1 and a v2 artifact can't coexist for the same index — a rebuild would silently replace the old one. Putting the version in the filename solves all three.

virtool/indexes/utils.py
```diff
@@ -11,6 +25,9 @@ def check_index_file_type(file_name: str) -> str:
     if file_name.endswith(".json.gz"):
         return "json"
 
+    if file_name.endswith(".sqlite.gz"):
```
igboyes: This bare `".sqlite.gz"` is the literal that duplicates `COMPRESSED_INDEX_SQLITE_FILE_NAME`. Derive it from the constant so a filename change can't desync the two.

virtool/indexes/utils.py
```diff
@@ -1,3 +1,17 @@
+import asyncio
+from collections.abc import AsyncIterator
+from pathlib import Path
+
+from virtool.storage.protocol import STORAGE_CHUNK_SIZE
+
+
+async def iter_file_chunks(path: Path) -> AsyncIterator[bytes]:
```
igboyes: `iter_file_chunks` has no callers anywhere in the diff. Remove it, or wire it up if something downstream is meant to use it.

virtool/indexes/api.py
```diff
@@ -194,6 +194,9 @@ async def upload(req):
     if name not in INDEX_FILE_NAMES:
         raise APINotFound("Index file not found")
 
+    if name not in JOBS_API_UPLOAD_INDEX_FILE_NAMES:
```
igboyes: This check is unreachable. `JOBS_API_UPLOAD_INDEX_FILE_NAMES` is an alias for `LEGACY_INDEX_FILE_NAMES`, and `INDEX_FILE_NAMES` is `LEGACY + TASK` where `TASK` is just `("reference.json.gz",)` — already in `LEGACY`. The two tuples have identical membership, so the `INDEX_FILE_NAMES` check on the line above has already returned by the time we get here.

Either the Jobs API is supposed to reject `reference.json.gz` (in which case `JOBS_API_UPLOAD_INDEX_FILE_NAMES` is wrong) or this should be deleted.

virtool/indexes/db.py
```diff
@@ -45,6 +46,12 @@
     "reference.rev.2.bt2",
 )
 
+TASK_INDEX_FILE_NAMES = ("reference.json.gz",)
+
+INDEX_FILE_NAMES = (*LEGACY_INDEX_FILE_NAMES, *TASK_INDEX_FILE_NAMES)
+
+JOBS_API_UPLOAD_INDEX_FILE_NAMES = LEGACY_INDEX_FILE_NAMES
```
igboyes: We now have five names for two actual sets:

```
LEGACY_INDEX_FILE_NAMES           # real
TASK_INDEX_FILE_NAMES             # real
INDEX_FILE_NAMES                  # union — contains reference.json.gz twice
JOBS_API_UPLOAD_INDEX_FILE_NAMES  # = LEGACY_INDEX_FILE_NAMES
READY_INDEX_FILE_NAMES            # = LEGACY_INDEX_FILE_NAMES (test fixture)
```

The two aliases carry no information, and one of them hides the unreachable check in `api.py`. `INDEX_FILE_NAMES` also has a duplicate entry, since `reference.json.gz` is in both source tuples.

virtool/indexes/checks.py
```diff
@@ -1,20 +1,46 @@
...
 
 
-async def check_fasta_file_uploaded(results: dict) -> None:
-    if IndexType.fasta not in results.values():
+async def check_legacy_index_files_uploaded(results: dict, data_type: str) -> None:
+    if data_type == "genome":
+        required_files = [
+            file_name
+            for file_name in LEGACY_INDEX_FILE_NAMES
+            if file_name != "reference.json.gz"
```
igboyes: This comprehension is rebuilt on every call, and it's duplicated verbatim — along with the surrounding `genome`/`barcode` branch — in `tests/fixtures/workflow_api/indexes.py`. Promote `JOB_REQUIRED_INDEX_FILE_NAMES` and `BARCODE_REQUIRED_INDEX_FILE_NAMES` to module constants and have the test fixture import them rather than reimplement the rule.

Also note `check_task_index_files_uploaded` below is never called from production code — only from its own test.

virtool/workflow/data/indexes.py
```diff
@@ -152,6 +289,268 @@ def otus_json_path(self) -> Path:
...
+def _shape_reference_json_metadata(
+    data: Mapping[str, Any],
+    index_: Index,
+) -> dict[str, Any]:
+    return {
+        "id": data.get("_id") or data.get("id") or index_.reference.id,
+        "created_at": data.get("created_at") or index_.created_at,
+        "data_type": data.get("data_type") or index_.reference.data_type,
+        "name": data.get("name") or index_.reference.name,
+        "organism": data.get("organism") or "unknown",
```
igboyes: `data.get("organism") or "unknown"` turns a legitimately empty organism into the string `"unknown"` — `""` is the default for a reference, so this will fire routinely.

One line up, `data.get("created_at") or index_.created_at` substitutes the *index's* creation time for the *reference's*.

Both are the silent-fallback pattern the new `AGENTS.md` section in this same PR warns against. `_get_id()`'s `_id`-or-`id` dual shape is the same smell.

virtool/indexes/data.py
```diff
@@ -338,6 +390,99 @@ async def finalize(self, index_id: str) -> Index:
 
         return await self.get(index_id)
 
+    @emits(Operation.UPDATE)
+    async def generate_index_file(self, index_id: str, _work_path: Path) -> Index:
```
igboyes: `_work_path` is never used — drop it, and drop the `temp_path` argument from `CreateIndexTask.build_index`.

Separately, the rollback here is fragile: `upsert_index_file` commits the PG row before the Mongo update runs, so the `except BaseException` compensating delete is the only thing keeping the two stores consistent. A crash in that window leaves an orphan row and a stored object. Consider deferring the row write until after the Mongo update, or at least documenting the window.

virtool/indexes/db.py
```diff
@@ -407,33 +419,107 @@ async def get_unbuilt_stats(
...
+                )
+                next_to_schedule += 1
+
+            if next_to_yield in completed:
+                patched = completed.pop(next_to_yield)
+                next_to_yield += 1
+                yield patched
+                continue
+
+            done, _ = await asyncio.wait(
```
igboyes, line 461: This is ~55 lines of hand-rolled ordered windowing, but both call sites immediately materialize the result into a list, so nothing actually streams — the machinery buys only bounded concurrency. An `asyncio.Semaphore` wrapped around the existing `gather` gives the same ordering and the same bound in about ten lines.

As written, a `concurrency` or `window_size` of `0` reaches `asyncio.wait(set())` here and raises `ValueError`.

tests/indexes/test_tasks.py
```diff
@@ -0,0 +1,369 @@
...
+    manifest: dict[str, int],
+    mongo: Mongo,
+    static_time: StaticTime,
+    reference_id: int | str = "hxn167",
+) -> int:
+    user = await fake.users.create()
+
+    task = await data_layer.tasks.create(CreateIndexTask, {"index_id": "task_index"})
+
+    if not await mongo.references.count_documents({"_id": "hxn167"}):
```
igboyes: This conditional exists only because `test_create_index_task_resolves_integer_reference_id` pre-creates the reference itself. Hoist reference creation out of the helper so it does one thing.

More broadly, this file is five module-level functions repeating the same eight-fixture signature and the same `_insert_indexed_otu` → `_create_task_backed_index` preamble. A `class TestCreateIndexTask` with a setup fixture would remove nearly all of it.

The `fake.references.create(id_="hxn167", name="Test Reference", organism="virus")` overrides are also avoidable — they're pinned only because the raw Mongo index document hardcodes `{"reference": {"id": "hxn167"}}`. Create the reference first and interpolate `reference.legacy_id` into the document. `organism="virus"` is then asserted back verbatim, which tests nothing.

tests/indexes/test_checks.py
```diff
@@ -1,34 +1,60 @@
...
-from virtool.indexes.db import INDEX_FILE_NAMES
+from virtool.indexes.checks import (
+    check_legacy_index_files_uploaded,
+    check_task_index_files_uploaded,
+)
+from virtool.indexes.db import LEGACY_INDEX_FILE_NAMES, TASK_INDEX_FILE_NAMES
+from virtool.indexes.utils import check_index_file_type
 
 
 @pytest.mark.parametrize("error", [None, 409])
```
igboyes: `@pytest.mark.parametrize("error", [None, 409])` with an `if error == 409: ... return` body is the flag-driven parametrize `AGENTS.md` asks us to split into explicit tests. The two new sibling tests in this same file are already written that way, so the file now mixes both styles — good opportunity to drop the parametrize on both of these.

tests/alembic/test_b82907bcac9d_convert_index_file_type_to_text.py
```diff
@@ -0,0 +1,111 @@
...
+from pathlib import Path
+
+import alembic.command
+import alembic.config
+import pytest
+from sqlalchemy import text
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
+
+DOWN_REVISION = "2b53ffa573a3"
```
igboyes: The revision's actual `down_revision` is `1ffbe2dcb108`, not `2b53ffa573a3`. This passes only because `2b53ffa573a3` is an ancestor, which means `test_downgrade` reverts several unrelated revisions on the way. Use the real parent.

igboyes (changes requested):

virtool/workflow/data/indexes.py
ReeceHoffmann: Mostly implemented. I've ignored the `id` vs `_id` concenr as imo the risk of this being a dangerous silent fallback is minimal and ti helps gaurd against mongo vs other datasource jank

virtool/indexes/db.py
ReeceHoffmann: I can remove it. The this was originally created when we were doing the sqlite version. Hopefully this will be helpful again once we move away from json as was the original plan

igboyes (approved):

PR #3597, opened 2026-06-30
Create legacy_samples table for the samples Postgres migration

igboyes:
## Summary
- Add the `legacy_samples` table plus `legacy_sample_labels` and `legacy_sample_subtractions` join tables in a single reversible Alembic revision, with matching SQLAlchemy 2.0 models in `virtool/samples/sql.py` (new `BigInteger`/`Identity(always=True)` convention).
- Real integer FKs for `user_id`, `job_id`, and `subtraction_id` (those collections are already migrated); `group` stays a nullable string and `quality` is JSONB. Mongo `workflows`/`nuvs`/`pathoscope` are dropped and derived on read later.
- Indexes cover sample listing (`(user_id, created_at DESC)`) and rights filtering (partial on `all_read`/`group_read`, btree on `group`). No code reads or writes the new tables yet — this is the schema-only first step (VIR-2527) that unblocks dual-write (VIR-2524).

### DIFF
--- a/assets/alembic/versions/edacc4a083f1_create_legacy_samples_table.py
+++ b/assets/alembic/versions/edacc4a083f1_create_legacy_samples_table.py
@@ -1,0 +1,95 @@
+"""create legacy_samples table
+
+Revision ID: edacc4a083f1
+Revises: ed265b939a84
+Create Date: 2026-06-30 20:08:49.120626+00:00
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import JSONB
+
+# revision identifiers, used by Alembic.
+revision = "edacc4a083f1"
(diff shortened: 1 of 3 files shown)


igboyes:
Addressed sourcery-ai feedback:

- **BigInteger on join FKs**: Changed `sample_id` (both join tables) and `subtraction_id` from default `Integer` to `BigInteger` to match the migration and the `BigInteger` PKs on `legacy_samples` and `subtractions`. `label_id` stays `Integer` — `labels.id` is `Integer`.
- **created_at server_default**: Skipping. The application always supplies `created_at` explicitly (copied from the Mongo document during dual-write/backfill), matching the reference model `SQLSubtraction.created_at`. A server default would diverge from convention and mask a missing timestamp.

PR #3600, opened 2026-07-02
Backfill existing history documents to legacy_history

igboyes:
## Summary
- Adds an idempotent Alembic revision (step 3 of the history → Postgres migration) that copies every Mongo `history` document into the `legacy_history` table so both stores converge before reads switch over.
- Puts the copy in a shared `copy_history_to_postgres` helper the later re-backfill revision can reuse, and reuses the dual-write `legacy_history_values` mapping so field mapping (including the sentinel-to-NULL rules) cannot drift between the two paths.
- Resolves each document's `user` to a Postgres `users.id`, raising rather than storing NULL when it can't be resolved; snapshots ids up front and commits per document with `ON CONFLICT (legacy_id) DO NOTHING` for safe re-runs.

### DIFF
--- a/assets/revisions/rev_iltemfg2gul8_copy_history_to_postgres.py
+++ b/assets/revisions/rev_iltemfg2gul8_copy_history_to_postgres.py
@@ -1,0 +1,38 @@
+"""Copy history to postgres.
+
+Revision ID: iltemfg2gul8
+Date: 2026-07-02 16:12:31.701659
+
+"""
+
+import arrow
+from structlog import get_logger
+
+from virtool.history.migration import copy_history_to_postgres
+from virtool.migration import MigrationContext
+
+logger = get_logger("migration")
+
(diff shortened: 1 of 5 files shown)


igboyes:
@sourcery-ai All three points are intentional and follow the project's documented migration process (CLAUDE.md, MongoDB migration step 3):

1. Snapshotting `_id`s up front is required — the projection-only snapshot plus per-`_id` fetch is deliberate to avoid holding a cursor open for the whole collection (cursor timeouts). Not changing.
2. Per-document commits are prescribed to bound memory and allow mid-run recovery. Not batching.
3. `user` is a required relationship for history, so raising on an unresolvable user is correct — the convention is log+NULL only for optional/historically-deletable references, which `user` is not.

CI (stale `test_apply.ambr` snapshot) fixed in 44ee56de6.

PR #3601, opened 2026-07-02
Drop the ml_models and ml_model_releases tables and ML blobs

igboyes:
## Summary
- Add an Alembic revision that drops the `ml_model_releases` and `ml_models` tables (reversing `814389ec5021` and `f8ad70032e9c`); the downgrade recreates both, including the `size` column and `model_id` FK.
- Add a Virtool revision that deletes the orphaned ML release archives under the `ml/` storage prefix. Deletion is irreversible and idempotent.
- Final teardown step of the ML/iimi removal; closes VIR-2555.


igboyes:
Both suggestions skipped:

1. `required_alembic_revision` — not read anywhere in the migration apply logic (vestigial field); ordering is enforced by `alembic_down_revision`, already set to `14c5bc110756`. The blob cleanup also touches only object storage, so it has no schema dependency on the table drop.
2. Logging — `error=repr(exc)` matches the `rev_icp41e0o5dwn_purge_iimi_analyses` precedent this revision was modeled on. These are best-effort orphan logs; the key + error message is what an operator needs, not a stack trace into obstore internals.

PR #3609, opened 2026-07-02
Read patch_to_version change list from Postgres

igboyes:
## Summary
- Switch `patch_to_version`'s change-list read from the Mongo `history` collection to the `legacy_history` table (ordered by `otu_version` descending), completing the step-4 read switch for the History → Postgres migration.
- Iterate the ordered rows with the original append/break loop so the `"removed"` sentinel, the `version > target` check, and the terminating break are preserved exactly; diff resolution stays on Postgres via `_resolve_diffs`. The function signature is unchanged, so no call sites needed edits.
- Extract a shared `coerce_otu_version` helper (reused in `legacy_history_document` and `get_most_recent_change`), and update tests: add an intermediate-version case, reseed the missing-diff test against `legacy_history`, and move the inline-diff guard test to `_resolve_diffs` where that path is now reachable.

### DIFF
--- a/tests/history/__snapshots__/test_db.ambr
+++ b/tests/history/__snapshots__/test_db.ambr
@@ -419,3 +419,58 @@
     '6116cba1.2',
   ])
 # ---
+# name: test_patch_to_version_intermediate[current]
+  dict({
+    '_id': '6116cba1',
+    'abbreviation': 'TST',
+    'imported': True,
+    'isolates': list([
+    ]),
+    'last_indexed_version': 0,
+    'lower_name': 'prunus virus f',
+    'name': 'Test Virus',
+    'reference': dict({
+      'id': 'hxn167',
+    }),
+    'schema': list([
+    ]),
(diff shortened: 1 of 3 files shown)


igboyes:
Addressed in 750822b4d:

1. **Full materialization** — switched `.scalars().all()` to `session.stream_scalars(...)` and break early, so a heavily-edited OTU no longer loads its whole history. Kept the Python-side break rather than a SQL `WHERE otu_version > target` filter: the sequential-scan-with-break semantics are a deliberate faithful port of the legacy Mongo cursor (VIR-2500 requires preserving it), and a WHERE filter would diverge on any gapped or non-monotonic version sequence.

2. **Pseudo-change dict contract** — extracted `_change_to_revert(row)` so the shape shared with `_resolve_diffs` (`_id`, `"postgres"` diff sentinel, `method_name`, coerced `otu.version`) is defined and documented in one place.

PR #3618, opened 2026-07-03
Create references Postgres tables and models

igboyes:
## Summary
- Adds the `SQLReference`, `SQLReferenceGroup`, and `SQLReferenceUser` models in `virtool/references/sql.py` and an Alembic revision creating the `legacy_references`, `legacy_reference_groups`, and `legacy_reference_users` tables — Step 1 of the references Mongo→Postgres migration (schema only; no dual-write, backfill, or reads).
- Normalizes per-reference group/user rights into composite-PK child tables and maps `source_types` to JSONB; drops `space`, `data_type`, `internal_control`, and the retired remote-flow fields (`release`, `remotes_from`, `installed`, `updates`, `updating`).
- Removes the official-reference archive guard and stops returning `installed`/`updating` (dead after the remote-machinery removal).
- FKs into `legacy_references` (`cloned_from_id`, child `reference_id`) are `BigInteger` to match the PK; outbound FKs to serial PKs stay `Integer`. `organism` is non-nullable `String` defaulting to `""`, so backfills must coalesce missing Mongo values.


igboyes:
@sourcery-ai re your two points:

1. **FK type mismatch — fixed** in 700c5943. `cloned_from_id` and the child tables' `reference_id` now use `BigInteger` to match the `legacy_references.id` BigInteger Identity PK (mirroring the `legacy_samples` convention). The remaining FKs (`user_id`, `upload_id`, `task_id`, `group_id`) point at Integer serial PKs and stay `Integer` intentionally.

2. **Nullability — no change needed.** The flagged fields (`name`, `description`, `organism`, `created_at`) are non-Optional `Mapped[...]` annotations, which SQLAlchemy 2.0 already maps to `NOT NULL`; the ORM and the migration are aligned. This is the established pattern across the codebase's models.

PR #3620, opened 2026-07-03
remove remote-reference machinery

igboyes:
## Summary
- Retire the GitHub release-check/install/update flow for references (routes, tasks, data-layer and db methods) now that the frontend no longer calls it or reads `official_installed`. Stored Mongo remote-state is left untouched with no migration, so the installed Plant Viruses reference stays analyzable.
- Drop the `release`/`remotes_from` response fields and the orphaned bulk-update modules; keep the read-only `installed` derivation and the official-reference archive guard intact.
- Fold the now HMM-only helpers from `github.py`/`releases.py` into `virtool/hmm/` and delete the two top-level modules, along with the dead `ReleaseType.REFERENCES` enum member.


igboyes:
@codex

igboyes:
Re: Sourcery review:

- **`ReferenceInstalled`** is intentionally kept — it's still the type of `ReferenceMinimal.installed` (`virtool/references/models.py:108`). The read-only `installed` derivation is deliberately retained (VIR-2594) so the already-installed Plant Viruses reference still renders; only `ReferenceRelease`/`ReferenceRemotesFrom` were removed.
- **`DataFaker.tasks.create()`** — good catch. Dropped the now-single-element `random_element` wrapper and create `CloneReferenceTask` directly in 13e7e0f9.

PR #3621, opened 2026-07-03
fix: revert removed history changes without 500

igboyes:
## Summary
- Deleting a `.removed` history change ran `"removed" - 1`, raising an uncaught `TypeError` (not a `DatabaseError`) and returning a 500.
- `HistoryData.delete()` now computes an explicit target version for removal changes — the OTU's highest numbered `otu_version` — so `patch_to_version` reverts only the removal and restores the OTU to its pre-removal state.
- Adds a regression test for reverting a `.removed` change; the pre-existing `test_revert` parametrized `remove` but always deleted a numbered change, so the path was never exercised.

### DIFF
--- a/tests/history/test_data.py
+++ b/tests/history/test_data.py
@@ -124,6 +124,32 @@ class TestDelete:
         assert await legacy_history_ids(pg) == ["6116cba1.0", "6116cba1.1"]
         assert await history_diff_ids(pg) == ["6116cba1.0", "6116cba1.1"]
 
+    async def test_removed(
+        self,
+        create_mock_history,
+        mongo: Mongo,
+        pg: AsyncEngine,
+    ):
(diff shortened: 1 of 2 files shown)


igboyes:
Re: Sourcery review — declining both suggestions:

1. `scalar_one()` won't raise here. The query is an aggregate (`SELECT max(cast(otu_version AS Integer))`), which always returns exactly one row — NULL when no rows match — so `scalar_one()` returns `None` rather than raising `NoResultFound`.

2. The "`.removed` change with no prior numbered versions" state can't occur. A removal is always preceded by the OTU's `.0` creation change (an OTU must exist before it can be removed), which carries a numbered `otu_version`. So `max(...)` is always ≥ 0 and `target_version` is never `None`. No extra guard needed.

PR #3623, opened 2026-07-03
refactor: collapse history dual-write to postgres only

igboyes:
- Removes Mongo access from history create, update, delete, and index-assignment flows — `legacy_history` and `history_diffs` in Postgres are now the sole write target (VIR-2501, migration step 6).
- Keeps `both_transactions` blocks that still coordinate non-migrated Mongo collections (indexes, otus, sequences); only the `mongo.history` writes are dropped from them.
- Leaves `migration.py`'s Mongo reads in applied backfill revisions untouched.

### DIFF
--- a/tests/history/__snapshots__/test_api.ambr
+++ b/tests/history/__snapshots__/test_api.ambr
@@ -162,54 +162,6 @@
 # name: test_revert[True-False-None].1
   list([
     dict({
-      '_id': '6116cba1.0',
-      'created_at': datetime.datetime(2017, 7, 12, 16, 0, 50, 495000),
-      'description': 'Description',
-      'diff': 'postgres',
-      'index': dict({
-        'id': 'unbuilt',
-        'version': 'unbuilt',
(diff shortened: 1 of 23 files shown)


igboyes:
Re Sourcery review:

1. **Optional caller-supplied `AsyncSession` for `add()`** — declining. Decoupling the history write from the caller's Mongo transaction is the intended outcome of migration step 6 (the recipe prescribes replacing the `both_transactions` block with a plain `AsyncSession` write). Reintroducing an optional session would re-couple the two stores and conflicts with this repo's "no optional deps with silent fallback" rule. History and the OTU Mongo documents are separate stores now by design.

2. **Stale dual-write / Mongo-history wording** — fixed in 8097da28a (history docstrings and the index-delete test docstring). The remaining `both_transactions` / `retry_both_transactions` references are accurate: those blocks still coordinate non-migrated Mongo collections (indexes, otus, sequences) with the PG history write.

PR #3636, opened 2026-07-07
feat(migration): reference analyses and history by integer foreign key

igboyes:
## Summary
- Add nullable integer `reference_id` foreign keys to the analyses and `legacy_history` Postgres tables, backfilling existing string references to their integer primary keys
- Populate `reference_id` on the write path and switch reads to the integer foreign key
- Add `fake.references.create()` to dual-write references to Mongo and Postgres for test seeding

### DIFF
--- a/assets/alembic/versions/91b32f49a8b2_add_reference_id_to_analyses_and_legacy_.py
+++ b/assets/alembic/versions/91b32f49a8b2_add_reference_id_to_analyses_and_legacy_.py
@@ -1,0 +1,55 @@
+"""add reference_id to analyses and legacy_history
+
+Phase 1 of keying ``analyses`` and ``legacy_history`` by an integer FK to
+``legacy_references`` instead of the legacy Mongo ``reference`` string. Purely
+additive: adds a nullable ``reference_id BIGINT REFERENCES legacy_references(id)``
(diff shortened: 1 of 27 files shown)


igboyes:
Thanks — engaging with both points:

**Backfill NULL detection.** `reference` is a NOT NULL column (`Mapped[str]`) on both `analyses` and `legacy_history`, so a row can never legitimately have a blank/NULL `reference`. After the set-based `UPDATE`, `reference_id IS NULL` therefore unambiguously means the legacy string resolved to no `legacy_references` row — there is no missing-vs-unresolved distinction to draw, and the error already lists the offending strings. Leaving as-is.

**Duplicated seeding helpers.** `seed_pg_reference`/`seed_reference` are identical, but the wider set Sourcery groups together actually diverges where it matters: `ensure_reference` is session-scoped, get-or-create, and returns the integer id; the migration test uses a raw insert; and each helper pins a fixed `created_at` for snapshot determinism that the dual-writing `fake.references.create()` intentionally doesn't offer. The real dual-write path is already centralized in that new faker. Consolidating the remaining test-only Postgres seeders would trade per-module clarity for a new shared-helper pattern with little benefit, so leaving them local.

PR #3638, opened 2026-07-07
feat(migration): purge orphaned reference data before backfill

igboyes:
## Summary
- Add a revision that runs before the reference-id backfill to clean up references that were deleted from Mongo without cascading, which previously made the backfill abort
- Archive orphaned OTUs, sequences, and indexes into `deleted_*` collections and drop their `legacy_history` rows, after verifying no analysis still depends on them
- Leave the backfill as a strict correctness guard once orphans are cleared

### DIFF
--- a/assets/revisions/rev_0uwpffz8adx0_purge_orphaned_references.py
+++ b/assets/revisions/rev_0uwpffz8adx0_purge_orphaned_references.py
@@ -1,0 +1,44 @@
+"""Purge orphaned references.
+
+Remove the OTU, sequence, index, and history data belonging to references whose
+Mongo document was deleted without cascading to its children. These orphans were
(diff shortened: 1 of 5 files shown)


igboyes:
Addressing Sourcery's review:

- **Null/missing `reference.id`**: valid catch — fixed in d69156b99. `_find_orphaned_reference_ids` now filters falsy ids, so a null/missing reference can't enter the orphan set and get swept by the archival `$in` query. Added a regression test.

- **`to_list` in `_archive_and_delete`**: declining — this isn't a materialization concern. `$merge` is a terminal write stage that returns **no** documents to the client, so `to_list(length=None)` yields an empty list, not the matched set. It's also required: motor's `aggregate()` is a latent cursor, so the pipeline (and thus the server-side merge) only executes once the cursor is consumed. The archival + delete run entirely server-side.

PR #3660, opened 2026-07-09
chore: add copy to the storage backend protocol

igboyes:
Object stores have no rename — re-keying an object is always a copy followed by a delete. This adds `copy` to `StorageBackend` and implements it across all five backends, as groundwork for VIR-2626 (keying sample storage objects by integer primary key).

`ObjectProvider` issues a server-side copy via obstore, so multi-gigabyte reads files never transit the API server. `FallbackStorageRouter` promotes a fallback-resident source into the primary rather than moving it, leaving the source in place until the caller has committed.

The re-keying migration itself is deferred until after the Mongo work, so `copy` has no call site yet. Note it must land before `legacy_id` can be dropped from `legacy_samples`, since sample storage prefixes still derive from it.

One known gap: S3's single-request `CopyObject` caps at 5 GiB and larger objects need a multipart copy. The integration tests cover objects spanning multiple write chunks but not that ceiling, and sample reads files exceed it. Whether obstore switches to multipart transparently needs to be established before any migration re-keys real reads files.

### DIFF
--- a/tests/storage/test_filesystem.py
+++ b/tests/storage/test_filesystem.py
@@ -88,6 +88,60 @@ class TestWrite:
         assert not (tmp_path / "target.txt").exists()
 
 
+class TestCopy:
+    async def test_ok(self, provider, tmp_path):
+        await provider.write("samples/abc/reads.fq.gz", _async_iter(b"read data"))
+
+        await provider.copy("samples/abc/reads.fq.gz", "samples/12/reads.fq.gz")
+
+        assert (
+            tmp_path / "samples" / "12" / "reads.fq.gz"
+        ).read_bytes() == b"read data"
+
+    async def test_leaves_source_in_place(self, provider, tmp_path):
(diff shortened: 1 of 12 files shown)


igboyes:
Closing unmerged. The re-keying migration this was groundwork for is superseded by [VIR-2631](https://linear.app/virtool/issue/VIR-2631/remediate-storage-keys-across-all-resources), which recommends opaque `storage_key` columns instead — under that approach no object ever moves, so `copy` would have no call site.

Rather than merge a `StorageBackend` method that all five backends must implement and nobody calls, this stays closed. If VIR-2631 lands on the integer-id approach after all, reopen this: the implementation is complete and tested, including against Garage and Azurite.

Worth recording from the work here: obstore's `copy_async(store, from_, to, *, overwrite=True)` issues a server-side copy, overwrites by default, and raises `FileNotFoundError` on a missing source. Its `rename` is copy-then-delete, not an atomic move.

PR #3673, opened 2026-07-10
feat(migration): add sequence isolate_id/segment and otu reference_id index

igboyes:
## Summary
- Promote `isolate_id` and `segment` onto `legacy_sequences` so isolate-scoped operations (e.g. removing an isolate) can filter/delete on real columns instead of JSONB queries
- Add the missing `ix_legacy_otus_reference_id` index so reference-scoped OTU list reads don't sequential-scan once that path moves to Postgres
- Completes the promoted-column contract from VIR-2637 ahead of dual-write landing in VIR-2638

### DIFF
--- a/assets/alembic/versions/5de38ebeaa78_add_sequence_isolate_id_and_segment_.py
+++ b/assets/alembic/versions/5de38ebeaa78_add_sequence_isolate_id_and_segment_.py
@@ -1,0 +1,30 @@
+"""add sequence isolate_id and segment columns and otu reference_id index
+
+Revision ID: 5de38ebeaa78
+Revises: 8bbc31ae5a8a
+Create Date: 2026-07-10 20:20:06.481417+00:00
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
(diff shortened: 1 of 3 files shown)


igboyes:
Thanks — both points were considered; declining both:

1. **`isolate_id` NOT NULL without backfill/default.** `legacy_sequences` is empty in every environment at this point in the migration chain. It was created empty by the immediately preceding revision (`8bbc31ae5a8a`), and nothing reads or writes it yet — dual-write and backfill land in VIR-2638. Adding a NOT NULL column to an empty table needs no default or backfill. Every future insert sets `isolate_id` unconditionally (`OTUData.create_sequence`), so nullable-then-tighten would just add churn.

2. **Index on `isolate_id`/`segment`.** Out of scope for this issue by design (VIR-2651 scopes exactly one index, on `legacy_otus.reference_id`). Isolate-scoped deletes filter by `otu_id` (already indexed) AND `isolate_id`; the per-OTU sequence set is tiny, so an `isolate_id` index isn't warranted. `segment` is a promoted display column, never a filter predicate, so it needs no index.

PR #3680, opened 2026-07-13
test: seed sample fixtures from the faker instead of by hand

igboyes:
## Summary

- Replace hand-built sample seeding (`mongo.samples.insert_*` and direct `SQLLegacySample` rows) with `fake.samples.create` across the samples, analyses, subtractions, and jobs suites, mutating through the data layer rather than writing DB rows.
- Rewrite the finalize/upload cluster as a real round-trip — create a sample, upload its reads, finalize, then assert the reads are still downloadable and the input uploads are cleaned from storage — replacing tests that hand-wired `SQLSampleReads`/`SQLSampleUpload` rows and asserted on internals.
- `seed_analysis` no longer invents a `legacy_samples` row from a sample's Mongo document; it resolves the parent sample by integer id or `legacy_id`.


igboyes:
Re: the Sourcery review.

**Consolidated the rights helpers** in efd1f090c. `create_rights_sample` now lives in `tests/fixtures/samples.py` and is shared by the samples and analyses suites; `create_analysis_sample` in the analyses API tests is now a thin wrapper over it. That module's `mock_sample`/`mock_samples` fixtures went at the same time — they were hand-built mongo sample documents and this branch removed their last callers.

**Not extracting a shared `QUALITY` constant.** It is already a single module-level constant in `tests/samples/test_data.py`, not repeated inline. The other two occurrences are not duplicates of it: `tests/samples/test_api.py` uses the same values as an HTTP request body for a `PATCH` (pre-existing, and it is testing the API contract, not the data layer), and `tests/workflow/data/test_samples.py` has a different, more realistic quality payload of its own. Hoisting a 7-line literal into a cross-suite constant would couple three unrelated test modules for no behavioural reason.

PR #3692, opened 2026-07-14
fix: apply migrations with long revision names

igboyes:
## Summary
- Widen the `revisions` table's `name` and `revision` columns, which were capped at 64 and 18 characters, so revisions with longer names can be recorded.
- Alter existing tables as well as widening the `CREATE TABLE IF NOT EXISTS` DDL, since that clause won't widen a table that already exists.
- Create the `revisions` table in tests with the same DDL production uses so the two can't drift apart again.

### DIFF
--- a/tests/fixtures/migration.py
+++ b/tests/fixtures/migration.py
@@ -22,8 +22,8 @@ from sqlalchemy.ext.asyncio import create_async_engine
 
 from virtool.api.custom_json import dump_string
 from virtool.config.cls import MigrationConfig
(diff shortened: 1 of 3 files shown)


igboyes:
Combined the two `ALTER TABLE`s into one in 7ad07ce2e.

Skipping the type check before the `ALTER`: `revisions` holds one row per applied revision and the migration job is the only thing touching the database at that point, so the redundant DDL on a fresh install costs nothing measurable, while a catalog lookup adds a branch that would itself need covering.

PR #3706, opened 2026-07-17
fix: reject forked and orphaned migration revisions

igboyes:
## Summary
- `order_revisions` silently dropped revisions when two shared a downgrade pointer (a fork), and silently omitted any it could not reach from the root (an orphan). Both now raise, naming the offending revisions.
- The guard lives in `order_revisions`, the chokepoint for `apply`, `show`, `depend`, and `create`, so a broken chain fails on every invocation rather than only in CI.
- Covers structural errors only. Data-dependency inversions remain out of scope (VIR-2749).

### DIFF
--- a/tests/migration/test_show.py
+++ b/tests/migration/test_show.py
@@ -331,6 +331,45 @@ class TestOrderRevisions:
         ):
             order_revisions(revisions)
 
+    def test_duplicate_ids(self):
+        """Test that an error is raised when two revisions share an id.
+
+        Such a chain orders cleanly, but `apply` records applied revisions by id and
+        would silently skip the second as already applied.
+        """
+        revisions = [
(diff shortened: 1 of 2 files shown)


igboyes:
Correct, and removed in 540977676 — the check was unreachable, not just redundant. Every revision is either the root or an entry in `revisions_by_downgrade`, and the walk only moves entries between the two, so a count mismatch implies a leftover the orphan check has already raised on.

Chasing it down did surface a real gap it never covered: two revisions sharing an id but pointing at different downgrades order cleanly through every guard, and `apply` records applied revisions by id (`apply.py:74`), so the second is silently skipped as already applied. Replaced the dead check with a duplicate-id check that fails loudly instead.

PR #3718, opened 2026-07-17
test: seed otus through fake.otus instead of hand-built mongo docs

igboyes:
## Summary
- Convert index finalize, index task, and isolate-deletion tests to seed OTUs (and sequences) via `fake.otus.create` instead of hand-inserting current-shape Mongo documents.
- Keeps seeded test data in step with production OTU/sequence shape as it evolves.
- Leaves legacy-shape, Mongo-round-trip, and Mongo-only-negative tests hand-built, since those intentionally exercise shapes `fake.otus` doesn't produce.


igboyes:
Both points are already handled in the pushed commits:

1. No double-write — `test_finalize` no longer adds an `SQLOTU` row. The old `bulk_insert_otu_rows` call was removed; `fake.otus.create` is now the sole OTU writer (both stores), and the `AsyncSession` block only inserts `SQLIndexFile` rows.
2. The snapshot was regenerated (10548bc4a) and the assertion keys off `otu.id`. `6116cba1` is the shared `test_otu` fixture id used by many tests that still seed it themselves; this test no longer depends on it, so no cross-test brittleness is introduced.

PR #3739, opened 2026-07-20
refactor: drop redundant legacy_history.index_version column

igboyes:
## Summary

- `indexes.version` is the authoritative build version, read back through the `index_id` join, making the string copy stored on `legacy_history.index_version` redundant.
- A tripwire migration verifies the copy still agrees with `indexes.version` before dropping the column; the downgrade repopulates it from the join.
- API responses are unchanged.


igboyes:
Addressed the Sourcery tripwire feedback in 0587f4f33. The inconsistency query now guards the `::bigint` cast with a `CASE` that only casts when `index_version ~ '^[0-9]+$'`, treating any non-numeric value as inconsistent. A non-numeric `index_version` now surfaces as the intended `RuntimeError` rather than an opaque cast error, and a new `test_tripwire_non_numeric_version_raises` covers it.

PR #3745, opened 2026-07-23
feat: remove legacy job-driven index build endpoints

igboyes:
## Summary
- Remove the legacy jobs-API endpoints for uploading index files and finalizing an index, since indexes are now built through the task-backed pipeline instead.
- Drop the `job` field from index API responses and models, along with the workflow SDK surface (`WFNewIndex`, `new_index`) that only supported the removed endpoints.
- Keep the `otus.json.gz` download and index read/delete endpoints, which analysis workflows and historical job-backed indexes still rely on.


igboyes:
Addressing the Sourcery review — declining both, with reasons:

1. **`JOB_INDEX_FILE_NAMES` is not dead.** It still defines `INDEX_FILE_NAMES = (*JOB_INDEX_FILE_NAMES, REFERENCE_JSON_V2_FILE_NAME)`, which gates the downloadable filename set in `indexes/api.py` (`IndexFileView`) and `indexes/data.py` (`get_index_file`). Analysis workflows still download these `.bt2`/`.fa.gz`/`reference.json.gz` files for historical job-backed indexes, so the constant is load-bearing despite the legacy name. The genuinely-dead legacy helpers (`checks.py`, `check_index_file_type`, `upload_file`, `finalize`) were removed.

2. **No internal caller relies on index `job`.** The only `.job` reads in the codebase are on *analysis* documents (analyses are job-backed), not indexes. `IndexNested` never carried `job`, and the full test suite passes with the field removed.

PR #3757, opened 2026-07-29
feat: import references from sqlite indexes

ReeceHoffmann:
- Add a versioned SQLite reference snapshot format and support importing references from .v1.sqlite files.
- Reuse complete snapshots in workflows while naming locally generated, potentially incomplete databases index.v1.sqlite.


virtool/references/tasks.py
```diff
@@ -49,31 +61,86 @@ def __init__(self, task_id: int, data, context, temp_dir):
...
         try:
-            import_data = await load_reference_from_storage(
+            self.import_data = ReferenceSourceData.parse_obj(import_data)
+        except ValidationError as err:
+            await self._set_error(f"Invalid reference data: {err}")
+
+    async def _load_json(self, key: str) -> dict | None:
+        try:
+            return await load_reference_from_storage(
                 self.data.references._storage,
```
igboyes: I think it would be best to move the whole task body into the data layer.

Reaching through to storage here is breaking the abstraction.

virtool/references/tasks.py
```diff
@@ -49,31 +61,86 @@ def __init__(self, task_id: int, data, context, temp_dir):
...
+        try:
+            async with aiofiles.open(sqlite_path, "wb") as handle:
+                async for chunk in self.data.references._storage.read(key):
+                    await handle.write(chunk)
+
+            sqlite_reference = SQLiteReference.load(sqlite_path)
+            await sqlite_reference.validate()
+            reference = await sqlite_reference.get_metadata()
+            otus = [otu async for otu in sqlite_reference.iter_otus()]
+        except (SQLiteReferenceReadError, SQLAlchemyError):
```
igboyes: `SQLiteReference` should entirely encapsulate SQLite. SQLAlchemy exceptions should not bubble up here. Catch and handle or re-raise in the class.

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+            shape_row,
+        ):
+            yield batch
+
+
+def _create_reference_sqlite_engine(path: Path) -> Engine:
+    engine = create_engine(URL.create("sqlite", database=str(path)))
+    event.listen(engine, "connect", _enable_reference_sqlite_foreign_keys)
+
+    return engine
```
igboyes: Does the SQLAlchemy async implementation not work with SQLite. Using async natively would clean up this module a lot. No more private functions called in to_thread.

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+    """A portable reference stored in SQLite."""
+
+    path: Path
+
+    @classmethod
+    async def create(
+        cls,
+        path: Path,
+        reference: Mapping[str, Any] | None,
+        otus: Iterable[Mapping[str, Any]],
```
igboyes: This is keeping the memory issue in the codebase. Can you make this an AsyncIterable?

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+            otus,
+        )
+
+    @classmethod
+    def load(cls, path: Path) -> "SQLiteReference":
+        """Load an existing SQLite reference."""
+        if not path.exists():
+            raise FileNotFoundError(path)
+
+        return cls(path)
```
igboyes, line 181: This doesn't work as a dataclass. A dataclass should take it's fields on instantiation. Those would probably be reference and otus.

This class is really a handle/abstraction on a SQLite database.

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+    Column("sequence", Text, nullable=False),
+)
+
+Index("isolates_otu_id_idx", isolates_table.c.otu_id)
+Index("sequences_isolate_id_idx", sequences_table.c.isolate_id)
+Index("sequences_segment_idx", sequences_table.c.segment)
+
+
+@dataclass(frozen=True)
+class SQLiteReference:
```
igboyes, line 134: Make this class and module the only place that reads and writes the database. SQL statements and exceptions are leaking up into callers.

virtool/workflow/data/indexes.py
```diff
@@ -3,45 +3,37 @@
...
     sequences_table,
 )
+from virtool.utils import decompress_file
+from virtool.workflow.client import WorkflowAPIClient
 
 logger = get_logger("api")
 
 _SQLITE_SEQUENCE_BATCH_SIZE = 500
-_SQLITE_OTU_BATCH_SIZE = 1
+INDEX_SQLITE_FILE_NAME = "index.v1.sqlite"
```
igboyes, line 29: WFIndex inheriting from SQLiteReference is causing a lot of problems. It makes SQLiteReference's interface and implementation messy just to support the subclassing.

There should ideally be nothing about SQLite in this module. `SQLiteReference` should be a complete and tight wrapper for the database. It is leaking heavily here.

I think if you will find a more understandable and reliable solution by following this strict layering: WFIndex -> SQLiteReference -> SQLite.

--

I think it might be worth writing the reference export code in this branch as well. It will help solidify the design with WFIndex, Import, and Export all reaching to the same API surface of SQLiteReference.

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+        self,
+        reference: Mapping[str, Any] | None,
+        otus: Iterable[Mapping[str, Any]],
+    ) -> None:
+        await asyncio.to_thread(
+            _create_reference_sqlite,
+            self,
+            reference,
+            otus,
+        )
```
igboyes: Make it a single method. Look to WFIndex for additional comment.

tests/references/test_tasks.py
```diff
@@ -205,6 +281,250 @@ async def test_import_reference_task(
...
+
+async def test_import_reference_task_rejects_invalid_source_data(
+    assert_reference_not_populated,
+    data_layer: DataLayer,
+    reference_sqlite_path: Path,
+    spawn_import_task,
+):
+    with (
+        SQLiteReference.load(reference_sqlite_path).connect() as connection,
+        connection.begin(),
```
igboyes: I don't think connect() should be exposed on SQLiteReference. That is leaky. Shortcircuit it in testing by using SQLAlchemy directly.

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+    @classmethod
+    def load(cls, path: Path) -> "SQLiteReference":
+        """Load an existing SQLite reference."""
+        if not path.exists():
+            raise FileNotFoundError(path)
+
+        return cls(path)
+
+    @contextmanager
+    def connect(self) -> Iterator[Connection]:
```
igboyes: This is leaking to the caller.

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+
+        try:
+            with engine.connect() as connection:
+                yield connection
+        finally:
+            engine.dispose()
+
+    async def validate(self) -> None:
+        """Validate that this is a complete, compatible v1 SQLite reference."""
+        await asyncio.to_thread(_validate_reference_sqlite, self)
```
igboyes: Can you just validate in load? Then we are guaranteed any instance of SQLiteReference is valid and we also don't have to call _validate_otu in iter_otus.

virtool/references/sqlite.py
```diff
@@ -0,0 +1,615 @@
...
+def _create_reference_sqlite(
+    sqlite_reference: SQLiteReference,
+    reference: Mapping[str, Any] | None,
+    otus: Iterable[Mapping[str, Any]],
+) -> None:
+    path = sqlite_reference.path
+    path.parent.mkdir(parents=True, exist_ok=True)
+
+    if path.exists():
+        path.unlink()
```
igboyes: Good idea?

igboyes (changes requested):
This need to be better layered. I think implementing export now as well might help the design and testing.

igboyes (approved):

PR #3809, opened 2026-08-19
fix: scope the sample list total to what the caller may read

igboyes:
## Summary

- `findSamples` counted every sample on the instance for `totalCount`, ignoring the readable filter that already scopes `foundCount` and the page itself, so a non-admin caller learned how many samples existed outside their permissions. The readable filter is now split from the narrowing ones (term, users, labels, workflows) and applied to the total query, matching `findReferences` and the other list endpoints.
- `totalCount` meant a scoped count on analyses and an unscoped grand total on samples (VIR-3005). Both were pinned for wire compatibility with an implementation that no longer exists. The rule the other ten endpoints already follow is now stated once on `SearchResult`: both counts are scoped to what the caller may read, and `foundCount` additionally applies the caller's filters.
- Analyses is unchanged. Its filters only scope, never narrow, so its two counts genuinely agree; the comment calling that an exception was corrected.

Admins see no change, since the readable filter is empty for them. No UI reads `samples.totalCount`, so the effect is limited to direct API consumers.

### DIFF
--- a/packages/contracts/src/search.ts
+++ b/packages/contracts/src/search.ts
@@ -2,9 +2,15 @@
  * The paginated search-result envelope returned by every server-function-backed
  * list. A domain carrying extra summary fields — a ready count, per-state job
  * counts — intersects this rather than restating the five.
+ *
+ * Both counts are scoped to what the caller may read. `totalCount` then counts
+ * everything in that scope, while `foundCount` also applies the narrowing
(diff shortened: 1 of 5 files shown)


igboyes:
Declining the Sourcery suggestion: Drizzle intentionally accepts undefined in .where() and and(), omitting those predicates. This is also the established pattern in findSamples and other data queries, so special-casing the admin path would add branching without changing query behavior.

PR #3814, opened 2026-08-20
chore: drop the alembic and legacy migration bookkeeping tables

igboyes:
## Summary

- Migration `0007` drops `alembic_version` and `revisions`, the applied-revision pointers belonging to Alembic and to Virtool's own Python migration runner. Both systems are gone — nothing in the tree is Alembic or Python, and the cluster's migration Job runs the Drizzle migrator — so `migrations.ts` and its `schema/index.ts` entry go with them. Neither table was read or written from this side, neither has nor is targeted by a foreign key, and nothing imported either mirror.
- Dropped without `CASCADE`, against what `db:generate` emits. `0004` found production carrying a trigger absent from the baseline; under `CASCADE` an equivalent unknown dependent would be destroyed silently, where the default `RESTRICT` fails and rolls the migration back. `revisions_id_seq` is owned by its table and goes with it either way.
- Verified against a scratch Postgres: `0000` through `0007` apply clean in order, and `0007` applies in one transaction to a database seeded with an `alembic_version` row and two `revisions` rows. The drop is irreversible, so confirm the row counts in production first — the point is that both tables are unread, not that they are empty.

Closes VIR-2985


igboyes:
Sourcery review:

**`IF EXISTS`** — applied in 2d5c1edce. `revisions` is created by the Python migration runner's own bootstrap rather than by Alembic, and production was stamped at `0000` rather than having it applied, so the baseline describes production rather than having built it. An instance that never ran that runner lacks the table even though `0000` declares it, and an unguarded drop there fails the migration Job and blocks every migration after it. `RESTRICT` still applies in both statements, so an unknown dependent object remains a loud failure rather than a silent cascade.

**Statement breakpoint on the second `DROP`** — declined. `--> statement-breakpoint` separates statements; there is already one between the two `DROP`s, and a trailing marker after the final statement is not what drizzle-kit emits. `0004` and `0005` both end without one.

Verified against a scratch Postgres: `0000` through `0007` apply clean in order, `0007` drops both tables from a seeded production-shaped database, and it is now a no-op against a database where both are already absent.

PR #3823, opened 2026-08-20
feat(migration): drop six dead columns and the guards around them

igboyes:
## Summary

- Migration `0009` drops `indexes.storage_key`, `index_files.index`, `settings.enable_api`, `analyses.index`, `legacy_history.reference`, and `legacy_history.index`. Each was declared in the mirror only so it would appear in the snapshot and could be dropped; nothing reads any of them. `createIndex` no longer mints a throwaway uuid for `storage_key`, and registering an artifact no longer writes the stringified `index_id`.
- `legacy_history.reference_id` becomes `NOT NULL`, so the two history reads inner join their reference and `MalformedHistoryRowError` goes. The second thing it guarded — an index with no version — was already unreachable behind the foreign key, since `indexes.version` is `NOT NULL`. `clearReferenceContents` lets sequences go with their OTUs; the comment claiming the mirror does not declare that cascade was stale.
- Verified on a scratch Postgres: `0000`–`0009` apply clean in order, and `0009` applies to a database seeded with a row in each affected table, leaving every row intact. Run against a `legacy_history` row with a NULL `reference_id` it stops on the `SET NOT NULL` rather than dropping anything.
- The `job_id`/`task_id` hand-check stays. `ck_indexes_job_or_task` is `num_nonnulls(...) <= 1`, so it forbids "both" and permits "neither", which is the case the guard exists for. Tightening it needs a production count first and is VIR-3025.

### DIFF
--- a/apps/jobs-api/src/analyses/handlers.test.ts
+++ b/apps/jobs-api/src/analyses/handlers.test.ts
@@ -78,7 +78,6 @@ beforeEach(async () => {
 			manifest: {},
 			ready: true,
 			reference_id: referenceId,
-			storage_key: "indexes/1/abc",
 			user_id: userId,
 			version: 1,
 		})
--- a/apps/tasks/src/tasks/create-index.test.ts
+++ b/apps/tasks/src/tasks/create-index.test.ts
@@ -138,7 +138,6 @@ async function claimBuild(count = 2): Promise<{
 			manifest,
 			ready: false,
 			reference_id: referenceId,
-			storage_key: `5f9a${handleCounter}legacymongoslug`,
 			user_id: userId,
 			version: 0,
 			task_id: taskId,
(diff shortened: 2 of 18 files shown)


igboyes:
Declining both Sourcery points.

`indexVersion` cannot be `number`. It is selected through a `leftJoin` on `indexes`, and `index_id` is legitimately null for a change no build has claimed yet — the join misses and Postgres returns null. Typing the field `number` would be a lie the runtime disproves on the first unbuilt change. `indexes.version` being `NOT NULL` is what makes the *pair* both-or-neither, which is what the comment above the mapping says and what the null check encodes.

The `legacySequences.otu_id` cascade is already declared. `legacy_sequences_otu_id_fkey` in `packages/data/src/db/schema/otus.ts` carries `.onDelete("cascade")`, which is the precondition this change relies on — the stale comment claiming otherwise is what the diff removes.

PR #3849, opened 2026-08-25
feat: screen pathoscope references with cd-hit representatives

ReeceHoffmann:
Replace Pathoscope's default screening behavior with CD-HIT representative selection. This reduces redundant reference sequences before screening while preserving the existing workflow integration.

### DIFF
--- a/apps/pathoscope/src/steps/createIndexes.ts
+++ b/apps/pathoscope/src/steps/createIndexes.ts
@@ -25,7 +25,9 @@ import type { PathoscopeStep } from "./types";
  */
 export const createRepresentativeIndexStep: PathoscopeStep = {
 	id: "create_representative_index",
-	description: "Ensure the representative Bowtie2 index exists locally.",
+	name: "Prepare Screening Reference",
+	description:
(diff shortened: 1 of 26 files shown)


apps/pathoscope/src/steps/createIndexes.ts
```diff
@@ -1,31 +1,35 @@
...
+ * CD-HIT-EST representatives selected from every source OTU and declared
+ * segment go into it. The point of this pass is to find which OTUs are present
+ * without paying to map against every source sequence.
  */
-export const createReferenceIndexStep: PathoscopeStep = {
-	id: "create_reference_index",
-	description: "Ensure the reference Bowtie2 index exists locally.",
+export const createRepresentativeIndexStep: PathoscopeStep = {
+	id: "create_representative_index",
+	description: "Ensure the representative Bowtie2 index exists locally.",
```
igboyes: I'd like to take the opportunity to improve these descriptions. "Ensure", "representative", "locally" are not useful to users.

Try to make descriptions that convey the flow/meaning of the workflow a bit better for non-technical audiences.

igboyes (changes requested):

